PR #4, opened 2026-02-27
prune unused parser functions

alrs:
This prunes two unused `Parser` functions.


odvcencio:
thanks for jumpin in and helping keep it clean! really!

PR #21, opened 2026-04-01
perf: Pass 3 + 4 — allocation elimination across query, walk, detection, and lexer hot paths

rsnodgrass:
## Summary

Continues the allocation-reduction campaign from #20. Eight targeted changes across the parser, query engine, walk traversal, and language detection paths — each measured with benchmarks before merging.

```
Benchmark                              Before          After           Delta
─────────────────────────────────────────────────────────────────────────────
QueryExecCompiled (allocs)             4,513/op        29/op           -99.4%
QueryExec (allocs)                     5,285/op        801/op          -85%
TypeScript full parse (allocs)         3,016/op        14/op           -99.5%
LoadLanguage (time / mem)              2.2ms / 1,461KB 1.6ms / 1,241KB -27% / -15%
DetectLanguage (10 files, 0 allocs)    ~2µs            ~500ns          ~4x faster
DetectLanguageUnknown (0 allocs)       ~2µs            ~70ns           ~28x faster
Walk DFS (time / allocs)               70ns / 2        45ns / 0        -36% / -100%
ParserPool concurrent throughput       —               1.1ms/op        (new benchmark)
```

### `query.go` — capture arena
Query execution allocated a `[]Capture` per match (4,513 allocs/call for a typical Go highlight query). A slab arena pre-allocates capture storage once per `Execute` call and slices into it per match, reducing allocs to 29.

### `language.go` / `parser_dfa_token_source.go` — LoadLanguage buffer pre-sizing
Reads the gzip ISIZE trailer (last 4 bytes) to pre-size the decompression buffer, eliminating repeated `bytes.Buffer` growth. -27% time, -15% memory for grammar blob loading.

### `parser_dfa_token_source.go` — TypeScript scanner scratch
The TypeScript external scanner allocates a `[]bool` masked-states scratch slice on every GLR retry. Storing it as a field on `dfaTokenSource` and preserving it across pool resets eliminates 3,002 allocs/parse (3,016 → 14).

### `grammars/registry.go` — O(1) extension index + `sync.RWMutex`
`DetectLanguage` previously did an O(n²) scan across 206 languages × extensions per call. A lazy `extIndex map[string]*LangEntry` — built once after the registry stabilises and invalidated on `Register` — makes extension lookup a single map lookup. Suffix matching uses a stack-allocated `[4]string` array (zero heap allocs). ~4–28x faster for the common case.

A `sync.RWMutex` now protects `registry`, `highlightInheritanceResolved`, `extIndex`, and `extensionAliases`. An `ensureReady()` helper calls `ensureBuiltinLanguagesRegistered()` before acquiring any lock (preventing the deadlock that would occur if `Register` were called while holding the lock), then uses a double-checked write lock for lazy metadata resolution. Readers hold `RLock()` for concurrent access.

### `highlight.go` — reuse match buffer
`Highlighter.highlightTree` now calls `ExecuteInto` with a cached `matchBuffer []QueryMatch` field instead of `Execute`, eliminating the per-call backing array allocation.

### `walk.go` — pool traversal stack, index-based child access, panic-safe defer
`Walk` allocated a `[]entry` stack on every call and called `n.Children()` per node (allocating a `[]*Node` slice). Replace with a `sync.Pool`-backed stack reused across calls and `Child(i)` index access. The pool return uses a `defer` closure that captures `stack` by reference — ensuring the pool always receives the grown capacity and the stack is returned even if the callback panics. -36% time, -100% allocs.

### New benchmarks
`BenchmarkParserPoolSerial` and `BenchmarkParserPoolConcurrentThroughput` cover the pool checkout→parse→release path. `BenchmarkDetectLanguage` and `BenchmarkDetectLanguageUnknown` cover the extension lookup hot and cold paths. `BenchmarkLoadLanguage`, `BenchmarkQueryCompile`, `BenchmarkNodeSExpr`, `BenchmarkRewriterApply`, `BenchmarkInjectionParseFull`, and `BenchmarkInjectionParseIncremental` cover previously unmeasured critical paths.

## Pre-existing failures

`TestMultiGrammarImportPipeline/tsx` fails due to a missing `/tmp/grammar_parity/typescript/` grammar repo — confirmed pre-existing on `origin/main` before this branch.

### DIFF
--- a/benchmark_uncovered_test.go
+++ b/benchmark_uncovered_test.go
@@ -239,6 +239,47 @@ func makeMDWithGoBlocks(n int) []byte {
 	return []byte(sb.String())
 }
 
+// BenchmarkParserPoolSerial measures checkout→parse→release in a single goroutine.
+// This isolates pool overhead (sync.Pool Get/Put + applyDefaults) from parse time.
+func BenchmarkParserPoolSerial(b *testing.B) {
+	lang := grammars.GoLanguage()
+	pool := gotreesitter.NewParserPool(lang)
+	src := makeGoBenchmarkSource(benchmarkFuncCount(b))
+
+	b.ReportAllocs()
+	b.SetBytes(int64(len(src)))
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		tree, err := pool.Parse(src)
+		if err != nil || tree == nil {
+			b.Fatalf("ParserPool.Parse: %v", err)
+		}
+	}
+}
+
+// BenchmarkParserPoolConcurrentThroughput measures throughput under goroutine
+// contention — the scenario that justifies pooling over per-request allocation.
+// RunParallel drives GOMAXPROCS goroutines simultaneously; sync.Pool shines here
+// because each OS thread maintains a per-P free list, minimising cross-core
+// cache traffic on the Parser's reuse cursor and arena hint fields.
+func BenchmarkParserPoolConcurrentThroughput(b *testing.B) {
+	lang := grammars.GoLanguage()
+	pool := gotreesitter.NewParserPool(lang)
+	src := makeGoBenchmarkSource(benchmarkFuncCount(b))
+
+	b.ReportAllocs()
+	b.SetBytes(int64(len(src)))
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			tree, err := pool.Parse(src)
+			if err != nil || tree == nil {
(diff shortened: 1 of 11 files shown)


rsnodgrass:
Looks like I was working in too much isolation and this second round wasn’t as effective :(

odvcencio:
heya no worries! i think anything helps, the perf story is a constant work in progress.... by now its gonna take real use cases exposing the edges to move the needle. this round proved that theres definitely ground to gain in real workloads - immensely valuable!

odvcencio:
also thank you!!!

PR #22, opened 2026-04-03
perf: fix GLR parser unusable on large Go files (147KB pb.go: 4+ min → 420ms)

vdergachev:
## Problem

In its current state, gotreesitter is **unusable for scanning large files** — parsing a 147KB protobuf-generated Go file (`backoffice.pb.go`, 3420 lines) takes **over 4 minutes** and consumes **9+ GB of RAM**. For comparison, the C tree-sitter reference implementation parses the same file in under 1 second.

According to the README benchmarks, full-parsing a 19KB file costs ~4.2ms, which scales linearly to ~32ms for 147KB. The raw parse itself is not the bottleneck — the problem is in the **GLR stack machinery**. Go's grammar triggers GLR ambiguity on type assertions, composite literals, and generic-like syntax patterns that are extremely common in generated protobuf code. With the default `maxGLRStacks=8`, this creates exponential node allocation blowup (~33M `allocNodeFast` calls) and O(N²) pairwise stack equivalence comparisons.

### CPU profile breakdown (before fix, 3-second sample)

| Hotspot | CPU % | Root cause |
|---|---|---|
| `allocNodeFast` | 59% | Redundant zero-fill — nodes already zeroed by `arena.reset()` or `make()` |
| equiv cache + comparison | 24% | Cache thrashing (32K entries for 33M nodes) + expensive field checks before cheap rejection |
| `gssScratch.allocNode` | 11% | Non-inlineable hot path |

### 1. Remove redundant node zeroing in `allocNodeFast`/`allocNodeSlow`
Arena lifecycle guarantees that nodes are zero-valued: fresh slabs from `make()` are zero, reused slabs are cleared by `reset()`. The explicit `*n = Node{}` was a redundant write + write-barrier on every allocation (~33M calls).

### 2. Optimize GLR equivalence cache
- **4× larger cache** (32K → 131K entries) to reduce thrashing under heavy GLR workloads
- **Reorder comparisons**: move cheap field checks (symbol, byte range, children count) *before* cache lookup — trivially-false cases skip the cache entirely
- **Improve hash distribution**: mix in symbol bits for better spread with arena-sequential pointers

### 3. Split `gssScratch.allocNode` into inlineable fast path
The original function exceeded Go's inline budget. Splitting into a small fast path (current slab has space → direct allocation) and a `allocNodeSlow` fallback allows the compiler to inline the hot path.

### 4. Go-specific `maxGLRStacks=2`
Go's grammar has few genuine GLR ambiguities compared to Ruby or C#. The default cap of 8 causes exponential stack blowup on large generated files. Setting the default to 2 — consistent with the approach already used for JS/TS/Python/Rust/CSS — keeps parsing fast while the existing retry-widening mechanism handles edge cases.

## Result

| Metric | Before | After | Improvement |
|---|---|---|---|
| Wall time (147KB pb.go) | 4+ min | 420ms | ~570× |
| Allocations | 9+ GB | 837 MB | ~11× |

All existing tests pass. The fix is backward-compatible — no API changes.

### DIFF
--- a/arena.go
+++ b/arena.go
@@ -426,7 +426,7 @@ func (a *nodeArena) allocNodeFast() *Node {
 	if a.used < len(a.nodes) {
 		n := &a.nodes[a.used]
 		a.used++
-		*n = Node{}
+		// Node is already zeroed: fresh slabs by make(), reused slabs by reset().
 		return n
 	}
 	return a.allocNodeSlow()
@@ -459,7 +459,7 @@ func (a *nodeArena) allocNodeSlow() *Node {
 		a.nodeSlabCursor = i
 		a.used++
 		n := &slab.data[idx]
-		*n = Node{}
+		// Node is already zeroed: fresh slabs by make(), reused slabs by reset().
 		return n
 	}
 }
--- a/glr.go
+++ b/glr.go
@@ -352,7 +352,7 @@ func stackHash(s glrStack) uint64 {
 	return h
 }
 
-const glrNodeEquivCacheSize = 32768
+const glrNodeEquivCacheSize = 131072
 
 func (s *glrMergeScratch) beginEquivEpoch() {
 	if s == nil {
@@ -409,6 +409,8 @@ func nodeEquivCacheIndex(a, b *Node, depth int) int {
 	x := uint64(uintptr(unsafe.Pointer(a)))
 	y := uint64(uintptr(unsafe.Pointer(b)))
 	h := x ^ (y + 0x9e3779b97f4a7c15 + (x << 6) + (x >> 2))
+	// Mix in symbol to improve distribution for arena-sequential pointers.
+	h ^= (uint64(a.symbol) | uint64(b.symbol)<<16) * 0x85ebca6b
 	h ^= uint64(depth) * 0x517cc1b727220a95
 	return int(h & uint64(glrNodeEquivCacheSize-1))
 }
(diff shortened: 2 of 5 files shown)


vdergachev:
Addressed review comment from gemini-code-assist: all non-recursive field comparisons moved before cache lookup in a single block (d8a5b08). Cache is now used exclusively for the recursive children comparison.

odvcencio:
Heya, thanks so much for the PR! Appreciate you opening it and your contribution. I’ll review this in a bit. Thanks again!

odvcencio:
## Benchmark validation (same machine, same settings)

Ran a synthetic benchmark with GLR-triggering Go code (type assertions + composite literals, protobuf-style):

| Benchmark | `main` | this PR | Speedup |
|---|---|---|---|
| 100 funcs (28KB) | 215ms / 43MB / 3,467 allocs | 30ms / 20MB / 316 allocs | **7× faster, 2× less memory** |
| 500 funcs (139KB) | 57.8s / 199MB / 13,455 allocs | 499ms / 135MB / 3,002 allocs | **116× faster, 1.5× less memory** |

Settings: `-count=3 -benchtime=500ms -benchmem`, `GOMAXPROCS=20`

The 500-func case is close to the 147KB pb.go described in the PR. The 116× speedup on that workload is consistent with the claimed improvement.

## Fixes applied

I rebased onto current main (which now includes PR #21's allocation elimination work) and applied two small fixes on top:

1. **Missing audit tracking in GSS fast path** — The new fast path in `gssScratch.allocNode` skipped the `s.audit.recordGSSAlloc(n)` call that exists in the slow path. Added the audit hook so runtime audit tracking remains complete.
2. **Comment corrections** — Changed "Kept small for inlining" to "Kept small for hot-path optimization" (Go compiler reports `allocNode` at cost 424 vs budget 80, so it doesn't actually inline — the benefit is instruction cache locality). Added a comment explaining the 131072 cache size rationale.

All tests pass, CI pending on the push.

PR #28, opened 2026-04-13
feat(grammars): export DecodeLanguageBlob for out-of-tree grammar registration

odvcencio:
## Summary

Exports a single new public function, `grammars.DecodeLanguageBlob([]byte) (*gotreesitter.Language, error)`, so consumer packages can ship their own precompiled grammar blobs via `go:embed` and wire them into the registry without modifying gotreesitter itself.

This closes the one API gap preventing fully out-of-tree grammar registration. `Register`, `RegisterExternalScanner`, `RegisterExtension`, and `LangEntry.Language` were already public — the only missing piece was a public "decode these bytes into a `*Language`" entry point. All the plumbing already exists (`decodeLanguageBlobData`), it just wasn't exported.

## Why

Every niche grammar (language servers, in-house DSLs, experimental grammars) currently has to be absorbed into `grammars/` to work: blob committed under `grammar_blobs/`, scanner ported into the package, registration file added, manifest and linguist entries touched. That bloats gotreesitter, couples grammar release cadence to the runtime's release cadence, and forces every consumer to carry every grammar.

With this export, a language-server author can:

1. Commit their `.bin` blob in **their own** repo.
2. Hand-port the external scanner to Go in **their own** package.
3. Call `grammars.Register` + `grammars.RegisterExternalScanner` from an `init()` in **their own** package.
4. Import for side effects and the grammar is live — `DetectLanguage(\"foo.qml\")`, `AllLanguages()`, and highlight-query resolution all see it exactly as if it were built in.

No changes to gotreesitter, no waiting for a release, full control of their own grammar.

## What's in the PR

- `grammars/blob_decode.go` — the exported wrapper (~30 lines, including doc comment). Thin delegate to the existing internal `decodeLanguageBlobData` with empty-input guard.
- `grammars/blob_decode_test.go` — TDD-authored tests:
  - `TestDecodeLanguageBlob_Roundtrip` — decodes a real blob via `BlobByName`, verifies `SymbolCount`/`TokenCount` populated, then actually parses Go source through `gotreesitter.NewParser` to confirm the result is end-to-end usable.
  - `TestDecodeLanguageBlob_Garbage` — rejects non-blob bytes cleanly (error, nil language, no panic).
  - `TestDecodeLanguageBlob_Empty` — rejects nil and empty input.
- `grammars/example_external_grammar_test.go` — executable godoc example (`ExampleDecodeLanguageBlob_externalGrammar`) showing the full consumer pattern end-to-end. Has `// Output:` so `go test` runs it, which keeps the documented pattern verified in CI and makes sure it never rots.

## How it works in prod

The example file documents the full recommended shape for a consumer package (e.g. a QML language server):

\`\`\`go
package qmlgrammar

import (
    _ \"embed\"
    \"sync\"

    \"github.com/odvcencio/gotreesitter\"
    \"github.com/odvcencio/gotreesitter/grammars\"
)

//go:embed qmljs.bin
var qmljsBlob []byte

func init() {
    // 1. Register the hand-ported Go external scanner first so it is
    //    attached when the Language is decoded on first access.
    grammars.RegisterExternalScanner(\"qmljs\", newQMLScanner())

    // 2. Register the language with a lazy loader that decodes the
    //    embedded blob on demand.
    grammars.Register(grammars.LangEntry{
        Name:           \"qmljs\",
        Extensions:     []string{\".qml\"},
        Language:       loadQMLJS,
        GrammarSource:  grammars.GrammarSourceTS2GoBlob,
        HighlightQuery: qmljsHighlights,
    })
}

var loadQMLJS = sync.OnceValue(func() *gotreesitter.Language {
    lang, err := grammars.DecodeLanguageBlob(qmljsBlob)
    if err != nil {
        panic(\"qmljs grammar blob is corrupt: \" + err.Error())
    }
    return lang
})
\`\`\`

Any binary that imports `qmlgrammar` for side effects now has full QML support — detection, parsing, highlighting — with zero gotreesitter-side code.

## Relationship to #27

This is the alternative to PR #27 (\"feat(grammars): add QML support via tree-sitter-qmljs\"). Rather than absorbing a 407-line scanner + blob + registration + manifest/linguist touches into gotreesitter, this exposes the one missing API and lets the QML work live in [cushycush/qml-language-server](https://github.com/cushycush/qml-language-server) where it belongs. Plan is to close #27 with a link to this PR and thanks to the author for surfacing the gap.

## Parser-lane safety

Additive-only. Zero changes to parser, lexer, scanner runtime, existing registered grammars, or any blob. New file, three tests, one example. Cannot regress parser correctness or any existing parity gate.

## Test plan

- [x] `go test ./grammars/ -run TestDecodeLanguageBlob -count=1 -v` — all pass
- [x] `go test ./grammars/ -run ExampleDecodeLanguageBlob -count=1` — executable example passes (prints \"parsed: true\")
- [x] `go build ./grammars/...` — clean
- [x] `go vet ./grammars/` — clean (no new issues)

### DIFF
--- a/grammars/blob_decode.go
+++ b/grammars/blob_decode.go
@@ -1,0 +1,31 @@
+package grammars
+
+import (
+	"errors"
+
+	"github.com/odvcencio/gotreesitter"
+)
+
+// DecodeLanguageBlob decodes a gzip+gob encoded grammar blob (as produced by
+// the ts2go pipeline and stored under grammars/grammar_blobs/*.bin) into a
+// ready-to-use *gotreesitter.Language.
+//
+// This is the public entry point for consumer packages that ship their own
+// grammar blobs via go:embed rather than having the grammar absorbed into
+// gotreesitter's built-in registry. The expected usage pattern is to pair
+// DecodeLanguageBlob with [Register] or [RegisterExternalScanner] from an
+// init() function in the consumer package, so the grammar becomes available
+// through [DetectLanguage], [DetectLanguageByName], and [AllLanguages] without
+// modifying gotreesitter itself.
+//
+// The input bytes are consumed once; the returned Language is safe to cache
+// and reuse across parsers. Returns an error if the input is empty or cannot
+// be decoded (wrong format, truncated data, incompatible schema).
+//
+// See the package example for an end-to-end consumer integration.
+func DecodeLanguageBlob(data []byte) (*gotreesitter.Language, error) {
+	if len(data) == 0 {
+		return nil, errors.New("gotreesitter: DecodeLanguageBlob called with empty data")
+	}
+	return decodeLanguageBlobData("<external>", data)
+}
--- a/grammars/blob_decode_test.go
+++ b/grammars/blob_decode_test.go
@@ -1,0 +1,69 @@
+package grammars_test
+
+import (
+	"testing"
+
+	"github.com/odvcencio/gotreesitter"
+	"github.com/odvcencio/gotreesitter/grammars"
+)
+
+// TestDecodeLanguageBlob_Roundtrip exercises the public pathway a consumer
+// package uses: fetch blob bytes (here from the built-in store, but in prod
+// typically from the consumer's own go:embed), decode into a *Language, and
+// verify the result is usable.
+func TestDecodeLanguageBlob_Roundtrip(t *testing.T) {
+	blob := grammars.BlobByName("go")
+	if len(blob) == 0 {
+		t.Fatal("precondition: BlobByName(go) returned empty blob")
+	}
+
+	lang, err := grammars.DecodeLanguageBlob(blob)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


odvcencio:
Closing — strategic direction is to invest in grammargen parity rather than grow the out-of-band ts2go blob path. The grammargen route (`ImportGrammarJSON` + `RegisterExtension`) already covers every greenfield out-of-tree grammar use case we currently have, so exporting `DecodeLanguageBlob` isn't load-bearing right now.

The export itself is still a reasonable idea for a future change — if and when we hit a grammar where grammargen genuinely can't catch up and a consumer needs to ship a precompiled ts2go blob out-of-tree, this wrapper will get pulled in then. Leaving branch `feat/export-decode-language-blob` (commit `baf3a43`) intact for that moment.

PR #30, opened 2026-04-16
fix(parser): avoid pool aliasing for recovery token sources

rasros:
## What does this PR do?

Fixes a data race in concurrent Parse calls on Go source files. Recovery token
sources created by dfaReparseFactory were drawn from the global
dfaTokenSourcePool; a token source released at the end of one recovery
iteration could be acquired by another goroutine while the first was still
reading it, causing a race on glrStates and a nil-pointer dereference.

## Why this approach?

Add newDFATokenSourceDirect which allocates a dfaTokenSource directly
(setting noPool: true) instead of drawing from the pool. Close skips
dfaTokenSourcePool.Put when noPool is set. Only the recovery reparse path
uses direct allocation; all other acquisition sites continue to use the pool.

The alternative (a per-parser pool scoped to a single Parse call) avoids
the GC pressure of direct allocation but is more invasive. Since recovery
reparsing is not on the hot path, direct allocation is the simpler fix.

## Correctness

- [ ] If grammargen or parity logic changed, ran the smallest relevant focus target in Docker

Verified with a race-detector repro: [rasros/gotreesitter-race-repro](https://github.com/rasros/gotreesitter-race-repro), a concurrent Go-file parser that uses [lx](https://github.com/rasros/lx) to stream source files and hammer Parse from multiple goroutines. go run -race main.go triggered the race reliably before this fix; no race after. Full go test -race ./... passes. Docker parity not run; this change only affects token source lifetime, not parse table logic.

## Performance

Not a perf change. Recovery reparsing is not on the hot path; the extra allocation per recovery call has negligible impact.

## Maintainability

## Self-review

- [ ] Called out anything I'm unsure about or want a second opinion on

## Due diligence

- [ ] If adding a grammar or scanner: verified against upstream C output

### DIFF
--- a/parser_api.go
+++ b/parser_api.go
@@ -64,7 +64,7 @@ func (p *Parser) dfaReparseFactory() normalizationTokenSourceFactory {
 	}
 	return func(source []byte) (TokenSource, error) {
 		lexer := NewLexer(p.language.LexStates, source)
-		return acquireDFATokenSource(lexer, p.language, p.lookupActionIndex, p.hasKeywordState), nil
+		return newDFATokenSourceDirect(lexer, p.language, p.lookupActionIndex, p.hasKeywordState), nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


odvcencio:
thanks for the contribution!

PR #54, opened 2026-04-22
feat(c_lexer): support `#embed` directive (C23 and C++26 features)

sortA0329:
## What does this PR do?

C/C++ lexer bridge now accepts `#embed` directive lines and `__has_embed(...)` conditional feature-test forms (including  parameter variants) without parse errors.

NOTE: The `#embed` directive is not yet supported in the original `tree-sitter`. This PR implements it ahead of upstream to support modern C23/C++26 codebases in Go. Please note that if upstream introduces a different AST structure in the future, this implementation may need to be adjusted for continued parity.

## Why this approach?

The previous C/C++ lexer was unable to process the `#embed` directive correctly.

## Correctness

- [x] Affected parity validation ran in Docker, one language or grammar at a time
- [x] If grammargen or parity logic changed, ran the smallest relevant focus target in Docker

Although `bash cgo_harness/docker/run_single_grammar_parity.sh cpp` shows failures, I've verified that the error count and specific failure points remain identical to the `main` branch.

## Performance

- [x] If perf-relevant, used the stable bench settings (`GOMAXPROCS=1`, `-count=10`, `-benchtime=750ms`, `-benchmem`)
- [x] If large-grammar behavior changed, checked bounded RSS, timeout, or OOM behavior under Docker

Performance is not goal of this PR.

## Maintainability

## Self-review

Review your own diff before requesting review. Walk through it as if you're seeing it for the first time, and keep any self-review comments concise and focused on the non-obvious parts.

- [x] Called out anything I'm unsure about or want a second opinion on

## Due diligence

- [x] If adding a grammar or scanner: verified against upstream C output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
-- Nothing yet.
+### Added
+- C/C++ lexer bridge now accepts `#embed` directive lines and
+  `__has_embed(...)` conditional feature-test forms (including parameter
+  variants) without parse errors.
 
 ## [0.15.2] - 2026-04-21
 
--- a/grammars/c_lexer.go
+++ b/grammars/c_lexer.go
@@ -1291,7 +1291,8 @@ func (ts *CTokenSource) scanOpaqueQuoted(delim byte) {
 
 func isPreprocOpaqueBuiltin(text string) bool {
 	switch text {
-	case "__has_include", "__has_include_next", "__has_embed":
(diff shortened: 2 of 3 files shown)


odvcencio:
Heya thanks so much for the contribution! 

quick note on the upstream-divergence caveat: gotreesitter's AST is authoritative, so in future releases, we wont be treating the C tree-sitter grammars as doctrine— they're just one of potentially many reference implementations, and future releases won't gate on parity. Users who want a different shape can bring their own `grammar.json` (or write in Go) via `ImportGrammarJSON`— thats the extension pattern this repo is trying to grow with.

PR #55, opened 2026-04-23
recover: re-land #48 (arena retention) + #49 (repo cleanup) after accidental reset

odvcencio:
## What happened

Main was hard-reset on 2026-04-21 (reflog: `reset: moving to release/v0.15.2`), silently dropping two already-merged PRs:

- #48 — `perf(arena): lower retention constants to reduce concurrent RSS` (vdergachev, merged 2026-04-18)
- #49 — `chore: clean repo layout and unused helpers` (merged 2026-04-18)

GitHub still marks both PRs as MERGED (original merge-commit oids are `43d6341` and `630e4e3`), but those commits no longer exist in main's linear history. Content verified missing: `arena.go` still has the pre-#48 node-count constants, `cgo_harness/pure_c/samples/*.txt` still exists under the old path.

## What this PR does

Cherry-picks both commits onto current main. Original authorship preserved via `git cherry-pick` (author field keeps vdergachev for #48). New hashes are assigned — the old #48 / #49 PR pages will continue to reference the orphan hashes in their "merge commit" link, but the content is back where it belongs.

## Conflicts resolved

- `CHANGELOG.md` — merged the v0.15.2 Unreleased section (from #49) with the PR 54 `#embed` Added entry that landed after the reset. Both are now under Unreleased in the correct order.
- `grammars/c_lexer.go` — auto-merged. Both #49 and PR 54 independently removed the dead `preprocDirectiveState` / `preprocEndToken` methods; PR 54's opaque-preproc-arg additions are intact (34 marker hits confirmed post-merge).

## Validation

- `go build ./...` passes in Docker (clean exit).
- Will validate via CI once this PR is open.

## Follow-up

Once this lands, PR #50 (ts2go parallelization) rebases cleanly on main — its 6000-line phantom diff collapses to the actual ~3-file change.


odvcencio:
Validated gemini's three findings:

**Finding 1 (parser_limits.go:70, HIGH)** — False alarm on close inspection. `sourceLen` is used as a loose node-count upper bound ("≤1 node per byte"), not as a direct byte-size. The sister function `parseFullArenaInitialNodeCapacity` uses `sourceLen / 4` (empirical ratio); `parseFullArenaHintLimit` intentionally sits *above* that to leave headroom for same-size hints. Confirmed via `TestParseFullArenaNodeCapacityKeepsUsefulSameSizeHint` which depends on `initial < limit`; applying gemini's suggested `nodeCapacityForBytes(sourceLen)` conversion would collapse the limit (128KB → ~1400 nodes, below the ~32768 initial) and break the hint progression. Added a clarifying comment in 44cb092 to prevent future confusion.

**Finding 2 (parser_result_c_go_cobol_test.go:24, MEDIUM)** — Valid nit, not a bug. `normalizeCTranslationUnitRoot` is a sub-step that `normalizeCCompatibility` (the dispatcher entry) calls alongside 8 other C-specific normalizations. The tests exercise the sub-step in isolation for unit coverage. A complementary dispatcher-level integration test would widen coverage but is orthogonal to this recovery PR.

**Finding 3 (parser_result_c_go_cobol_test.go:54, MEDIUM)** — Same pattern as #2. `normalizeGoSourceFileRoot` is called by `normalizeGoReturnedTreeCompatibility` (the dispatcher entry) alongside `normalizeGoCompatibility` + `normalizeRootEOFNewlineSpan`. Test targets the sub-step.

No content changes needed for this recovery PR beyond the comment clarification. Full dispatcher-level integration coverage is a separate follow-up.

PR #64, opened 2026-05-05
fix(csharp): bound recovery on pathological inputs

qiangli:
## Summary

Three independent bugs in the C# recovery + retry path let a single file containing pathological constructs (e.g. partial classes with CJK identifiers in XML doc comments) blow past `SetTimeoutMicros` and allocate gigabytes of heap. This PR adds bounded loops + propagates the parser timeout into the recovery sub-parses.

### 1. `parseWithSnippetParser` ignored the parent parser's timeout
`acquireSnippetParser` → `resetSnippetParser` (parser.go) sets `timeoutMicros = 0` and `cancellationFlag = nil`. So when `csharpRecoverNamespaceFromChildren` triggers a snippet sub-parse via `parseWithSnippetParser`, the inner parse has no timeout — even if the outer caller configured one via `Parser.SetTimeoutMicros`.

**Fix**: new `parseWithSnippetParserTimed(lang, source, timeoutMicros)` that propagates the parent timeout. The old `parseWithSnippetParser` is preserved as a pass-through so other call sites are untouched.

### 2. C# namespace recovery had no per-file iteration cap
`normalizeCSharpRecoveredNamespaces` walks every top-level `ERROR` / `global_statement` / `statement` node in the root and triggers a fresh snippet sub-parse per match. On files with many error nodes (e.g. a `partial class Logic贴合` whose initial parse failed in multiple places due to mixed-language identifiers), this loop runs dozens of times — each iteration is a full snippet parse.

**Fix**: cap recoveries at `csharpMaxNamespaceRecoveries = 32` per top-level parse. This mirrors the existing `csharpMaxTopLevelChunkRecoverySpans = 128` cap at the top of the same file. Beyond 32 recoveries the parse is so malformed that more recovery rarely helps, but reliably amplifies cost.

### 3. `retryFullParse` didn't honor `SetTimeoutMicros` between iterations
The timeout is checked inside the GLR parse loop, but `retryFullParse` runs up to four sequential retry passes (initial-merge, node-limit-override, secondary-node-limit, final merge-per-key). Each pass is its own `parseInternal` call with its own `parseStart`. So a 10-second `SetTimeoutMicros` budget could spend 40+ seconds before returning.

**Fix**: track wall-clock from the start of `retryFullParse` and bail with the best tree so far once `p.timeoutMicros` is exceeded.

## Verification

- All existing short tests in the root package pass (`go test -count=1 -short -timeout 600s .` → ok).
- A real-world drop of 330 customer C# files (mix of standard code, WinForms designer files, and partial classes with CJK identifiers) that previously OOM-killed the process at 12+ GB RSS now completes with bounded memory after combining these patches with a small per-file timeout on the calling side.
- The library tests for `ParseStopMemoryBudget` and the snippet-parser round-trip continue to pass.

## What this doesn't fix (separate concern)

A single `parseInternal` invocation can still allocate hundreds of MB during one outer iteration on highly ambiguous inputs (the GLR algorithm forks stacks during `mergeStacksWithScratch` and `stackEntryNodesEquivalentFrontierWithScratch`). The existing `parseMemoryBudget` infrastructure is wired up correctly at `parser.go:1241` (`arena.setBudget(parseMemoryBudget(len(source)))`) so each pass is bounded by the configured 512 MB ceiling, but multiple retries × multiple snippet recoveries can multiply this. Filing this as a separate concern — likely needs either a parser-wide cumulative budget or tighter inner-loop allocation checks.

## Diff stats

- `parser_api.go` — new `parseWithSnippetParserTimed` (+11 lines)
- `parser_result_csharp.go` — namespace-recovery cap + timeout propagation (+30 / -11 lines)
- `parser_retry.go` — retry deadline (+30 lines)

Total: 3 files, +71 / -11 lines. No public API changes; existing callers unaffected.

### DIFF
--- a/parser_api.go
+++ b/parser_api.go
@@ -98,11 +98,22 @@ func (p *Parser) parseForRecovery(source []byte) (*Tree, error) {
 }
 
 func parseWithSnippetParser(lang *Language, source []byte) (*Tree, error) {
+	return parseWithSnippetParserTimed(lang, source, 0)
+}
+
+// parseWithSnippetParserTimed is parseWithSnippetParser with a per-parse
+// timeout. Used by recovery paths (e.g. csharpRecoverNamespaceFromChildren)
+// to inherit the parent parser's timeout — without this, the snippet pool
+// resets timeoutMicros to 0 and recovery sub-parses could run unbounded.
+func parseWithSnippetParserTimed(lang *Language, source []byte, timeoutMicros uint64) (*Tree, error) {
 	parser := acquireSnippetParser(lang)
 	if parser == nil {
 		return nil, ErrNoLanguage
 	}
 	defer releaseSnippetParser(parser)
+	if timeoutMicros > 0 {
+		parser.timeoutMicros = timeoutMicros
+	}
 	return parser.Parse(source)
 }
 
--- a/parser_api_internal_test.go
+++ b/parser_api_internal_test.go
@@ -2,6 +2,7 @@ package gotreesitter
 
 import (
 	"testing"
+	"time"
 )
 
 type parserTestUnsafeExternalScanner struct{}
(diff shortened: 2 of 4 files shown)


qiangli:
## Update — root cause is deeper than these patches

After landing the three patches in this PR locally and re-testing on the same customer drop (330 C# files), we found memory was still ballooning to 12+ GB on individual files like a 9 KB `ModbusTcp.cs`. The patches do fix what they claim — recovery propagation, recovery-iteration cap, retry deadline — but the GLR machinery in `parseInternal` itself can allocate gigabytes during a *single* outer iteration on inputs the C# grammar gets stuck on (the `csharpRecoverNamespaceFromChildren` / `retryFullParseWithDFA` recovery paths that show up at the top of `sample(1)`'s output).

For comparison on the same files via the official `tree-sitter` C library + `tree-sitter-c-sharp` grammar (`tree-sitter parse <file>`):

| File | Size | gotreesitter v0.15.3 | gotreesitter + this PR | Official tree-sitter |
|---|---|---|---|---|
| `ModbusTcp.cs` | 9 KB | OOM (12 GB RSS, killed at 2 min) | OOM (10 GB RSS) | **2 ms** (1 ERROR node, valid tree) |
| `Math2.cs` | 42 KB | OOM | OOM | **9 ms** |
| `Logic贴合.cs` | 87 KB | OOM | OOM | **8 ms** |
| `FormHome.Designer.cs` | 220 KB | 41 s | (didn't reach) | **21 ms** (via `smacker/go-tree-sitter`) |

So we ended up switching off `gotreesitter` to `smacker/go-tree-sitter` (CGo to the official C library); the same workload now finishes in **1.6 seconds** for the full station.

I'm leaving this PR open because the three fixes are still real, separately-valuable bug fixes — the timeout-in-snippet-parser propagation in particular feels like an obvious leak even outside the C# context. But the PR alone won't make these inputs parse cleanly. The deeper issue lives in either the GLR fork-explosion control (when many stacks survive a merge step) or the C# grammar's recovery cost on Unicode-heavy ambiguous regions. Happy to dig further if it'd be useful — let me know whether you'd like me to:

- (a) close this PR and re-open targeted ones for the three fixes individually,
- (b) keep this as-is for the small wins,
- (c) take a swing at the deeper issue with guidance on which path you'd consider acceptable (per-iteration arena allocation cap? Per-stack-merge timeout poll? Something else?).

odvcencio:
heya thanks so much for the PR and for highlighting the deeper issue- im going to merge this with a follow-on im investigating now. Thanks again!

PR #89, opened 2026-05-28
perf(intern): 1-entry hot-path cache for leaf intern lookups

odvcencio:
## Summary

Adds a 1-entry hot-path cache to the leaf interning lookups introduced in
the Phase 2/3 substitution path. When the shift loop hits the same
`(symbol, span, parseState, preGotoState, flags)` tuple in immediate
succession — typical for GLR forks processed back-to-back at the same
token — the table hash + probe is skipped in favor of a single key
compare.

## Honest bench framing

This PR is **not a JS-ratio improvement**. The xC ratio is essentially
flat (5.95x → 5.91x on JS large, `count=30`). The real signal is
narrower: a measurable reduction in GLR merge time.

| Metric | Baseline (sub off) | Sub on + hot-cache | Delta |
|---|---:|---:|---:|
| Full Go | 408.6ms | 313.4ms | -23% (mostly thermal — see note) |
| Full C | 68.7ms | 53.0ms | -23% (same thermal effect) |
| **Full xC ratio** | **5.95x** | **5.91x** | **~flat** |
| `glr_merge` | 107.3ms | 73.7ms | **-31% relative** |
| `glr_merge / parser_loop` | 29.5% | 25.9% | -3.6 pp |

The 23% wall drop on BOTH Go and C reflects machine thermal/load
variance between back-to-back 30-minute runs, not the optimization.
The honest delta is the per-bucket attribution: **interning collapses
about a third of GLR merge work** by making canonical leaves
pointer-equal, so `lookupNodeEquivCache` short-circuits at the cheap
`a == b` check (`glr.go:1516`).

## Why merge — but not headline — improves

Per-shift hash + table lookup overhead approximately offsets the merge
savings on net wall time. The hot-path cache trims that overhead
materially: in test runs against the real corpus, the cache absorbs
~75% of lookups (210k → 55k table lookups for JS large), leaving the
table to handle only the first lookup per unique key.

For workloads where GLR merge is the dominant cost bucket (large JS,
Python, anything with high parser fork count), this is a usable
optimization. For workloads where merge is already small (Java, C),
substitution is roughly neutral or a slight loss.

## Gates

All behavior is opt-in via env vars; default builds pay nothing:

- `GOT_PARSE_INTERN_LEAVES_OBSERVE=1` — measurement only, no behavior change
- `GOT_PARSE_INTERN_LEAVES_SUBSTITUTE=1` — actually substitute canonical leaves

Both also have runtime toggles for benches:
`SetInternLeavesObserveEnabled(bool)`, `SetInternLeavesSubstituteEnabled(bool)`.

## Verification

- `TestInternLeafSubstitutionParityMatrix` parses 4 real-corpus files
  (JS large/small, C large, Python large) twice — substitution off then
  on — and asserts byte-identical `(type, span, childCount, named,
  hasError)` tuples at every node. Passes.
- `TestInternTable` covers insert/lookup, hash collision resolution,
  children pointer identity, growth past load factor, and reset
  semantics. Passes.

## Background

This is the bench-validated piece of a multi-phase GLR node interning
exploration. Documented in
`spore.2026-05-28.cedar.node-interning-design`,
`spore.2026-05-28.cedar.node-interning-phase2`,
`spore.2026-05-28.cedar.node-interning-phase3`,
`spore.2026-05-28.cedar.interning-branch-reorg` in
`hypha://m31labs/gotreesitter`.

The honest assessment in the trace `trace.2026-05-28.cedar.5a30` is that
single-session tactical interning hits its ceiling here — Phase 5+
(parent interning at reduce time, grammar-level GLR fork reduction)
is multi-week work and out of scope for this PR.

## Test plan

- [ ] `go test ./...` passes
- [ ] `TestInternLeafSubstitutionParityMatrix` passes
- [ ] Confirm default-off behavior matches main on a representative
      corpus (any bench, no env vars)
- [ ] Confirm sub-on path produces byte-identical trees on JS, TS, C,
      Python (covered by the parity matrix test)

### DIFF
--- a/arena.go
+++ b/arena.go
@@ -90,6 +90,15 @@ type nodeArena struct {
 	// canonically substituted; the counter is needed to compute the
 	// "safe to dedup" leaf population (total leaves minus shift leaves).
 	internShiftLeafObserved uint64
+	// internLeafLastKey + internLeafLastHit + internLeafLastValid form a
+	// 1-entry hot-path cache for canonical-leaf lookups. When the shift
+	// loop hits the same (sym, span, state, flags) tuple in immediate
+	// succession — typical for GLR forks processed back-to-back at the
+	// same token — the table hash and probe walk are skipped. Reset on
+	// arena.reset(). Valid only when internLeafLastValid is true.
+	internLeafLastKey   internKey
+	internLeafLastHit   *Node
+	internLeafLastValid bool
 
 	nodeSlabs                       []nodeSlab
 	nodeSlabCursor                  int
@@ -499,6 +508,9 @@ func (a *nodeArena) reset() {
 	a.internLeaves.reset()
 	a.internLeavesFull.reset()
 	a.internShiftLeafObserved = 0
+	a.internLeafLastValid = false
+	a.internLeafLastHit = nil
+	a.internLeafLastKey = internKey{}
 }
 
 func (a *nodeArena) resetPrimaryNodes() {
--- a/intern.go
+++ b/intern.go
@@ -319,21 +319,23 @@ func observeLeafInternFull(arena *nodeArena, n *Node) {
 // On miss, returns nil; the caller allocates and calls storeCanonicalLeaf
 // afterward.
 func lookupCanonicalLeafKey(arena *nodeArena, key internKey) *Node {
+	// Hot-path 1-entry cache. GLR forks at the same parse position shift
+	// the same token in immediate succession; this cache skips the table
+	// hash + probe in that common case.
+	if arena.internLeafLastValid && arena.internLeafLastKey == key {
+		return arena.internLeafLastHit
+	}
 	if arena.internLeavesFull == nil {
 		arena.internLeavesFull = newInternTable()
 		return nil
 	}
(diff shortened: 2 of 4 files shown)


odvcencio:
Closing as superseded by #90.

The 1-entry intern hot-path cache here is net-neutral on the JS wall (5.95x → 5.91x, flat); its value was purely as a merge-cost reducer (~-31% relative glr_merge time). #90 (JS GLR fork reduction, 5.95x → 4.41x) attacks fork *count* directly and is the actual JS ratio lever — it makes the per-fork optimization this PR targets largely moot.

This branch also currently carries a byte-identical copy of #90's fork-reduction commit (9ef6330a == #90's 1199c682), so landing both would double-apply / conflict on the parser.go conflict resolver.

The interning design and phase work is preserved (design + phase spores in the hyphae m31labs/gotreesitter space). If we later want the leaf-intern cache as an opt-in, default-off (GOT_PARSE_INTERN_LEAVES_SUBSTITUTE=1) merge-cost reducer, we can revive it as a clean intern-only PR rebased on post-#90 main — without the duplicated fork-reduction commit.

PR #90, opened 2026-05-28
perf(parser): cut JS GLR forks — 5.95x → 4.41x via 4 conflict-state resolutions

odvcencio:
## Summary

Resolves the four JavaScript parse-table states that produce the vast majority of GLR forks, matching tree-sitter C's deterministic behavior. The first two states (statement / class-member ASI boundaries) cut total GLR forks on `text-editor-component.js` from **63,688 → 7,549 (-88%)** and improved the Go-vs-C full-parse ratio from **5.95x → 4.72x**. Two further comma-boundary states (`variable_declaration_repeat1`, `object_repeat1`) extend the reduction to a final **5.95x → 4.41x**. Byte-for-byte parity with tree-sitter C is preserved.

## Root cause

Profiling (`parse_gap_report -hot-shapes`) found ~95% of JS forks come from two ASI (automatic semicolon insertion) boundary states; two additional comma-separator boundaries fork on every multi-declarator / multi-property list:

| State | Lookahead | Conflict | Notes |
|---|---|---|---|
| 985 | `identifier` | reduce `class_body_repeat1` vs shift | class-member boundary (~63% of forks) |
| 9 | `this` | reduce `program_repeat1` vs shift | statement boundary (~32% of forks) |
| 1335 | `,` | reduce `variable_declaration_repeat1` vs shift | `var a, b, c` comma separator |
| 1417 | `,` | reduce `object_repeat1` vs shift | `{a, b, c}` comma separator |

Tree-sitter C does **not** fork at the equivalent cells — they are single-action (`count=1`) shifts, with the repeat reduce gated onto the separate `_automatic_semicolon` / separator lookahead. We were forking where C resolves deterministically.

## Fix

Extend the existing `javascriptRepetitionShiftConflictChoice` resolver (`parser.go`) to resolve all four states to the repetition shift — exactly what C does. Gated to `lang.Name == "javascript"` (dispatch in `parseInternal`) and shape-guarded by `repetitionShiftConflictChoice`, which only fires on a `{repetition-shift, reduce}` conflict, so these states in other grammars are untouched.

## Validation

- **Docker C-parity:** curated parity suite (all languages) **0 failures**; `TestParityFreshParse/javascript` + `TestParityHasNoErrors/javascript` pass byte-for-byte against tree-sitter C. Re-validated on the final 4-state change via the exhaustive 12-lang ring matrix.
- **Fork reduction:** `TestJSForkReductionParity` confirms a clean parse and logs the fork drop.
- **Bench (2-state milestone):** JS full-parse Go median 311ms vs C 65.9ms = **4.72x** (from a 5.95x baseline); per-bucket attribution `action_dispatch` -61%, `action_apply` -63%. The two comma-boundary states (1335/1417) take the final ratio to **4.41x**.

> Note: the per-bucket bench figures above were captured at the 2-state milestone (4.72x). The headline **4.41x** reflects the final 4-state change. Re-confirm the exact 4-state bench median before release if a precise number is needed.

### DIFF
--- a/js_fork_reduction_test.go
+++ b/js_fork_reduction_test.go
@@ -1,0 +1,51 @@
+package gotreesitter_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/odvcencio/gotreesitter"
+	"github.com/odvcencio/gotreesitter/grammars"
+)
+
+// TestJSForkReductionParity asserts that the extended
+// javascriptRepetitionShiftConflictChoice (resolving states 9 and 985 to the
+// repetition shift, matching tree-sitter C's deterministic behavior) does not
+// change the parse tree on the real-corpus JS files, and reports the total
+// GLR fork count so the reduction is visible in `go test -v` output.
+//
+// The parser change is unconditional (no env gate) — it extends an existing
+// deterministic conflict resolver. This test is the parity safety net: if the
+// extra shift resolutions were wrong, the tree would have errors or a
+// different shape, which we'd catch here against the structural expectations.
+func TestJSForkReductionParity(t *testing.T) {
+	cases := []string{
+		"cgo_harness/corpus_real/javascript/large__text-editor-component.js",
+		"cgo_harness/corpus_real/javascript/large__jquery.js",
+		"cgo_harness/corpus_real/javascript/small__functions.js",
+	}
+	lang := grammars.JavascriptLanguage()
+	for _, path := range cases {
+		src, err := os.ReadFile(path)
+		if err != nil {
+			t.Skipf("corpus not present: %v", err)
+		}
+		profile := gotreesitter.NewAmbiguityProfile()
+		parser := gotreesitter.NewParser(lang)
+		parser.SetAmbiguityProfile(profile)
+		tree, err := parser.Parse(src)
+		if err != nil {
+			t.Fatalf("%s: parse: %v", path, err)
+		}
+		root := tree.RootNode()
+		if root.HasError() {
+			t.Errorf("%s: tree has parse error after fork-reduction change", path)
+		}
+		var totalForks uint64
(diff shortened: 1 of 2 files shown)


odvcencio:
## Final numbers (count=30, all 4 states resolved)

Added jquery's two dominant fork states on top of the original 9/985 fix:
- State 1335: `variable_declaration_repeat1` (\`var a, b, c\`) on \`,\`
- State 1417: `object_repeat1` (\`{a, b, c}\`) on \`,\`

Both are the same `_repeat1`-reduce vs repetition-shift pattern, both C deterministic-shifts.

| Stage | JS Full xC | Go | C |
|---|---:|---:|---:|
| Baseline | 5.95x | 408.6ms | 68.7ms |
| states 9+985 | 4.72x | 311.0ms | 65.9ms |
| + jquery 1335+1417 | **4.41x** | **294.1ms** | **66.6ms** |

C-side normalizer stable (~3% spread) across all runs, so the ratio comparison holds. **Total 5.95x → 4.41x (-26% relative).** Attribution: `action_dispatch` -66%, `action_apply` -69% — fewer live stacks, less per-token work.

jquery forks 12,040 → 4,058 (-66%); text-editor-component.js 63,688 → ~2,570 (proxy metric). C-parity (`TestParityFreshParse`, all langs) passes with 0 failures.

### Scope note
Attempted extending the same pattern to Python (states 72/2309) — it **broke parsing** (clean `module` → `ERROR`) because Python statement/collection boundaries are NEWLINE/INDENT-driven, not ASI, so the reduce is correct where a blanket shift is not. Reverted. The lesson: this resolution only applies where tree-sitter C deterministic-shifts at the exact state (verified against parser.c), and must not be assumed to transfer across languages. This PR is JS-only and each of the 4 states is C-verified.

PR #112, opened 2026-06-08
Fix/go error root parity

frozen425:
## What does this PR do?

Returns `source_file` as the root node type for Go error-recovery trees, matching C tree-sitter behavior. Previously gotreesitter returned `ERROR` (65535) as the root symbol, making the entire root an ERROR node instead of `source_file`.

Also adds `TestRootShapeParity` in `cgo_harness/` — an empirical test that parses 7 inputs (valid, partially valid, fully invalid) through both C tree-sitter and gotreesitter, comparing tree shapes side by side.

## Why this approach?

C tree-sitter always uses the grammar start symbol (`source_file`) as the root, even for fully invalid input. The change follows the existing pattern already used for SQL and Swift — a one-line addition to `syntheticRootSymbol`.

**Rejected alternative**: wrapping all error-recovery children in a single ERROR node (`source_file → ERROR(children...)`). Empirical testing showed C tree-sitter does NOT do this unconditionally — for partially valid files it keeps valid top-level nodes as direct children of `source_file`. The wrapping also masked a deeper parser-level difference: C tree-sitter successfully reduces expression-like patterns into `expression_statement → call_expression(...)`, while gotreesitter produces fragments. That fragmentation is a GLR parser difference, not a root assembly issue.

## Correctness

  - `go test -run "TestParser|TestTree|TestRootNode|TestResult|TestBuildResult" -count=1` — all pass
- [x] Added `TestRootShapeParity` (cgo_harness, build tag `treesitter_c_parity`) covering 7 error scenarios
  - Recommend: `bash cgo_harness/docker/run_single_grammar_parity.sh go`
  - Deeper GLR parser fragmentation (e.g. `x.Query("..." + input)` parses as `expression_statement` in C but fragments in gotreesitter) is a parser-level difference, not addressed here. Downstream consumers (xgrep) handle this with a fragmentation guard.

## Performance

  - **Perf is not the goal.** The change adds one `isLanguage("go")` string comparison on the error-recovery path only. Normal valid-Go parses are unaffected.

## Maintainability

- [x] No unnecessary abstractions or speculative cleanup outside the PR scope

## Self-review

- [x] Called out anything I am unsure about or want a second opinion on

### Crux of the solution

**`syntheticRootSymbol` (6 lines)**: Adds Go to the list of languages that return `expectedRootSymbol` (usually `source_file`) even when `rootHasError` is true. This follows the identical pattern already established for SQL and Swift.

### Empirical findings from `TestRootShapeParity`

| Input | C tree-sitter | gotreesitter (with this fix) | Parity? |
|---|---|---|---|
| valid Go | `source_file → pkg, func` | Same | ✅ |
| `"x"` (single ident) | `source_file → ERROR(ident)` | Same | ✅ |
| `package main\nfunc broken` | `source_file → pkg_clause, ERROR(func,ident)` | `source_file → pkg_clause, ERROR(func), ident` | ❌ fragment grouping |
| `x.Query("..." + input)` | `source_file → expr_stmt(call_expr)` | `source_file → sel_expr, (, binary_expr, )` | ❌ parser-level |

Root type is now correct for all cases. Fragment grouping differences are deeper GLR parser issues tracked separately.

### Open question: generalize vs per-language

| | Generalize (always use start symbol) | Per-language opt-in (current) |
|---|---|---|
| **C parity** | Full | Partial — only opted-in languages |
| **New language coverage** | Automatic | Manual |
| **Blast radius** | Wide | Narrow — only Go in this PR |
| **Regression risk** | Python/Dart custom error handling could break | None |

**Recommendation**: Keep per-language for now. Python (`syntheticRootCanDropError`) and Dart (`dartProgramChildrenLookComplete`) have custom logic that depends on seeing specific root shapes.

## Due diligence

- [x] Read the code before changing it
- [x] Checked that similar patterns exist elsewhere (SQL, Swift) and stayed consistent
- [x] Validated against C tree-sitter output via TestRootShapeParity


odvcencio:
heya,

thanks for the contribution. I appreciate you taking the time to put this together.

odvcencio:
heya,

thanks for the PR, I appreciate the root-shape parity contribution. the useful core here is Go recovery roots staying `source_file`; could you rebase and narrow this to that root-shape behavior? if not, I may land the narrow version separately.

adrianliechti:
ciao @odvcencio

I might have tried to tackle down the same issue in parallel (with heavy help of an llm)
In bigger repos (e.g. the golang stdlib), the lib hanged (in master branch; or stack overflowed in the last release). this pull request has not fully solved it to me

my app and the tests seem to be green with ~https://github.com/odvcencio/gotreesitter/pull/120~ #121 
maybe the workaround helps to find further root causes

best, adrian

odvcencio:
Preservation update pushed to this contributor branch:

- merged current `origin/main` into the PR branch without force-pushing
- resolved the `parser_result_go_compat.go` conflict in favor of `main`'s newer iterative DFS normalizer
- appended a cleanup commit so the final PR diff is narrowed to the Go `source_file` root behavior and focused unit test
- left Gomez's original commits in the PR history

Local validation:

- `go test . -run 'TestBuildResultFromNodesKeepsGoSourceFileRootWhenChildrenHaveErrors|TestBuildResultFromNodesKeepsSQLSourceFileRootWhenChildrenHaveErrors|TestBuildResultFromNodesKeepsSwiftSourceFileRootWhenChildrenHaveErrors|TestNormalizeGoSourceFileRootRetagsRecoveredTopLevelChildren' -count=1`
- `go test . -run 'TestBuildResult|TestRoot|TestNormalizeGo' -count=1`

Issue #110, opened 2026-06-06
Parser.Parse stack-overflows (fatal, unrecoverable) on large source files instead of returning an error


townsendmerino:
**Version:** gotreesitter v0.20.1 · Go 1.26.3 · darwin/arm64 (also reproduces under the same version on linux/amd64 CI)

### Summary

`(*Parser).Parse` (via `(*ParserPool).Parse`) overflows the goroutine
stack on some large real-world source files. Because a Go stack overflow
is a **fatal runtime error**, not an `error` return, it:

- cannot be caught with `recover()`,
- bypasses the `(tree, err)` error path entirely (callers that carefully
  check `err` still crash), and
- **takes down the entire host process**, not just the parse.

### Reproduction

Two files in [spf13/cobra](https://github.com/spf13/cobra) at revision
`61968e893eee2f27696c2fbc8e34fa5c4afaf7c4` crash a Go-grammar parse:

| file | size | result |
|---|---:|---|
| `completions_test.go` | 117,138 B | **fatal stack overflow** |
| `command_test.go` | 80,115 B | **fatal stack overflow** |
| `command.go` | 61,142 B | parses fine |

Both crashers are large **table-driven test files** (long `[]struct{…}{…}`
literals with many entries). Minimal driver, adapted from how we call the
library:

```go
package main

import (
	"fmt"
	"os"

	"github.com/odvcencio/gotreesitter"
	"github.com/odvcencio/gotreesitter/grammars"
)

func main() {
	data, _ := os.ReadFile("completions_test.go") // cobra @ 61968e8
	lang := grammars.DetectLanguageByName("go").Language()
	pool := gotreesitter.NewParserPool(lang)
	tree, err := pool.Parse(data) // <-- never returns; fatal stack overflow
	fmt.Println(tree, err)
}
```

### What we ruled out — it's size/node-count, not nesting depth

We initially suspected deep nesting. It is **not**: a synthetic
10,000-deep parenthesized expression (`var x = ((((…1…))))`, ~20 KB)
parses cleanly and returns a tree. The crash correlates with overall file
size / node count of these wide table-driven literals, with the threshold
for these particular Go files sitting between 61 KB (ok) and 80 KB (crash).

### Observed stack trace (truncated, repeats deeply)

```
fatal error: stack overflow

runtime stack:
...
github.com/odvcencio/gotreesitter.(*Parser).buildResultFromNodes
	.../parser_result_root_build.go:111
[...]

PR #120, opened 2026-06-21
fix: no node is its own descendant (golang) -> replaced by #121

adrianliechti:
Fix hang/stack overflow on some Go files (#110 )
  
Some Go files (e.g. the stdlib's cmd/compile/internal/noder/reader.go, and packages like net/encoding/runtime) hang forever on main, or stack-overflow on v0.20.2.

Cause: on recovered input, the Go start-symbol reduce can leave a source_file node that has itself as a materialized child (node.children[i] == node). Raw .children looks fine, but the materialized view is cyclic, so every full tree walk loops — wireParentLinksWithScratch first, then the compat normalizers, etc. That's why it surfaces in different functions.

Fix: stripResultTreeSelfCycles, run at the top of finalizeResultRoot before anything walks the tree. It's a bounded DFS that drops any child edge pointing back at the node itself or an ancestor. Gated on root.hasError(), so clean parses are unaffected.

Verified: those files now parse with no hang/crash, full suite passes, added a unit test.
Complements #112 (which gives the proper source_file root); this is the part that stops the hang.

### DIFF
--- a/parser_result_cycle_guard.go
+++ b/parser_result_cycle_guard.go
@@ -1,0 +1,66 @@
+package gotreesitter
+
+// stripResultTreeSelfCycles enforces the structural invariant that no node is
+// its own descendant.
+//
+// On recovered input the GLR start-symbol reduce can encode a final-child-ref
+// that points back at the parent node itself — e.g. a `source_file` whose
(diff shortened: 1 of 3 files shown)


adrianliechti:
#121 seems a better approach

PR #121, opened 2026-06-21
fix: no node is its own descendant in parser.go

adrianliechti:
Fix infinite hang / stack overflow on Go files that hit trailing-EOF recovery (#110)
  
  Some Go files (e.g. the stdlib's cmd/compile/internal/noder/reader.go, and packages like net/encoding/runtime) hang forever on main, or stack-overflow on
  v0.20.2.

  Cause: the trailing-EOF recovery clones the GSS (cloneWithScratch keeps the same node payload pointers) and re-reduces, which can leave a recovered
  source_file referencing itself as a child. The raw .children looks fine, but the materialized child view is cyclic, so every later full tree walk loops —
  wireParentLinksWithScratch first, then the compat normalizers, recovery trim, etc. That's why it shows up in different functions.

  Fix: stripResultTreeSelfCycles, run in finalizeRecoveredNodes right after the recovery produces its nodes. It's a bounded DFS that drops any child edge
  pointing back at the node itself or an ancestor. It only runs on recovered trees, so normal parses are untouched.

  Verified: those files now parse with no hang/crash, full suite passes, added a unit test.

  Note: this keeps recovered trees well-formed; ideally the recovery clone/reduce wouldn't create the self-reference in the first place — happy to follow up on
  that. Complements #112 (root-symbol relabel); this is the part that stops the hang.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -2130,6 +2130,9 @@ func (p *Parser) parseInternal(source []byte, ts TokenSource, reuse *reuseCursor
 			parserLoopNanos = time.Since(parseStart).Nanoseconds()
 		}
 		materializeTransientParentNodes(nodes, arena, scratch.reduce.transientParents, scratch.reduce.transientChildren)
+		for _, n := range nodes {
+			stripResultTreeSelfCycles(n)
+		}
 		tree := p.buildResultFromNodes(nodes, source, arena, oldTree, &reuseState, &scratch.nodeLinks)
 		if root := tree.RootNode(); root != nil {
 			normalizeSQLRecoveredMissingNull(root, arena, p.language)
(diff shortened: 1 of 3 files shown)


odvcencio:
Preservation update pushed to this contributor branch:

- merged current `origin/main` into the PR branch without force-pushing
- appended a follow-up commit that keeps `children`, `fieldIDs`, and `fieldSources` aligned when cyclic child edges are stripped
- left Adrian's original commit in the PR history

Local validation:

- `go test . -run '^TestStripResultTreeSelfCycles$' -count=1`
- `go test . -run 'TestStripResultTreeSelfCycles|TestParser|TestTree|TestRootNode|TestResult|TestBuildResult' -count=1`

PR #126, opened 2026-06-28
[codex] fix recovered result tree self-cycles

odvcencio:
## Summary

- Add a recovered-tree guard that strips self-references and ancestor back-edges before result-tree construction.
- Keep `children`, `fieldIDs`, and `fieldSources` aligned when a cyclic child edge is removed.
- Add unit coverage for self-cycle removal, ancestor back-edge removal, clean-tree preservation, and field metadata alignment.

## Why

Maintainer follow-up / replacement branch for #121. It preserves the useful cycle-guard behavior from the contributor PR and incorporates the unresolved review feedback about parallel field metadata. Refs #110.

## Validation

- `go test . -run '^TestStripResultTreeSelfCycles$' -count=1`
- `go test . -run 'TestStripResultTreeSelfCycles|TestParser|TestTree|TestRootNode|TestResult|TestBuildResult' -count=1`

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -2130,6 +2130,9 @@ func (p *Parser) parseInternal(source []byte, ts TokenSource, reuse *reuseCursor
 			parserLoopNanos = time.Since(parseStart).Nanoseconds()
 		}
 		materializeTransientParentNodes(nodes, arena, scratch.reduce.transientParents, scratch.reduce.transientChildren)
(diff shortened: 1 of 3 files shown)


odvcencio:
Superseded by the contributor-preserving update pushed to #121. The original contributor commit remains in #121, with the metadata-alignment follow-up appended on top.

Issue #118, opened 2026-06-16
Swift: a comparison operator (< / > / ==) in an if/while condition fails to parse (ERROR tree, no symbols)


richardwooding:
## Summary

The Swift grammar fails to parse a function body containing a **binary comparison operator (`<` / `>` / `==`) inside an `if` / `while` condition**: the parse tree comes back with `RootNode().HasError() == true` and **no `function_declaration` node** is recovered, so symbol extraction over the file yields nothing. The same code with a non-comparison condition (`if x { … }`) parses cleanly.

This looks like the classic Swift `<` / `>` generics-vs-operator ambiguity tripping the parser. It's broad: most real Swift functions compare values in control flow, so a large fraction of real-world Swift files lose all symbols (e.g. extracting from Alamofire, only ~half the source files yielded any functions).

## Environment
- gotreesitter **v0.20.2** (latest)
- Go **1.26.2**, darwin/arm64
- Grammar: `grammars.SwiftLanguage()`

## Reproduction

```sh
mkdir swiftrepro && cd swiftrepro
go mod init swiftrepro
go get github.com/odvcencio/gotreesitter@v0.20.2
go run .
```

`main.go`:

```go
package main

import (
	"fmt"

	ts "github.com/odvcencio/gotreesitter"
	"github.com/odvcencio/gotreesitter/grammars"
)

func report(lang *ts.Language, label, src string) {
	tree, _ := ts.NewParser(lang).Parse([]byte(src))
	root := tree.RootNode()
	funcs := 0
	var walk func(*ts.Node)
	walk = func(n *ts.Node) {
		if n == nil {
			return
		}
		if n.Type(lang) == "function_declaration" {
			funcs++
		}
		for i := 0; i < n.ChildCount(); i++ {
			walk(n.Child(i))
		}
	}
	walk(root)
	fmt.Printf("%-22s hasError=%-5v function_declaration=%d\n", label, root.HasError(), funcs)
}

func main() {
	sw := grammars.SwiftLanguage()
	report(sw, "if x {}     (ok)", "func a() { if x { foo() } }")
	report(sw, "if x > 0 {} (bug)", "func a() { if x > 0 { foo() } }")
	report(sw, "if x < 0 {} (bug)", "func a() { if x < 0 { foo() } }")
	report(sw, "if x == 0 {}(bug)", "func a() { if x == 0 { foo() } }")
	report(sw, "while x > 0 (bug)", "func a() { while x > 0 { foo() } }")
}
[...]

PR #137, opened 2026-07-02
fix(swift): recover the ternary/conditional operator (#135)

### LINKED ISSUE #135: tree-sitter-swift: the ternary/conditional operator (? :) never parses — ERROR node everywhere
## Summary

With gotreesitter **v0.20.8**, the Swift ternary/conditional operator `cond ? a : b` fails to parse in **every** position I tried — the `?`/`:` are not recognised and an `(ERROR)` node is emitted (`RootNode().HasError()` is `true`). Inside a function this also degrades the enclosing `function_declaration` to `_modifierless_function_declaration_no_body`, so the whole function's structure is lost.

This is broader than the earlier Swift trailing-closure issues (#118 comparison-in-`if`, #131 `else if`): those needed a *parameter* plus a specific construct. The ternary fails even at top level and even with no comparison, so it looks like the conditional operator simply isn't wired in the grammar (or regressed) — even though a `ternary_expression` node type is referenced.

## Minimal repro

```go
ls := grammars.DetectLanguage("x.swift").Language()
pool := gotreesitter.NewParserPool(ls)

// Simplest possible — top level, no function, no comparison operator needed:
tree, _ := pool.Parse([]byte("let y = 3 > 0 ? 1 : 2\n"))
println(tree.RootNode().HasError()) // true
```

Observed s-expressions (all `HasError() == true`):

```
// let y = 3 > 0 ? 1 : 2
(source_file (value_binding_pattern) (simple_identifier)
  (_expression_without_willset_didset (comparison_expression (integer_literal) (integer_literal)) (ERROR)))

// func f() -> Int { return true ? 1 : 2 }   ← note: no parameter, no comparison
(source_file (_modifierless_function_declaration_no_body (simple_identifier) (user_type (type_identifier)))
  (ERROR) (control_transfer_statement (boolean_literal) (ERROR)))
```

The `? 1 : 2` tail is dropped into an `(ERROR)` node in every case.

## Trigger matrix

| Snippet | `HasError()` |
|---|---|
| `let y = 3 > 0 ? 1 : 2` (top level) | **true** |
| `func f() -> Int { return true ? 1 : 2 }` (no param, no comparison) | **true** |
| `func f(x: Int) -> Int { return x > 0 ? 1 : 2 }` | **true** |
| `func f(x: Int) { print(x > 0 ? 1 : 2) }` (ternary in a call arg) | **true** |
| `func f(x: Int) -> Int { if x > 0 { return 1 }; return 2 }` (no ternary) | false ✓ |

Only the ternary-free control-flow case parses cleanly — every use of `? :` errors.

## Impact

The ternary operator is extremely common in real Swift, so any function containing one loses its entire parse (no symbols, no cyclomatic/cognitive metrics for that function). Downstream analysis tools have to treat ternary-bearing files as unanalysable.

## Environment

- gotreesitter **v0.20.8**
- Go **1.26.2**, darwin/arm64
- Grammar: `grammars.DetectLanguage("x.swift")`

Found while validating the else-if fix (#131) downstream in richardwooding/file-search-on.

richardwooding:
Fixes #135.

## Problem

On the runtime Swift blob (ts2go-converted from tree-sitter-swift), the conditional operator `cond ? a : b` never fires the `ternary_expression` reduction. The `? a : b` tail is dropped into an `(ERROR)` node in **every** position — top level, `return`, call argument — and inside a function the enclosing declaration collapses to `_modifierless_function_declaration_no_body`, so the whole function's structure (and its symbols/metrics) is lost. Optional `?`, `??`, `?.` and `as?` all parse fine; only the conditional operator is missing.

```go
tree, _ := pool.Parse([]byte("let y = 3 > 0 ? 1 : 2\n"))
tree.RootNode().HasError() // was: true
```

## Why a recovery normalizer (not a grammar change)

The runtime blob is the ts2go conversion of `alex-pinkus/tree-sitter-swift`, not grammargen output, so editing `grammargen/swift_grammar.go` has no runtime effect. Regenerating the blob from the pinned grammar *does* restore ternary, but a corpus diff showed it also shifts the ts2go/GLR divergence profile and regresses other common constructs (`if let x: T = …`, `if case …`, compound `if let`), which would require re-tuning the whole Swift recovery layer — out of scope here. And unlike the trailing-closure ambiguity (#118/#123/#131), no rewrite of the ternary itself parses on this blob, so the paren-injection-and-reparse trick cannot apply.

So this follows the established recovery-normalizer approach and reconstructs the node instead.

## Approach (`parser_result_swift_ternary.go`)

1. Source-scan for ternary sites (a `?` that has a matching `:`), tracking each operand's start so the condition is bounded correctly, and skipping `??`, `?.`, postfix `x?`, and `as?`/`try?`.
2. Reparse the source with each `? if_true : if_false` tail blanked with spaces — length-preserving, so byte offsets are unchanged; the condition then parses cleanly in place.
3. Reparse each `if_true` / `if_false` operand span on its own.
4. Splice a synthesised `ternary_expression` — children `[condition, "?", if_true, ":", if_false]`, fields `condition`/`if_true`/`if_false` (the exact upstream layout) — in place of the condition node in the clean skeleton.

The rewrite is accepted **only** when the reconstructed tree is error-free and byte-faithful, so a mis-detected site can never make the result worse and non-ternary code is untouched. Chained ternaries (`a ? b : c ? d : e`) bail safely and are left unrecovered.

## Testing

- New `TestSwiftTernaryExpressionRecovers` (10 contexts: top-level `let`, `return` with/without params, call argument, string operands, nested-in-parens, parenthesised condition, enum-dot operands, multi-argument, ternary-as-if-condition) and `TestSwiftTernaryFieldsAndShape` (exact field layout).
- `go test ./grammars -run Swift` and `go test . -run 'Swift|Ternary|Recover'` pass.
- Broad `TestTop50ParseSmokeNoErrors` / `TestSupportedLanguagesParseSmoke` pass.
- Validated against tree-sitter-swift's own 220-snippet test corpus: the ternary snippets now parse cleanly, with **no regressions** on the other 218; the reconstructed tree matches upstream's expected shape (e.g. the `if a ? b : c == d {}` corpus case).

Docker parity harness was unavailable in this environment; validation used the upstream `tree-sitter` CLI on the pinned grammar as the reference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Fixed
+
+- Swift ternary/conditional operator (`cond ? a : b`) now recovers instead of
+  dropping `? a : b` into an `ERROR` node in every position. The runtime Swift
+  blob never fired the `ternary_expression` reduction, so any function containing
+  a ternary lost its whole parse (collapsing to
+  `_modifierless_function_declaration_no_body`). A post-parse recovery pass
+  reconstructs the `ternary_expression` — reparsing the source with each
+  `? if_true : if_false` tail blanked so the condition parses in place, then
+  splicing a synthesised node with the upstream `condition`/`if_true`/`if_false`
+  layout. The rewrite is accepted only when the result is error-free and
+  byte-faithful, so non-ternary code is never affected (#135).
+
 ## [0.20.8] - 2026-07-01
 
 Adds consumer-controllable forest parsing.
--- a/grammars/swift_parse_regression_test.go
+++ b/grammars/swift_parse_regression_test.go
@@ -380,3 +380,102 @@ func TestSwiftElseIfChainRecoversFunction(t *testing.T) {
 		})
 	}
 }
+
+// TestSwiftTernaryExpressionRecovers is the regression test for issue #135: the
+// conditional operator `cond ? a : b` never fired the ternary_expression
+// reduction on the runtime blob, dropping `? a : b` into an ERROR node in every
+// position and collapsing any enclosing function. The recovery normalizer now
+// reconstructs the ternary_expression in place.
+func TestSwiftTernaryExpressionRecovers(t *testing.T) {
+	lang := SwiftLanguage()
+	cases := []struct {
+		name string
+		src  string
+	}{
+		{"top-level-let", "let y = 3 > 0 ? 1 : 2\n"},
+		{"return-no-param", "func f() -> Int { return true ? 1 : 2 }\n"},
+		{"return-with-param", "func f(x: Int) -> Int { return x > 0 ? 1 : 2 }\n"},
+		{"call-argument", "func f(x: Int) { print(x > 0 ? 1 : 2) }\n"},
+		{"string-operands", "let a = cond ? \"x\" : \"y\"\n"},
+		{"nested-in-parens", "let z = a + (cond ? 1 : 2) + b\n"},
+		{"parenthesised-condition", "let availableRules = (context == nil) ? [.noContextRule] : []\n"},
+		{"enum-dot-operands", "let result: MyEnumType = condition ? .someEnumCase : .someOtherEnum\n"},
+		{"multi-argument", "let m = foo(a, b > 0 ? 1 : 2)\n"},
+		{"as-if-condition", "if a ? b : c == d {\n}\n"},
+	}
+	for _, tc := range cases {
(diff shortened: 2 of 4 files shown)


richardwooding:
Thanks — applied both suggestions in 55b8762: the ternary bounds checks now compare `uint32` byte offsets directly instead of casting to `int`, so they stay correct on 32-bit builds.

odvcencio (dismissed):

odvcencio (approved):

Issue #115, opened 2026-06-15
C#: large real-world files collapse to a single ERROR node (0 declarations extracted)


richardwooding:
## Summary

The C# grammar parses some real-world, valid C# source files into a **single top-level `ERROR` node** that spans the entire file, so no declarations are recoverable — `RootNode().HasError()` is true and a tree walk finds **0 `class_declaration` / `method_declaration` nodes**. Small C# files parse correctly, so the runtime works in general; this only shows up on certain larger files.

The prior C# recovery work (#64, #65, #98, #106) bounded the *cost* of recovery, but these files still fall into recovery and yield an empty/degenerate tree rather than a usable parse.

## Environment

- gotreesitter **v0.20.2**
- Go **1.26.2**, darwin/arm64
- Grammar: `grammars.CSharpLanguage()`

## Reproduction

```sh
mkdir gts-csharp-repro && cd gts-csharp-repro
go mod init gtsrepro
go get github.com/odvcencio/gotreesitter@v0.20.2

# A real, valid C# file from Newtonsoft.Json (pinned commit):
curl -sO https://raw.githubusercontent.com/JamesNK/Newtonsoft.Json/4f73e74372445108d2c1bda37b36e6f5e43402e0/Src/Newtonsoft.Json/JsonReader.cs

go run .
```

`main.go`:

```go
package main

import (
	"fmt"
	"os"

	ts "github.com/odvcencio/gotreesitter"
	"github.com/odvcencio/gotreesitter/grammars"
)

func report(lang *ts.Language, label string, src []byte) {
	tree, err := ts.NewParser(lang).Parse(src)
	if err != nil || tree == nil {
		fmt.Printf("%-26s parse error: %v\n", label, err)
		return
	}
	root := tree.RootNode()
	methods, classes := 0, 0
	var walk func(*ts.Node)
	walk = func(n *ts.Node) {
		if n == nil {
			return
		}
		switch n.Type(lang) {
		case "method_declaration":
			methods++
		case "class_declaration":
			classes++
		}
		for i := 0; i < n.ChildCount(); i++ {
			walk(n.Child(i))
		}
	}
	walk(root)
	fmt.Printf("%-26s hasError=%-5v  classes=%d  methods=%d\n", label, root.HasError(), classes, methods)
}

func main() {
	lang := grammars.CSharpLanguage()

	// Positive control: a small valid C# file parses correctly.
	report(lang, "small control file", []byte(
[...]

PR #138, opened 2026-07-02
fix(csharp): recover methods from large shredded namespaces (#136)

### LINKED ISSUE #136: C#: JsonTextReader.cs / JsonReader.cs still recover 0 method_declaration on v0.20.8 (follow-up to #115/#116)
## Summary

Follow-up to #115 / #116. The C# recovery work in #116 stopped these large real-world files collapsing to a single top-level `ERROR` node — they now parse to a `compilation_unit` and recover *some* declarations. But the affected files still yield **0 recoverable `method_declaration` nodes**, so downstream symbol/complexity extraction gets nothing for them.

#116 itself flagged this as out of scope ("Declarations whose header is fully shredded into ERROR fragments by the GLR failure … aren't fully reconstructed yet"). Opening this to track the remaining gap, since the practical outcome for a consumer is unchanged: 0 methods.

## Evidence (Newtonsoft.Json, pinned commit `4f73e74372445108d2c1bda37b36e6f5e43402e0`)

Measured via a downstream extractor on gotreesitter v0.20.8:

| File | Real methods (approx) | Extracted functions |
|---|---|---|
| `Src/Newtonsoft.Json/JsonTextReader.cs` | ~72 | **0** |
| `Src/Newtonsoft.Json/JsonReader.cs` | many | **0** |
| `Src/Newtonsoft.Json/JsonTextWriter.cs` | 11 | 11 ✓ |

`JsonTextWriter.cs` parses fine; the other two recover the namespace/class shell (per #116) but no methods.

## Reproduction

```sh
mkdir gts-csharp-repro && cd gts-csharp-repro
go mod init gtsrepro
go get github.com/odvcencio/gotreesitter@v0.20.8
curl -sO https://raw.githubusercontent.com/JamesNK/Newtonsoft.Json/4f73e74372445108d2c1bda37b36e6f5e43402e0/Src/Newtonsoft.Json/JsonTextReader.cs
# parse JsonTextReader.cs and walk for method_declaration nodes → 0 found
```

## Likely area

Per #116, this is the C# recovery normalizers: when the GLR parse of a class body fails, the class header/members are shredded into ERROR fragments and the lenient reconstruction recovers the type shell but not the methods. Fully recovering these would need source-based reconstruction of the method declarations (which #64/#98/#106 kept bounded for cost reasons).

## Environment

- gotreesitter **v0.20.8**
- Go **1.26.2**, darwin/arm64
- Grammar: `grammars.CSharpLanguage()` / `DetectLanguage("*.cs")`

Tracked downstream at richardwooding/file-search-on#440.

richardwooding:
Fixes #136. Follow-up to #115/#116.

## Problem

Large real-world C# files whose class body is shredded by a cumulative GLR failure (e.g. `Newtonsoft.Json`'s `JsonTextReader.cs` / `JsonReader.cs`) recovered a `namespace_declaration` shell (per #115) but its body held **only comments** — every class, and therefore every `method_declaration`, was lost. Downstream symbol/complexity extraction got 0 methods.

The source-based type/method reconstruction was hard-gated to files ≤ `csharpMaxTopLevelChunkRecoverySourceBytes` (4096 bytes), so on a large collapsed class nothing rebuilt the members; and the child-based reconstruction needs a clean `modifier* type identifier parameter_list` sibling sequence that the cumulative GLR failure destroys.

## Approach

When the child-based namespace pass surfaces **no method**, fall back to a **per-member bounded** source reconstruction (`parser_result_csharp_method_recovery.go`):

- The type shell (modifiers, name, type parameters, base list) is reparsed from the header with a synthetic empty body, so the parser handles all header shapes; the empty body is then swapped for the recovered members.
- Each body member is recovered on its own: a method via the existing signature-shell + lenient-block path (`csharpRecoverClassMethodDeclarationFromRange`, which blanks the body for the signature reparse and tolerates an unparseable body), other members (fields, properties, events, nested types) via a single small wrapped reparse. Members that still won't parse are skipped rather than failing the whole type.

## Boundedness

The bounds are **per declaration, not per file** — the key difference from the whole-file 4096 gate (kept unchanged):

- each reparse is a single small member snippet, capped by `csharpMaxMemberRecoverySourceBytes` and `csharpMaxTypeMemberRecoveries`;
- every reparse goes through `parseForRecovery`, which honors the parser timeout/cancellation, so a tight budget degrades gracefully to partial/zero recovery rather than running unbounded.

This preserves the anti-OOM guarantees from #64/#98/#106.

## Results

| File | Methods before | Methods after |
|---|---|---|
| `JsonTextReader.cs` | 0 | 68 |
| `JsonReader.cs` | 0 | 41 |

Members guarded by interspersed `#if`/`#endif` preprocessor directives remain out of scope (a handful of the ~72).

## Testing

- New `TestCSharpLargeShreddedNamespaceRecoversMethods` over a trimmed, MIT-attributed `JsonTextReader` excerpt fixture (>4096 bytes) — asserts `class_declaration >= 1`, `method_declaration >= 8` (was 0), byte-faithful root, and not stopped early.
- All existing `#115`/`#116` C# recovery tests and the C# boundedness suite (`TestCSharp*StaysBounded`) still pass, as do `TestTop50ParseSmokeNoErrors` / `TestSupportedLanguagesParseSmoke`.

Docker parity harness was unavailable in this environment; validation used the real Newtonsoft files plus the scoped Go tests above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Fixed
+
+- Large C# files whose class body is shredded by a cumulative GLR failure (e.g.
+  Newtonsoft.Json's `JsonTextReader.cs` / `JsonReader.cs`) now recover their
+  `method_declaration` nodes instead of yielding only a comment-filled namespace
+  shell. Follow-up to #115/#116: the source-based type/method reconstruction was
+  gated off above 4096 bytes, so nothing rebuilt the members of a large collapsed
+  class. Namespace recovery now falls back, when the child-based pass surfaces no
+  method, to a **per-member bounded** source reconstruction — the type shell's
+  header is reparsed for its modifiers/name/base list, and each member is
+  recovered on its own (a method via signature-shell + lenient block, other
+  members by a single small wrapped reparse), skipping any that still won't parse.
+  Each reparse is a single small snippet capped by size and count and honors the
+  parser timeout, so the anti-OOM guarantees from #64/#98/#106 are preserved and
+  the whole-file 4096-byte gate is unchanged. `JsonTextReader.cs` now recovers 68
+  methods (was 0) and `JsonReader.cs` 41 (was 0) (#136).
+
 ## [0.20.8] - 2026-07-01
 
 Adds consumer-controllable forest parsing.
--- a/parser_csharp_namespace_recovery_test.go
+++ b/parser_csharp_namespace_recovery_test.go
@@ -1,6 +1,7 @@
 package gotreesitter_test
 
 import (
+	"os"
 	"testing"
 
 	gotreesitter "github.com/odvcencio/gotreesitter"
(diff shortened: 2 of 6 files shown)


richardwooding:
Thanks — both applied in afb5643:

- **Parent links (high):** `csharpReplaceDeclarationList` now re-wires the swapped `declaration_list`'s parent pointer and child index and calls `populateParentNode` (matching `csharpReplaceMethodBlock`). Verified end-to-end: all 68 recovered methods in JsonTextReader.cs now navigate up to their `class_declaration`.
- **Fallback (`>=`→`>`):** the source-based pass is now kept only when it recovers strictly more methods, so when neither pass finds a method the child-based members are retained.

richardwooding:
Thanks — all four applied in a0429a4: the three source-recovery bounds checks now compare `uint32` offsets directly against `uint32(len(source))`, and `csharpReplaceDeclarationList` guards against a nil `decl`/`declList` before dereferencing. That covers every `int()`-cast bounds check in the new file.

odvcencio (approved):
Reviewed via probe battery — see comment above.

PR #147, opened 2026-07-08
Wave 5: support large GOTOs and clean COBOL recovery

odvcencio:
## Summary

Wave 5 makes Cobol clean against the C oracle without timing out by addressing the large-state table surface and the remaining Cobol compatibility gaps found in the real corpus.

## Changes

- Add `Language.LargeStateGotos` for nonterminal GOTO targets that exceed the uint16 table-cell limit.
- Build grammargen parse-table intermediates as uint32 and spill only oversized nonterminal GOTOs, while preserving uint16 terminal action safety.
- Teach runtime `lookupGoto` and the C recovery gate/cache to validate and consume large-state GOTO entries.
- Tighten Cobol result normalization for fixed-format leading area starts, isolated `program-id.` trailing junk, division sibling ends, and recovered trailing paragraph headers.
- Add focused runtime, recovery gate, grammargen boundary, and Cobol corpus snippet coverage.

## Testing

- `go test . -run '^(TestNormalizeResultCompatibilityDispatchesUppercaseCobol|TestNormalizeCobolFixedFormatProgramIDStopsBeforeTrailingJunk|TestNormalizeCobolRecoveredParagraphHeader|TestLookupGotoUsesLargeStateGotos|TestCRecoveryGateAcceptsLargeStateGotos|TestCRecoveryGateValidatesParseTableActionAndGotoBounds)$' -count=1`
- `go test ./grammargen -run '^(TestCheckUint16IndexBoundary|TestCheckUint16IndexErrorIsActionable|TestMaxUint16IndexValue)$' -count=1`
- `bash cgo_harness/docker/run_parity_in_docker.sh --memory 8g --cpus 1 --pids 512 --label cobol-evaluate-focused --no-build -- "cd /workspace && GOMAXPROCS=1 GOFLAGS='-p=1' go test ./grammargen -run '^TestCobolCorpusSnippetDeepParity/evaluate_when_other_corpus_block_exact$' -count=1 -v -timeout 15m"`
- `bash cgo_harness/docker/run_parity_in_docker.sh --memory 8g --cpus 1 --pids 512 --label cobol-snippets --no-build -- "cd /workspace && GOMAXPROCS=1 GOFLAGS='-p=1' go test ./grammargen -run '^TestCobolCorpusSnippetDeepParity$' -count=1 -v -timeout 15m"`
[...]

PR #151, opened 2026-07-08
Fix: preserve error flags during COBOL header normalization

odvcencio:
## Summary

Fixes the Cobol release-hold HasError edge case from the #147 review/audit window.

`normalizeCobolRecoveredParagraphHeader` converted a trailing recovered paragraph-label `ERROR` into a clean `paragraph_header`, then unconditionally cleared `HasError` on both `procedure_division` and the root. That was correct for the clean recovered-header case, but wrong for mixed-error trees: a separate retained error elsewhere in the subtree could be silently hidden.

This PR keeps the paragraph-header repair but recomputes parent error flags from retained children after the rewrite.

## Changes

- Replace unconditional `setHasError(false)` calls in `normalizeCobolRecoveredParagraphHeader` with child-based `HasError` refreshes.
- Add `cobolRefreshHasErrorFromChildren` for the local Cobol normalization path.
- Add `TestNormalizeCobolRecoveredParagraphHeaderPreservesOtherErrors`, which keeps a non-header error inside `procedure_division` and verifies root/procedure `HasError` remains true after the trailing paragraph header is recovered cleanly.

## Verification

- `git diff --check`
- Docker focused Cobol normalization gate:
  `bash cgo_harness/docker/run_parity_in_docker.sh --label cobol-haserror-focused --timeout 20m --no-build -- "cd /workspace && go test . -run '^(TestNormalizeCobolRecoveredParagraphHeader|TestNormalizeCobolRecoveredParagraphHeaderPreservesOtherErrors|TestNormalizeCobolFixedFormatProgramIDStopsBeforeTrailingJunk|TestNormalizeResultCompatibilityDispatchesUppercaseCobol)$' -count=1 -v -timeout 20m"`
- Artifacts: `harness_out/docker/20260708T042527Z-cobol-haserror-focused`

## Release Note

This is intended to land before `v0.22.0` so that release does not knowingly ship the latent Cobol mixed-error `HasError` bug.

### DIFF
--- a/parser_result_c_go_cobol_test.go
+++ b/parser_result_c_go_cobol_test.go
@@ -467,3 +467,95 @@ func TestNormalizeCobolRecoveredParagraphHeader(t *testing.T) {
 		t.Fatalf("paragraph_header.HasError = true, want false")
 	}
 }
+
+func TestNormalizeCobolRecoveredParagraphHeaderPreservesOtherErrors(t *testing.T) {
+	lang := &Language{
+		Name:                  "COBOL",
+		GeneratedByGrammargen: true,
+		SymbolNames:           []string{"EOF", ".", "start", "program_definition", "identification_division", "procedure_division", "END_EVALUATE", "period", "paragraph_header"},
+		SymbolMetadata: []SymbolMetadata{
+			{Name: "EOF", Visible: false, Named: false},
+			{Name: ".", Visible: true, Named: false},
+			{Name: "start", Visible: true, Named: true},
+			{Name: "program_definition", Visible: true, Named: true},
+			{Name: "identification_division", Visible: true, Named: true},
+			{Name: "procedure_division", Visible: true, Named: true},
+			{Name: "END_EVALUATE", Visible: true, Named: true},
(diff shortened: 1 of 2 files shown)


odvcencio:
Review/update for the pre-release Cobol HasError fix:

- Local review: no blocking findings. The bug was the unconditional `setHasError(false)` on `procedure_division` and root after only the trailing paragraph-header `ERROR` had been repaired. Recomputing from retained children preserves the existing clean case while keeping unrelated retained errors visible.
- The new regression covers the mixed-error shape: retained non-header error inside `procedure_division`, trailing paragraph-label error converted to clean `paragraph_header`, and root/procedure `HasError` expected to remain true.
- Focused Docker gate passed: `harness_out/docker/20260708T042527Z-cobol-haserror-focused`.
- CI status at this comment: freshness, parity-cgo, and perf-regression are green; build is still running.

Buckley/GLM 5.2 review note: I attempted two review-only Buckley passes. One timed out with no output; the smaller retry emitted raw tool-call text instead of a usable verdict. I am not treating that as review evidence.

PR #152, opened 2026-07-08
refactor(ci): split build gate into parallel jobs

odvcencio:
## Summary

The CI build gate previously ran all build, race, and parity steps serially in a single 60-minute job. This created unnecessary bottleneck time since compile, root race tests, non-root race tests, grammargen visibility, and parity report could all run in parallel. This PR splits the monolithic `build` job into parallel jobs while keeping a lightweight aggregator `build` job as the required check, preserving the same blocking semantics for non-draft PRs (race tests + parity) and draft PRs (focused tests).

## Changes

- Split the monolithic `build` job into 6 independent jobs: `compile`, `race_root`, `race_packages`, `grammargen_visibility`, `draft_focused_tests`, and `parity_report`
- Separated root-package race tests (`race_root`) from non-root package race tests (`race_packages`), both running in parallel with 35m timeouts instead of serially under a single 60m budget
- Moved `grammargen_visibility` and `parity_report` into their own jobs so the serial grammargen/parity tail no longer blocks the build gate timeline
- Added a new `build` aggregator job with `if: always()` that enforces required checks: compile + parity always required; race_root + race_packages required for non-draft PRs; draft_focused_tests required for draft PRs
- Reduced compile timeout from 60m to 15m since it no longer includes the full race suite
- Preserved the non-blocking `continue-on-error: true` semantics for grammargen visibility
- Updated the draft-correctness comment block to reference the new parallel job names instead of the old monolithic step structure

## Testing

- Open a non-draft PR and confirm that `compile`, `race_root`, `race_packages`, `parity_report`, and `grammargen_visibility` all run in parallel rather than serially
- Verify the `build` aggregator job fails if any of `compile`, `race_root`, `race_packages`, or `parity_report` fails on a non-draft PR
- Verify the `build` aggregator job fails if `compile`, `draft_focused_tests`, or `parity_report` fails on a draft PR, while `race_root`/`race_packages` are skipped
- Confirm `grammargen_visibility` failures do NOT fail the `build` gate (continue-on-error)
- Check that the total wall-clock time for the pipeline is roughly the max of the individual jobs rather than their sum
- Validate that the parity report artifact (`parity-log`) is still uploaded correctly

## Review Focus

Focus on the `build` aggregator job's logic: verify that the IS_DRAFT conditional correctly maps to the right required jobs, and that no previously-blocking check was accidentally made non-blocking (or vice versa). Also confirm that the parallel `race_root` and `race_packages` jobs cover exactly the same test surface as the old single `go test` command (root package + non-root non-grammargen packages).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,9 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  compile:
     runs-on: ubuntu-latest
-    # 60m: the full -race suite (non-draft PRs) needs ~30-40m on 2-core
-    # runners with -p 1; see the Test step's comment.
-    timeout-minutes: 60
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


odvcencio:
Review note after the CI split follow-up:

- Buckley review was attempted twice on this branch. Both runs completed, but this Buckley install wrote only tool-call summaries to the requested output files instead of a findings/approval report, so I am not treating those artifacts as a valid approval.
- Tried the requested max model path as well; Buckley/provider rejected `z-ai/glm-5.2-max` as an invalid model ID, so the usable configured review model remains `z-ai/glm-5.2` here.
- Direct review of the workflow diff found no blocker: the required aggregate `build` gate now checks `compile`, `parity_report`, and the appropriate race/draft jobs, while `grammargen_visibility` stays informational.
- CI is green on the updated head, including `compile`, `race_root`, `race_packages`, `grammargen_visibility`, `parity_report`, `freshness`, `parity-cgo`, `perf-regression`, and aggregate `build`.

Ready to merge from my side.

PR #157, opened 2026-07-08
fix(cpp-recovery): normalize fused malformed class+method

odvcencio:
## Summary

This PR closes the focused Wave 5 C++ malformed-class recovery parity case without enabling cpp C-recovery globally. The Go parser already preserves the source content, but the C oracle folds a malformed class body plus the following `void A::b() {}` into one recovered `function_definition`. This adds a narrow cpp result-compatibility normalizer for that exact shape.

## Changes

- Add `normalizeCppMalformedClassFunctionDefinition` in `parser_result_cpp.go` and call it from the existing C/C++ compatibility pipeline.
- Reconstruct the C-compatible recovered shape: `class_specifier`, rewritten `function_declarator`, `void` retagged as `namespace_identifier`, and the original namespace `A` wrapped in an extra `ERROR(identifier)`.
- Keep the rewrite guarded by cpp language, translation-unit sibling shape, concrete node types, child counts, and `void` source text.
- Update the pure-Go C++ recovery regression to assert the C-compatible 2-child root and recovered qualified identifier shape.
- Add `REPRO_GO_BACKEND=registry` to the C tree-dump diagnostic so the registry parser path can reproduce token-source trees.

## Testing

- `git diff --check HEAD~1..HEAD`
- `staticcheck ./...`
- `go test . -run '^(TestCppMalformedClassFunctionDefinitionRecovery|TestNormalizeCppCompatibilityRestoresCollapsedKeywordChildren)$' -count=1`
- `go test ./grammars -run '^TestCppUserTypeTemplateArgumentInParameterListParsesWithoutError$' -count=1`
- `bash cgo_harness/docker/run_parity_in_docker.sh --label wave5-cpp-focused-cgo-clean-refactor --timeout 20m --no-build --memory 8g --cpus 1 --pids 512 -- "cd /workspace/cgo_harness && go test -tags 'cgo treesitter_c_parity' . -run '^TestCpp(TemplateTypeParameterParity|CollapsedKeywordCompatibilityParity|MalformedClassFunctionDefinitionRecoveryParity)$' -count=1 -timeout 10m -v"`

Docker artifact: `harness_out/docker/20260708T113321Z-wave5-cpp-focused-cgo-clean-refactor`, exit_code=0, oom_killed=false.

## Notes

This intentionally does not claim full cpp corpus parity. The broader single-grammar cpp sweep still reports known corpus mismatches outside this focused recovery fold.

## Review Focus

Please check the rewrite guards, arena-clone helper usage, and the retag/error-wrapper shape against the C-oracle tree for this malformed-class case.

### DIFF
--- a/cgo_harness/zz_c_tree_dump_test.go
+++ b/cgo_harness/zz_c_tree_dump_test.go
@@ -92,8 +92,17 @@ func TestCTreeDumpDiag(t *testing.T) {
 	if goLang == nil {
 		t.Fatalf("%s: not in grammars.AllLanguages", name)
 	}
-	gp := gts.NewParser(goLang)
-	gt, _ := gp.Parse(src)
+	var gt *gts.Tree
+	if os.Getenv("REPRO_GO_BACKEND") == "registry" {
+		var parseErr error
+		gt, _, parseErr = parseWithGo(parityCase{name: name, source: string(src)}, src, nil)
+		if parseErr != nil {
+			t.Fatalf("go registry parse failed: %v", parseErr)
+		}
+	} else {
+		gp := gts.NewParser(goLang)
+		gt, _ = gp.Parse(src)
+	}
 	if gt == nil || gt.RootNode() == nil {
 		t.Fatalf("go parse failed")
 	}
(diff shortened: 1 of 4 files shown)


odvcencio:
Review pass: I checked the focused C++ recovery normalizer manually and with Canopy. The rewrite guards are limited to cpp translation-unit sibling shape, concrete node types/child counts, and the `void` source text; arena cloning/population is localized through the new helpers; and the C tree-dump backend switch is opt-in via `REPRO_GO_BACKEND`. Local focused tests, `staticcheck ./...`, and focused Docker C-oracle parity passed. Self-approval is blocked by GitHub, so this is recorded as a review comment rather than an approval.

PR #174, opened 2026-07-08
ci: trigger exhaustive CGo parity on parser-result/recovery PRs

odvcencio:
## Summary

Previously, the exhaustive CGo parity job only ran on manual `workflow_dispatch` triggers, meaning PRs touching critical parser-result or C-recovery files could merge without full parity validation. This PR introduces an `exhaustive_parity_scope` job that conditionally triggers exhaustive parity on PRs when changed files match parser-result or C-recovery patterns, and wires it into the build gate so those PRs are blocked until parity passes.

## Changes

- Add `exhaustive_parity_scope` job that detects whether exhaustive CGo parity should run — always on `workflow_dispatch`, and on PRs only when changed files match `parser_result_*.go`, `parser_recover_c*.go`, or `parser_result_test/*.go`
- Rework `parity-cgo-exhaustive` to depend on the scope job (`needs: exhaustive_parity_scope`) and gate on `run_exhaustive == 'true'` instead of hardcoding `workflow_dispatch` only
- Update the `build` gate to always wait on and require `exhaustive_parity_scope`, and conditionally require `parity-cgo-exhaustive` only when the scope job signals exhaustive should run
- Add a GITHUB_STEP_SUMMARY block showing which files triggered (or why it skipped) exhaustive parity for PR transparency

## Testing

- Open a PR that modifies a file like `parser_result_tree.go` — verify `exhaustive_parity_scope` outputs `run_exhaustive=true` and `parity-cgo-exhaustive` runs and is required by the `build` gate.
- Open a PR that only touches unrelated files (e.g., a README or a non-parser `.go` file) — verify `exhaustive_parity_scope` outputs `run_exhaustive=false`, `parity-cgo-exhaustive` is skipped, and the `build` gate still passes.
- Trigger the workflow via `workflow_dispatch` — verify `exhaustive_parity_scope` sets `run_exhaustive=true` and `parity-cgo-exhaustive` runs as before.
- Inspect the PR's job summary (GITHUB_STEP_SUMMARY) to confirm the trigger rationale is displayed correctly.

## Review Focus

Pay attention to the grep regex pattern (`^(parser_result_.*\.go|parser_recover_c.*\.go|parser_result_test/.*\.go)$`) — make sure it matches the intended file naming conventions and doesn't accidentally match unrelated files. Also verify that the `build` gate correctly handles the case where `parity-cgo-exhaustive` is skipped (result will be 'skipped', not 'success').

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,6 +196,49 @@ jobs:
           GOWORK=off go run ./cmd/perf_scan_budget \
             -budget perf_scan/perf_ratio_budgets.json
 
+  exhaustive_parity_scope:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      run_exhaustive: ${{ steps.scope.outputs.run_exhaustive }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - id: scope
+        name: Detect exhaustive parity scope
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          set -euo pipefail
+
+          run_exhaustive=false
+          if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
+            run_exhaustive=true
+          elif [ "$EVENT_NAME" = "pull_request" ]; then
+            git diff --name-only "${PR_BASE_SHA}...HEAD" > /tmp/changed-files.txt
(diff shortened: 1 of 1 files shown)


odvcencio (commented):
Buckley diff review (z-ai/glm-5.2): no blocking findings.\n\nReviewed the workflow gate changes specifically for trigger scope, skipped-job handling, and build aggregation. The scope job preserves workflow_dispatch behavior, skips exhaustive parity on unrelated PRs, and the build gate only requires parity-cgo-exhaustive when the scope output is true.

PR #185, opened 2026-07-09
add parent-side RSS watchdog to perf scan harness

odvcencio:
## Summary

This PR introduces a configurable parent-side RSS watchdog (GTS_PERF_SCAN_CHILD_RSS_LIMIT_MB) for the perf scan harness. Previously, when a language subprocess exceeded memory limits, the container OOM killer would terminate the parent process, causing loss of intermediate scoreboard data. The new watchdog monitors child process RSS via /proc/<pid>/status and kills it gracefully before a container-level OOM occurs, preserving partial results and preventing parent death.

## Changes

- Added GTS_PERF_SCAN_CHILD_RSS_LIMIT_MB env var and config field
- Implemented perfScanRunChild to spawn subprocesses with RSS monitoring via polling /proc/<pid>/status
- Added perfScanProcessRSSBytes and perfScanParseStatusRSSBytes helper functions
- Updated perfScanRunLanguageSubprocess to pass RSS kill details to scoreboard fragments on early exit
- Updated Markdown board rendering to display the active child RSS limit
- Added unit tests for VmRSS parsing and config loading
- Updated README and sweep status/budget files to reflect the new containment strategy for Groovy/F# memory blowups

## Testing

- Run `go test -run TestPerfScanHelpersUnit ./cgo_harness` to verify RSS parsing and config loading
- Verify that setting `GTS_PERF_SCAN_CHILD_RSS_LIMIT_MB=512` correctly kills a memory-heavy child and preserves the scoreboard
- Run a subset of the perf sweep to confirm the dashboard markdown includes the `- child RSS limit: <N> MiB` line
- Confirm existing perf scan behavior remains unchanged when the flag is unset (default 0)

### DIFF
--- a/cgo_harness/perf_scan/README.md
+++ b/cgo_harness/perf_scan/README.md
@@ -115,6 +115,7 @@ grammar repos, not corpus repos.
 | `GTS_PERF_SCAN_CONTENDED` | auto (loadavg) | mark run as contended (smoke-only numbers) |
 | `GTS_PERF_SCAN_INPROCESS` | 0 | debug: run languages in-process (no crash isolation) |
 | `GTS_PERF_SCAN_EDIT_CANDIDATES` | 16 | edit-site candidates tried per file |
+| `GTS_PERF_SCAN_CHILD_RSS_LIMIT_MB` | 0 | optional parent-side RSS watchdog for the per-language child process; when set, kills the child before a container cgroup OOM can kill the sweep parent |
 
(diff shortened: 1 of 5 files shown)


odvcencio:
Review follow-up: addressed the PID-reuse concern in commit 92e2c3a by checking the child Wait result before any further /proc sampling. I left the 10ms poll interval intentionally tight because GTS_PERF_SCAN_CHILD_RSS_LIMIT_MB is a containment/RCA knob for pathological memory probes, not the authoritative timing path; the default remains disabled.

PR #190, opened 2026-07-09
add(grammargen): derive non-terminal alias map

odvcencio:
## Summary

grammargen now derives `Language.NonTerminalAliasMap` for generated grammars, mirroring the `ts_non_terminal_alias_map` table that ts2go already extracts from tree-sitter's C parser output. Previously grammargen-generated blobs omitted this field entirely, which is one of the remaining structural gaps between grammargen and ts2go output.

For every nonterminal grammar symbol, the derivation collects the set of distinct display symbols it can take across all productions: the symbol itself when referenced unaliased anywhere, plus each resolved alias target when referenced aliased. A symbol whose only possible display is a single static value (never aliased, or always aliased the exact same way) gets no map entry, matching tree-sitter's `public_symbol_map` fallback behavior for unambiguous symbols.

`buildAliasSequences` now returns the alias-symbol resolution map it already builds internally, so the new derivation reuses the exact same symbol IDs rather than recomputing a second, potentially inconsistent resolution.

## Verification

- Scanned all 206 shipped `grammar_blobs/*.bin`: only `lua.bin` carries a non-empty `NonTerminalAliasMap` (its `_doublequote_string_content` / `_singlequote_string_content` aux wrappers, each referenced both unaliased via their own repeat1 self-recursion and aliased to `string_content`).
- Added a parity test that imports the pinned tree-sitter-lua `grammar.json`, generates the language through grammargen's own LR pipeline, and asserts the derived map matches the shipped ts2go blob's map exactly (same symbol -> alias-target-name-set entries) for both affected rows.
- Added an inventory test that fails if any other shipped blob ever gains a non-empty map without matching parity coverage being added.
- Added offline synthetic regression tests pinning the derivation rules directly: self+alias inclusion, singleton-alias omission, terminal-alias exclusion, and determinism across repeated generation of the same grammar.
- Verified the gob-encoded blob round-trip preserves the derived map exactly.
- Confirmed a pre-existing, unrelated metadata classification gap (grammargen marks certain hidden repeat-bodied rules `Named=true` where tree-sitter's compiler would use `Named=false`) does not affect this derivation or its consumer, and is out of scope here; it's logged non-fatally by the parity test for visibility.

## Test plan

- [x] `GOWORK=off go build ./...`
- [x] `GOWORK=off go test ./grammargen/ -run NonTerminalAliasMap -v` (6/6 pass)
- [x] Full `grammargen` test suite run, with targeted before/after comparisons against unmodified code confirming the pre-existing slow/divergent tests (Swift grammar generation timing, Markdown CommonMark parity gaps) are unrelated and unchanged by this diff

### DIFF
--- a/grammargen/assemble.go
+++ b/grammargen/assemble.go
@@ -222,8 +222,12 @@ func assemble(
 		}
 	}
 
-	// Alias sequences.
-	buildAliasSequences(lang, ng)
+	// Alias sequences, and the derived non-terminal alias map (mirrors
+	// tree-sitter C's ts_non_terminal_alias_map; must run after
+	// buildAliasSequences so alias target symbols already exist in the table
+	// and resolve to the same IDs used in AliasSequences).
+	aliasSymMap := buildAliasSequences(lang, ng)
+	buildNonTerminalAliasMap(lang, ng, aliasSymMap)
 
 	// Supertype map.
 	if err := buildSupertypeMap(lang, ng); err != nil {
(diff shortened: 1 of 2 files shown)


odvcencio:
Buckley/GLM review evidence for merge:

- Buckley review model: `z-ai/glm-5.2` (`buckley config show` reports Planning/Execution/Review all on GLM 5.2).
- Review artifact: `/tmp/buckley_review_190_glm.md`. Verdict: Approved: Yes; Blockers: No.
- The one potential critical note in the review was only to verify the `sort` import; confirmed present in `grammargen/assemble.go`, and the focused Docker test compiled/passed.
- Focused Docker validation passed:
  - `bash cgo_harness/docker/run_parity_in_docker.sh -- "cd /workspace && go test ./grammargen -run '^(TestShippedBlobsNonTerminalAliasMapInventory|TestLuaNonTerminalAliasMapParity|TestNonTerminalAliasMapSyntheticSelfAndAlias|TestNonTerminalAliasMapSyntheticSingletonAliasOmitted|TestNonTerminalAliasMapSyntheticTerminalAliasExcluded|TestNonTerminalAliasMapDeterministic)$' -count=1"`\n- `git diff --check origin/main...HEAD` is clean.\n- PR CI is green, including `build`, `compile`, `parity-cgo`, `perf-regression`, and `grammargen_visibility`.

PR #194, opened 2026-07-09
Add file exclusion support to performance scan tooling

odvcencio:
## Summary

Adds `GTS_PERF_SCAN_EXCLUDE_PATHS` so perf-scan runs can omit named language-relative or `language/path` witnesses from corpus selection while still recording the exclusion list in scoreboard config. This gives Wave 3 a reproducible way to seed scoped ratchet rows around known C-oracle cliff files without erasing those witnesses from the gap ledger.

## Changes

- Adds `exclude_paths` to perf-scan scoreboard config and perf-ratio budget measurement basis.
- Filters selected corpus files by exact path, `language/path`, simple `path.Match` globs, and trailing-slash directory prefixes.
- Teaches `perf_scan_budget` strict config comparison to reject mismatched exclusion lists.
- Documents `GTS_PERF_SCAN_EXCLUDE_PATHS` in `cgo_harness/perf_scan/README.md`.

## Testing

- `go test ./cmd/perf_scan_budget -count=1`
- `bash cgo_harness/docker/run_parity_in_docker.sh --label wave3-perfscan-exclude-helper-r2 -- "cd /workspace/cgo_harness && GOWORK=off GTS_PARITY_ALLOW_HOST=1 GTS_PERF_SCAN=1 go test -tags 'treesitter_c_parity treesitter_c_perfscan' -run '^TestPerfScanHelpersUnit$' -count=1 ."`\n- Docker smoke: `GTS_PERF_SCAN_LANGS=json GTS_PERF_SCAN_MAX_FILES=1 GTS_PERF_SCAN_ORDER=path GTS_PERF_SCAN_EXCLUDE_PATHS=json/medium__grammar.json ... TestPerfScanSweep`\n- Smoke scoreboard confirmed `exclude_paths=["json/medium__grammar.json"]` and selected `medium__node-types.json` instead of the excluded file.\n- `git diff --check`

### DIFF
--- a/cgo_harness/cmd/perf_scan_budget/main.go
+++ b/cgo_harness/cmd/perf_scan_budget/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path"
 	"sort"
 	"strings"
 )
@@ -188,6 +189,13 @@ func validateBudget(b *budgetFile) []evalFinding {
 			out = append(out, evalFinding{Axis: axis, Metric: "measurement_basis.axes", Got: strings.Join(b.MeasurementBasis.Axes, ","), Want: "include " + axis})
 		}
 	}
+	for _, pattern := range normalizedPathList(b.MeasurementBasis.ExcludePaths) {
+		if strings.ContainsAny(pattern, "*?[") {
+			if _, err := path.Match(pattern, "x"); err != nil {
+				out = append(out, evalFinding{Metric: "measurement_basis.exclude_paths", Got: pattern, Want: "valid path.Match pattern"})
+			}
+		}
+	}
 	for _, lang := range sortedBudgetLanguages(b) {
 		entry := b.Languages[lang]
 		if strings.TrimSpace(entry.Status) == "" {
@@ -295,7 +303,7 @@ func compareConfig(b budgetMeasurementBasis, s scoreboardConfig) []evalFinding {
 	if len(b.ExcludePaths) > 0 || len(s.ExcludePaths) > 0 {
 		got := normalizedPathList(s.ExcludePaths)
 		want := normalizedPathList(b.ExcludePaths)
-		if strings.Join(got, ",") != strings.Join(want, ",") {
+		if !stringSlicesEqual(got, want) {
 			out = append(out, evalFinding{Metric: "config.exclude_paths", Got: strings.Join(got, ","), Want: strings.Join(want, ",")})
 		}
 	}
@@ -431,7 +439,9 @@ func normalizedPathList(items []string) []string {
 	seen := map[string]bool{}
 	var out []string
 	for _, item := range items {
-		item = strings.Trim(strings.ReplaceAll(strings.TrimSpace(item), "\\", "/"), "/")
+		item = strings.ReplaceAll(strings.TrimSpace(item), "\\", "/")
+		item = strings.TrimPrefix(item, "./")
+		item = strings.TrimPrefix(item, "/")
 		if item == "" || seen[item] {
 			continue
 		}
@@ -440,6 +450,18 @@ func normalizedPathList(items []string) []string {
 	}
 	sort.Strings(out)
 	return out
+}
+
+func stringSlicesEqual(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
 }
 
 func contains(items []string, needle string) bool {
--- a/cgo_harness/cmd/perf_scan_budget/main.go
+++ b/cgo_harness/cmd/perf_scan_budget/main.go
@@ -32,6 +32,7 @@ type budgetMeasurementBasis struct {
 	MaxFiles     int      `json:"max_files,omitempty"`
 	Order        string   `json:"order,omitempty"`
 	Axes         []string `json:"axes"`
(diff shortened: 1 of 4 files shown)


odvcencio:
Buckley/GLM review evidence at current head `b1a28b67`:

# PR #194 Review — `exclude_paths` for perf_scan

## Verdict: **APPROVE**

The PR adds a path-exclusion mechanism to the perf_scan pipeline, with budget validation, config-strictness comparison, path normalization, glob/trailing-slash matching, and tests. The implementation is sound, backward-compatible, and well-tested for the core scenarios. Findings below are all low severity and do not block merge.

---

## Findings (ordered by severity)

### LOW-1: Duplicated normalization logic across packages
**Files:** `cgo_harness/cmd/perf_scan_budget/main.go` (`normalizedPathList`) and `cgo_harness/zz_perf_scan_test.go` (`perfScanPathList`)

Both functions implement identical normalization (backslash→slash, strip `./` and leading `/`, dedup, sort) but in different packages. Any future fix to one must be manually replicated in the other. Consider extracting to a shared utility or adding a cross-check test that feeds the same inputs to both and asserts identical output. Not a blocker since they're currently in sync.

### LOW-2: `strings.TrimPrefix(item, "./")` only strips one leading `./`
**Files:** `cgo_harness/cmd/perf_scan_budget/main.go:~460` (`normalizedPathList`), `cgo_harness/zz_perf_scan_test.go:~~350` (`perfScanPathList`)

A pattern like `././foo` would normalize to `./foo`, not `foo`. This could cause a silent matching failure if someone nests relative prefixes. Extremely unlikely in practice (env vars and JSON typically don't contain `././`), but `strings.Trim` with `"./"` as a cutset would be wrong too (it'd strip individual `.` and `/` chars). If this matters, a loop would be needed. Low risk.

### LOW-3: Missing test for glob depth limitation
**File:** `cgo_harness/zz_perf_scan_test.go` — `TestPerfScanHelpersUnit` table

The test cases cover exact match, backslash, language-relative, glob, and directory-prefix. But there's no case verifying that a glob like `*.fs` does **NOT** match `sub/file.fs` (i.e., `*` doesn't cross `/`). The "miss" test case only uses a completely unrelated pattern. Adding `{"glob no cross slash", "fsharp", "sub/file.fs", []string{"*.fs"}, false}` would strengthen confidence in the `path.Match` semantics claim.

### LOW-4: No integration test for exclusion within `perfScanSelectFiles`
**File:** `cgo_harness/zz_perf_scan_test.go`

The unit tests verify `perfScanPathExcluded` and `perfScanPathPatternMatches` in isolation. There's no test that creates a temp corpus, sets `ExcludePaths` in `perfScanConfig`, calls `perfScanSelectFiles`, and asserts the excluded file is absent from the result. The wiring (`rel = filepath.ToSlash(rel)` followed by `perfScanPathExcluded(lang, rel, cfg.ExcludePaths)`) looks correct, but an integration test would guard against regressions in the call-site ordering.

### LOW-5: `strings.Trim(candidate, "/")` in `perfScanPathPatternMatches` is broader than needed
**File:** `cgo_harness/zz_perf_scan_test.go` — `perfScanPathPatternMatches`

`strings.Trim(candidate, "/")` strips ALL leading and trailing slashes from the candidate. This is fine for file paths from `filepath.WalkDir` (which don't have trailing slashes), but it's asymmetric with the pattern normalization which only strips a **leading** `/` via `TrimPrefix`. If a pattern were `dir/` (trailing slash, directory prefix) and the candidate were `dir/` (the directory itself, which shouldn't happen for files), the candidate would become `dir` and wouldn't match. Not a real-world issue since only files are walked, but the asymmetry is worth a comment.

### INFO-1: `path.Match` cannot escape literal `*`, `?`, `[`
**File:** `cgo_harness/zz_perf_scan_test.go` — `perfScanPathPatternMatches`

Corpus paths containing literal glob metacharacters cannot be matched exactly. This is a documented limitation of Go's `path.Match` and is mentioned in the README. Users would need to use a trailing-slash directory prefix for such paths. No action needed.

### INFO-2: Backward compatibility is correctly handled
**Files:** All modified structs

- `ExcludePaths` fields use `omitempty`, so old budget files and scoreboards deserialize correctly.
- `compareConfig` only checks `exclude_paths` when at least one side is non-empty, so old budgets/scoreboards without the field don't trigger false config mismatches.
- `GTS_PERF_SCAN_EXCLUDE_PATHS` defaults to empty (no exclusions), maintaining backward-compatible behavior.
- The scoreboard schema string remains `gts-perf-scan/v1`.

### INFO-3: Trailing slash semantics are consistent and well-tested
**Files:** `main.go` (`compareConfig`), `zz_perf_scan_test.go` (`perfScanPathPatternMatches`), `main_test.go` (`TestCompareScoreboardReportsExcludePathConfigMismatch`)

- In config comparison: `groovy/subprojects/` ≠ `groovy/subprojects` (trailing slash preserved by `normalizedPathList`).
- In path matching: `groovy/subprojects/` means recursive directory prefix; `groovy/subprojects` means exact match only.
- Both behaviors are explicitly tested. The README documents the distinction. Consistent and correct.

---

## Summary

The PR is clean. Path exclusion works correctly across exact, glob, and directory-prefix modes. Config strictness properly enforces that scoreboard and budget exclude_paths match after normalization. The trailing-slash distinction is intentional, tested, and documented. The only actionable items are minor test-coverage gaps (glob non-cross-slash case, integration test) and the duplicated normalization code, none of which warrant blocking.

PR #220, opened 2026-07-11
fix(parser): re-lex JavaScript zero-width externals per GLR version

odvcencio:
## Summary

Tree-sitter lexes each GLR version with that version's parser state. gotreesitter normally shares one lookahead across the frontier, which can hide JavaScript's zero-width `_automatic_semicolon` from a version that needs it after reductions.

This change adds two narrow correction paths:

- Before materializing a selected single-token recovery insertion as `MISSING`, ask the external scanner whether that parser version has a real zero-width shift at the same byte. If so, shift the scanner token and retry the original lookahead.
- When JavaScript GLR states diverge and every eligible live version rejects the shared real lookahead, re-lex one parser state. The replacement is accepted only when it is a zero-width external-scanner token with exactly one non-extra shift action. If any live version can consume the original token, the shared token is preserved.

The token source's lexer, scanner checkpoint, parser-state, and GLR-state snapshots are restored on probe/failure paths.

## Poppler result

On V8's 3,447,275-byte Emscripten Poppler corpus, the previous v0.23.1 parser stopped at byte 3,385,180 with `no_stacks_alive` on the second closing brace. This head now reaches EOF and returns:

- root: `program [0:3447275]`
- `HasError=false`
- stop reason: `accepted`
- no truncation or early EOF
- last token: EOF at byte 3,447,275

The direct-C run still reports a separate deep-shape mismatch at byte 44,043: a trailing JavaScript comment is owned by the inner `statement_block` in Go but is a sibling in C (`go=children=11`, `c=children=12`). Exact tree parity is therefore not claimed by this PR.

The hard 2 GiB container is also still OOM-killed. Memory reduction remains a separate follow-up; this PR is the scoped late-file correctness fix.

## Validation

- Focused Docker parser/re-lex gate passed:
  `TestNoLiveStackCanAcceptLookaheadRequiresEveryEligibleVersionToReject`,
  `TestTryRelexSingleParserStateAcceptsOnlyZeroWidthExternalShift`,
  `TestUpdateCurrentRelexParserStateTokenSourceExcludesShiftedStacks`, and
  `TestHighlightParserGoRealFile`.
- Final one-file Docker Go/C run completed without OOM in the established 8 GiB correctness envelope and preserved the isolated trailing-comment mismatch as the first divergence.
- `git diff --check` is clean.

## Scope and risk

- The all-versions-dead fallback is gated to `Language.Name == "javascript"`.
- Non-zero-width, non-external, ambiguous, extra, self-loop, and missing-action replacements fail closed.
- Mixed frontiers never replace a real token that another live version accepts.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -1754,13 +1754,16 @@ func (p *Parser) tryRelexCurrentStateDFA(tok Token, parserState StateID, ts Toke
 }
 
 // peekZeroWidthExternalShiftForState asks an external scanner whether one
-// parser version needs a zero-width token before the already-lexed shared
-// lookahead. The scanner and token-source state are restored before return;
-// callers apply the returned shift only to that version, then retry the real
-// lookahead. This mirrors tree-sitter C's per-version lexing without replacing
-// a real token that another live version may still consume.
+// JavaScript parser version needs a zero-width token before the already-lexed
+// shared lookahead. JavaScript's external scanner has no serialized state, so
+// the scanner and token-source state can be restored before return; callers
+// apply the returned shift only to that version, then retry the real lookahead.
+// This mirrors tree-sitter C's per-version lexing without replacing a real
+// token that another live version may still consume. Keep this JavaScript-only
+// until stateful scanners can transfer the probed post-scan checkpoint onto
+// the shifted version instead of restoring it.
 func (p *Parser) peekZeroWidthExternalShiftForState(tok Token, state StateID, ts TokenSource) (Token, ParseAction, bool) {
-	if p == nil || p.language == nil || tok.NoLookahead || tok.StartByte == tok.EndByte || p.language.ExternalScanner == nil {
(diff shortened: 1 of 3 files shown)


odvcencio:
Stack-level disposition of the remaining Terra memory finding

PR #220 intentionally isolates the late-file JavaScript correctness fix; PR #222 is its stacked memory-economy follow-up. The final #220 -> #221 -> #222 stack now has both required resource/correctness proofs:

- a prebuilt single-pass Go runtime accepts the exact 3,447,275-byte Poppler witness under a hard 2 GiB cgroup at max RSS 1,729,836 KiB, exact EOF, program root, HasError=false;
- the separate 8 GiB Go/C oracle run passes no-error, S-expression, and deep parity;
- the integrated JavaScript corpus gate passes 25/25 on all three axes with no OOM.

The earlier 2 GiB OOM is reproducible only when the parse-gap oracle process holds both giant Go and C trees for deep comparison. Correctness and runtime-memory gating are therefore kept separate, as required by AGENTS.md. This addresses the deployment/runtime impact cited by the review without mixing #222 memory architecture into the scoped #220 correctness commit.

PR #221, opened 2026-07-11
fix: preserve JavaScript trailing comment ownership

odvcencio:
## Summary

- match the pinned JavaScript scanner's same-line comment probe so a zero-width `_automatic_semicolon` is emitted before the comment extra
- promote JavaScript's precise external lex-state table for ordinary scanner validity while keeping default faithful C recovery explicitly opted out for its measured large-file cost
- add focused returned-tree and Docker C-oracle regressions for trailing-comment ownership
- preserve the pinned scanner's block-comment loop boundary across adjacent comments, preventing the following token from being consumed during ASI arbitration

## Why

On Poppler, Go shifted `} -> comment(extra) -> _automatic_semicolon` while C shifted `} -> _automatic_semicolon -> comment(extra)`. The reversed ordering made the comment interior to the inner `statement_block`. This patch fixes scanner arbitration rather than adding a result-tree rewrite or broad reducer heuristic.

A deeper source-to-source review also caught a Go porting trap: `break` inside the probe's `switch` exited only the switch, not the enclosing block-comment loop. A labeled break now matches the C scanner and is pinned by an adjacent-block-comment parity regression.

## Validation

- `go test ./grammars ./parser_result_test -count=1`
- `go test . -run 'CRecoveryGate|CRecoveryCostCompetition' -count=1`
- focused Docker C oracle: `TestParityJavaScriptAutomaticSemicolonBeforeTrailingComment`
- focused Docker C oracle: `TestParityJavaScriptAdjacentBlockCommentsDoNotConsumeFollowingToken`
- JavaScript one-grammar Docker gate: 25/25 no-error, 24/25 deep parity; the prior template ERROR is gone and the remaining generated-grammar comment-ownership residual is separate from the embedded runtime
- Poppler 3,447,275-byte parse-gap run in an 8 GiB container: `no_error=true`, `sexpr=true`, `deep=true`, accepted EOF, no OOM; Go 21.706s vs C 13.087s (1.66x)

## Stack

Stacked on #220 (`codex/poppler-versioned-external-relex`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -478,7 +478,7 @@ jobs:
             --goflags -p=1 \
             --test-parallel 1 \
             --timeout 20m \
-            --run '^TestParityFreshParse$|^TestParityIncrementalParse$|^TestParityHasNoErrors$|^TestParityIssue3Repros$|^TestParityGLRCanaryGo$|^TestParityGLRCanarySet$|^TestParityGLRCapPressureTopLanguages$|^TestParityGateCoverageRatchet$|^TestParityHighlight$'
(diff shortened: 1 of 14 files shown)


odvcencio:
Direct exact-head review completed at 40c8107fb96b3ddf60868be07e51623e04b87174. I inspected the full patch and Canopy blast radius, matched the ASI/comment probe against the pinned tree-sitter-javascript scanner, verified the precise external lex-state promotion remains independently opted out of default C-recovery election, and checked the CI selector ratchet plus both C-oracle regressions. Hosted checks and the integrated 25/25 JavaScript no-error/S-expression/deep-parity gate are green. No actionable findings.

PR #222, opened 2026-07-11
optimize(parser): bank Poppler memory economy gains

odvcencio:
## Summary

- shrink `Node` from 152 to 144 bytes, `rawShape` from 32 to 24 bytes, and `rawShapeChild` from 24 to 16 bytes by removing alignment waste and packing the edge shape ref into the otherwise-unused entry state slot
- bound scratch overflow slab growth and demote resolved single-path GSS stacks back to flat entries
- attribute memory-budget stops to arena, scratch, runtime heap, or runtime sys growth and expose GSS/scratch/arena metrics in parse-gap reports
- add opt-in transient-reduce scratch checkpoints that materialize the live linear stack and reuse slabs without retaining recycled pointer keys
- follow raw-shape-only sidecar edges at the recycle boundary and defer checkpointing when pending-parent payloads can retain nodes outside that graph
- keep the primary parser benchmark trio allocation-neutral

## Poppler result

The exact 3,447,275-byte JavaScript Poppler witness accepts at EOF with `no_error=true`, `sexpr=true`, and `deep=true` under an explicit 2 GiB parser budget. A separate prebuilt single-Go-runtime probe accepts the same source inside a hard 2 GiB cgroup at 1,729,836 KiB maximum RSS; the combined Go/C oracle remains in an 8 GiB envelope because it retains both giant trees for comparison.

A same-day #221/#222 audit reproduced the memory gains:

| metric | #221 | this PR | delta |
| --- | ---: | ---: | ---: |
| Go B/op | 4,149,796,944 | 3,329,002,256 | -19.8% |
| arena capacity | 1,422,062,592 B | 1,282,135,776 B | -9.8% |
| Go wall time | 13.960 s | 13.673 s | -2.1% |

An earlier candidate sample recorded 11.754 s and a 1.63x Go/C ratio, but later loaded-host samples ranged from 2.92x to 3.46x. The host was unpinned and running above 25 load average, so this PR deliberately does **not** ratchet Poppler timing. The next performance pass should use a pinned quiet host. The shipped 512 MiB parser-budget gap also remains open and explicit in the fleet ledger.

Machine-readable release evidence, source and artifact hashes, exact settings, and that interpretation are checked in at `cgo_harness/perf_scan/evidence/v0.24.0_poppler.json`.

With JavaScript transient reductions explicitly enabled, a 256 MiB checkpoint reduced retained transient scratch from 570,164,432 B to 301,729,936 B (-47.1%) and B/op from 3,390,808,408 to 3,125,558,552 (-7.8%). It cost about 11% Go time and did not materially improve process RSS, so JavaScript transient mode remains opt-in.

## Validation

- Docker-isolated root package: `go test . -count=1`
- focused changed-unit tests for raw-shape packing, GSS demotion, slab bounds, memory attribution, checkpoint recycling, and the pending-parent incompatibility guard
- Docker-isolated `parse_gap_report` / correlator tests
- JavaScript single-grammar gate: 25/25 no-error, 25/25 S-expression, 25/25 deep parity
- primary perf trio: `GOMAXPROCS=1`, `-count=10`, `-benchtime=750ms`, `-benchmem`; full parse remains 978 B/op / 5 allocs, both incremental lanes remain 0 B/op / 0 allocs
- perf-budget validator and status-generator tests
- `git diff --check`

## Stack

Originally stacked on #221; retargeted to `main` after #221 merged.


odvcencio:
Exact-head blocker disposition for 1f181944ae904e3b33f0521db8d042ef8e3e6472

- The Terra checkpoint finding was reproduced before the fix: a raw-shape-only child still pointed into transient storage after checkpoint materialization.
- The fix follows raw-shape-only edges only for the recycle boundary, clones them into the arena, and retargets each packed child while preserving its captured raw-shape reference. Ordinary final-result materialization keeps the prior path.
- The nested raw-only child plus grandchild regression forces immediate reuse of all three transient slots and proves both sidecar pointers and captured refs remain stable.
- Focused count=25, focused Docker, and the full Docker root package passed on the exact head. Hosted correctness/race/parity checks are green.
- The first hosted perf sample missed the incremental threshold by 0.29 percentage points. The exact rerun passed: full +0.14%, incremental edit +2.49%, no-edit +0.06%, allocations unchanged, RSS +3.72%.

The Poppler gates are now separated as required by AGENTS.md:

- Correctness: the exact 3,447,275-byte witness passes no-error, S-expression, and deep Go/C parity in the 8 GiB oracle envelope. The integrated JavaScript corpus gate is also 25/25 on all three axes.
- Runtime memory: a prebuilt single-pass Go probe under a hard 2 GiB cgroup and GOMEMLIMIT=1536MiB accepted program at exact EOF with HasError=false and max RSS 1,729,836 KiB.
- The combined parse-gap process still exceeds 2 GiB because it holds the giant C and Go trees together for deep comparison. That is oracle-harness overhead, not a single-parser production OOM; the 8 GiB correctness run remains the appropriate structural gate.

odvcencio:
Direct exact-head review completed at 4c701f3428169e982c48e5f2714ec953ec311c69 after #221 merged. I inspected the full parser/memory patch and Canopy blast radius. Two actionable issues were corrected before approval: recycle materialization now follows raw-shape-only edges and checkpointing defers when pending-parent payloads could retain transient nodes outside that graph. I also re-ran base/head Poppler samples and removed the invalid 45.9% speed attribution: B/op (-19.8%) and arena capacity (-9.8%) reproduce, while timing is not stable enough to ratchet on the loaded unpinned host. Exact 2048MB-budget deep parity, the hard-2GiB single-runtime gate, the open shipped-512MiB gap, settings, source hash, and artifact hashes are now committed in cgo_harness/perf_scan/evidence/v0.24.0_poppler.json. Focused count=25, focused Docker, perf-budget validation, and status-generator tests pass. No remaining actionable code findings; hosted CI is the remaining gate.

PR #264, opened 2026-07-12
fix(grammargen): dedupe field-map entry runs by production ID

odvcencio:
## Summary

`buildFieldMaps` appended one field-map entry run per *production*, but `compactProductionIDs` collapses productions with identical alias+field fingerprints onto shared ProductionIDs — so every production after the first appended an orphaned, unreachable run (last-writer-wins on `FieldMapSlices[pid]`). The flat `FieldMapEntries` offset inflated until large grammars crossed the uint16 ceiling: bash failed generation at 65,536 entries and dart at 65,538, versus upstream's actual tables of 287 and 290.

Fix is a per-ProductionID seen-guard in `buildFieldMaps` (dedup is safe: the compaction fingerprint includes the field set, so same ID ⇒ identical fields). No blob-format or ABI change; the runtime keys field lookup by child index, so parse output is provably unchanged.

- Shipped grammargen blobs regenerated: `go.bin` 673 → **267** entries (60% were orphans), `swift.bin` 2904 → **384** (87%). `regex.bin` has no field entries and is untouched; ts2go blobs bypass `buildFieldMaps` entirely and are unaffected.
- **bash and dart now generate** (`grammargen doctor` on their lock-pinned grammar.json: `Generate: OK`, no uint16 error; remaining doctor warnings are pre-existing unreachable-rule notices).
- New regression test `TestBuildFieldMapsDedupesSharedProductionIDs`: asserts one run per field-bearing ID, zero orphaned entries, ID 0 reserved. Verified it fails on the pre-fix code.

## Verification

- `go build ./...` — pass
- Go field/byte-identity suite against regenerated go.bin (`TestParseGoFunctionDeclarationFields`, `TestGoZerrorsNormalizerByteIdentity`, `TestParseGoDeclarations/File/NoErrors/Import`, dot-import alias test) — all pass
- `go test ./grammars/ -run 'Swift' -count=1` — pass
- Blob encode/decode suite (`EncodeLanguageBlob|DecodeLanguageBlob`) — 7/7 pass
- `go test ./grammargen/` — 4 failures, all **verified pre-existing on unmodified origin/main** via a baseline run (markdown CST divergences and a separate parse-action-group overflow in markdown_inline).

## Known follow-up (documented, not hidden)

Regenerating go.bin/swift.bin at current HEAD means they now carry `NonTerminalAliasMap` data (a feature added after they were last built), so the already-failing `TestShippedBlobsNonTerminalAliasMapInventory` adds `go` and `swift` to its complaint alongside the pre-existing `caddy` entry. The test demands dedicated alias-map parity coverage per flagged language rather than an allowlist expansion — that coverage is the right follow-up and was deliberately not shortcut here. Confirmed the alias-map content is identical whether built with the buggy or fixed `buildFieldMaps`.


odvcencio:
Red gates root-caused and fixed: the regenerated `swift.bin` changed its SHA-256, so the certified swift retry profile in `grammars/runtime_profiles.go` (pinned per the fail-closed certification design) stopped attaching — exactly as that mechanism is meant to behave. The pin now references the regenerated blob's SHA. Certification evidence transfers because the field-map dedup provably does not alter selected trees (byte-identity and Swift suites verify this on the regenerated blobs); the profile's behavioral claims are about retry-ladder tree selection, which is unchanged. `TestBuiltinExternalScannerRetryProfilesAttach`, `TestBuiltinCompleteAcceptedErrorRetryProfilesAttach`, and `TestBuiltinCompleteAcceptedErrorRetryProfileRequiresCertifiedBlob` all pass locally against the new pin.

odvcencio:
Effectiveness validated end-to-end on the pinned real-corpus lane (merged validation worktree, Docker, lock-pinned corpora):

- Full 55-grammar sweep on this branch: all subtests PASS, including **bash** (generates for the first time; 306s) and **dart** (9.7s).
- First-ever floor rows landed in this PR: **bash 6/25 deep parity** (9/25 no-error — now an honest fidelity work item rather than a generation failure) and **dart 19/25 deep**.
- Floor file now covers 55 grammars, aggregate deep parity 851/1026 (83.0%). The docker wrapper's default skip list shrinks to `ocaml` only (its synthetic-state ceiling is a separate, unrelated limit).
- Note: the run-header `grammar_count`/totals are corrected to reflect the merged metric set; the update-mode header stamping (writes the observed-count rather than the merged total) and the rebase-gate's refusal to accept new-grammar additions as "stronger" are two small harness follow-ups surfaced by this validation.

odvcencio:
Adversarial review disposition (all findings addressed):

- **Dedup logic**: reviewed SAFE on all six refutation angles — fingerprint invariant verified (`ChildIndex` + full `FieldName` per field, sole ProductionID writer), first-writer/last-writer equivalence proven (runtime keys by ChildIndex; no duplicate ChildIndex runs exist in either grammar), skipped-entry ordering safe, uint16 checks and C emission consistent. Reviewer independently decoded both blob versions: zero field-content differences across all 117 go + 191 swift field-bearing IDs.
- **Blocking finding — unverified alias-map payload**: regenerating go.bin/swift.bin at HEAD ships grammargen-derived `NonTerminalAliasMap` rows (0→233 go, 0→593 swift) that the inventory tripwire rightly blocks without derivation-parity coverage. Resolved by folding the dedicated coverage into this PR (was #266): derivation-parity gates for go/swift/caddy mirroring `TestLuaNonTerminalAliasMapParity`, live-parse regression tests, registry updated. Inventory test is green on this branch — including caddy, which was already red on main before this PR.
- **Body correction**: the earlier claim "alias-map content is identical pre/post-fix" was true of the derivation code (independent of `buildFieldMaps`) but misleading for the shipped artifacts — the old blobs predated the alias-map feature entirely. The shipped payload change is now covered by the parity gates above.
- **Scope clarification (bash/dart)**: this PR removes their *generation* ceiling for the grammargen self-hosting lane (validated: both pass the pinned real-corpus sweep; first floor rows landed here). Their *shipped* blobs remain ts2go-extracted by design — promoting languages to grammargen-first blobs is the rollout ratchet's separate, gated decision.
- **Test-strength finding**: added `TestCompactProductionIDsSharesIDOnlyForIdenticalFieldSets`, which drives the real `compactProductionIDs` path and pins the shared-ID ⇒ identical-fields invariant end-to-end, alongside the existing isolated unit test.

PR #268, opened 2026-07-12
optimize(glr): compact gssNode layout and recycle transient GSS slabs

odvcencio:
## Summary

Completes and lands two stacked GLR memory optimizations that were left mid-integration on `codex/gss-node-layout`:

- **gssNode layout compaction** (7e63f411): `extraLinks` slice → pointer-backed array with uint8 count/cap (eliminating the slice header on the common single-link node), `depth` int → uint32, `aggVisGen` uint64 → `aggVisValid` bool. `TestGSSNodeLayoutSizeBudget` enforces the 64-byte target on 64-bit (52 on 32-bit). Mutation goes through `setExtraLink`/`appendExtraLink`/`extraLink` helpers.
- **Transient GSS slab recycling** (79f53985): reuses GSS storage after linear demotion. Address-keyed merge/recovery caches are invalidated before reuse (`invalidateGSSPointersForReuse`, prefix-path reset, agg generation bump); fingerprinted spine memoization is preserved.
- **Integration fixup** (9238630e): aligns the recycle tests with the compacted field set.

Semantic audit of the seam between the two: both `recycleForParse()` and `reset()` zero slab slots wholesale via `clear()`, so the compacted fields (`extraLinks` ptr/count/cap, `aggVisValid`) are reset by construction, with explicit re-zeroing on both `allocNode` paths as defense in depth.

## Verification

- `go build ./...`, `go vet .` — clean
- GSS/recycle/demotion suite: 85 subtests PASS (size budget, recycle-reuse, pointer-holder invalidation, fingerprinted-spine retention, merge/reduce/forest)
- Full root package: ok (9.9s)
- `go test -race` over GSS/GLR/recycle machinery: clean
- Quiet-host A/B (pinned core, count=10, vs same-morning main baseline): full parse −0.17%, one-byte edit −1.54%, no-edit +2.07% — timing-neutral within noise, allocs identical (9/0/0). The win is per-node GSS memory on GLR-heavy paths, enforced by the size-budget test; the canonical lanes confirm no common-path regression.

### DIFF
--- a/glr.go
+++ b/glr.go
@@ -162,6 +162,7 @@ type glrMergeScratch struct {
 	cRecoveryCost     bool
 	audit             *runtimeAudit
 	equivEpoch        uint32
+	gssPointerEpoch   uint32
 	equivCache        []glrNodeEquivCacheEntry
 	stackEquivCache   []glrStackEquivCacheEntry
 	spineEquivCache   []glrSpineEquivCacheEntry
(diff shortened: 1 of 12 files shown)


odvcencio:
Review comment addressed: `maxMainLinkCount` is 6 today (`maxStacksPerMergeKey`), well inside uint8, and `appendExtraLink` already panics at the limit — but the concern is right that nothing stopped a future config from outgrowing the field width silently. Added a compile-time assertion (`const _ = uint8(maxMainLinkCount - 1)`) so any growth past 256 fails the build instead of overflowing at runtime.

PR #272, opened 2026-07-12
add(parser): layered memory-budget containment — volume polling, in-merge checks, hard ceiling

odvcencio:
## Summary

Closes the containment escape root-caused in the budget RCA: the runtime budget was armed on every path but sampled only by iteration count (mask 255 above 64 MiB budgets — measured 16-26x overshoot), and the GLR stack-merge lane could allocate multi-GB and grind O(survivors²) without ever reaching a poll site. Three layers, per the RCA's fix shape (deliberately not a mask tighten, which would truncate legitimate bounded-overshoot parses):

1. **Volume-triggered polling** — a real `ReadMemStats` check fires whenever tracked arena growth exceeds 64 MiB since the last check, bypassing the poll mask; lanes that allocate GBs in few iterations get checked regardless of iteration count.
2. **In-merge polling** — `mergeStacksWithScratchLargeCap`'s survivor loop checks the budget every 4096 comparison trips and stops early with `ParseStopMemoryBudget`, converting the previously unpollable fatal window.
3. **Absolute hard ceiling** — `GOT_PARSE_MEMORY_HARD_CEILING_MB` (default 2048, 0 = off), decoupled from soft-budget overshoot tolerance; margin over the Poppler completion point locked in by test (≥256 MiB asserted).

Also corrects the stale "overshoot stays small" comment.

## Verification

- Budget/scratch/arena suite: **115 pass, 0 fail**; race clean over Budget/Merge/GSS; full root suite stable across repeated runs.
- End-to-end witness: bare `NewParser().Parse` of a synthetic 300k-entry Go table literal at a 96 MiB budget now stops with `ParseStopMemoryBudget` at 2.6-4.0x budget heap growth (previously ballooned unbounded).
- **Poppler ledger gate (quiet-host A/B, exact 3,447,275-byte witness, certified 2 GiB budget):** branch and main identical — both complete full-span, `hasError=false`, 15.89 s vs 15.58 s, peak RSS within 0.03%. The certified bounded-overshoot behavior is preserved.
- Canonical benchmark neutral (p=0.589, allocs unchanged).

## Follow-up flagged (separate lane)

`walkGoCompatSubtree` runs budget-blind at result finalization on recovered Go trees and can OOM independently of the parse loop (reproduced at 11-16 GiB pre- and post-fix); tracked as its own containment item.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ for tags and release notes while still in `0.x`.
 
 ### Added
 
+- Memory-budget containment is now layered: volume-triggered polling
+  forces a real budget check whenever tracked arena growth exceeds 64 MiB
+  since the last check (bypassing the iteration-count poll mask), the GLR
+  stack-merge survivor loop polls the budget mid-grind, and a decoupled
+  absolute hard ceiling (`GOT_PARSE_MEMORY_HARD_CEILING_MB`, default
+  2048, 0 = off) stops runaway growth regardless of soft-budget
+  overshoot tolerance. A bare-`Parse` giant-table witness now stops with
+  `ParseStopMemoryBudget` at 2.6-4x budget instead of ballooning; the
+  Poppler witness still completes full-span under its certified 2 GiB
+  budget.
+
 - A hard zero-cliff gate for nightly fleet perf sweeps, with a
   hard-gate-only mode on the perf-scan budget checker; the scheduled
   perf-scan gate is disabled in favor of the nightly hard gate.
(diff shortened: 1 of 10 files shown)


odvcencio:
race_root failure root-caused and fixed: under `-race`, the detector's shadow memory inflates `MemStats.Sys`, so the sys-growth guard legitimately fires before the heap-growth guard — the volume trigger and budget stop behaved correctly; the test over-specified the stop source. The assertion now accepts either runtime source (with the reasoning documented inline). Verified green with and without `-race` locally.

PR #273, opened 2026-07-12
optimize(parser): elide raw-shape capture on single-stack parses

odvcencio:
## Summary

Raw-shape capture and content hashing (~8% of canonical full-parse CPU per the pinned quiet-host profile) are pure write-only work on parses that never fork: every raw-shape reader is gated on fork, forest, recovery, checkpoint-recycle, or diagnostics. This elides capture while a parse has only ever had a single GLR stack and has not entered error recovery.

Mechanism: `captureRawShape` takes the GSS scratch and skips as its first line when `singleStackMode && !everForked` — one choke point covering all production call sites. `everForked` is a sticky per-parse latch set at every single-stack-mode writer, explicitly at the conflict-fork decision point (fork stacks can reduce within the deciding iteration), and at the `cHandleError` recovery funnel (recovery both spawns versions and reads shapes; latching at entry protects the entire episode including later re-collapse).

Safety story, adversarially reviewed and then strengthened: memory safety was confirmed on all six review angles (readers nil-guarded, all-or-nothing per node). Behavior preservation does **not** rest on the materialized comparison fallback being sign-preserving — an included abstract test proves it is not in isolation. It rests on structural pointer-sharing: any node built while elision is active predates the parse's first fork/recovery event and is therefore only ever compared to itself. Evidenced by:
- a forced-descent differential (shared root symbol, hidden-child pre/post-flatten count mismatch, comparison descends into the elided prefix): identical trees gate-on vs gate-off;
- a recovery differential (C-recovery entered from a single-stack prefix): identical trees;
- the original pure-single-stack and prefix-then-fork differentials.

A diagnostics toggle (`SetRawShapeElisionDisabledForDiagnostics`, matching existing plain-bool toggle precedent) forces pre-change capture for A/B work.

## Verification

- `go build ./...`, `go vet ./...` — clean; full root suite ok; RawShape/Forest/Recover/Incremental focused suites pass; `-race` over Recover/RawShape/GSS clean.
- Reclaim accounting gates updated: nonzero raw-shape bytes now asserted on a genuinely forking grammar; new zero-bytes assertion for single-stack parses; breakdown-sum invariant balanced.
- Quiet-host receipt (pinned core, count=10): full parse 12.246 → 11.654 ms, **2.142x → 2.039x C**, allocations unchanged (9/0/0); incremental lanes neutral. Toggle-isolated contended A/B: −7.46% (p=0.015) post-recovery-gate, consistent with −7.90% pre-gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,13 @@ for tags and release notes while still in `0.x`.
   maps, alias and supertype query tables are byte-identical, and supertype
   query predicates match by concrete descendant so observable trees and
   captures are unchanged. Canonical quiet-host full parse improves ~3.4%.
+- Raw-shape capture and content hashing are elided while a parse has only
+  ever had a single GLR stack and has not entered error recovery; capture
+  resumes permanently at the first fork or recovery event. Shape-dependent
+  tie-breaks are unaffected: elided prefix nodes are only ever compared to
+  themselves (structural pointer-sharing), evidenced by forced-descent and
+  recovery differential tests. Canonical quiet-host full parse improves
+  ~4.8% with allocations unchanged (9/0/0 preserved).
 
 - gssNode layout compacted to a 64-byte budget on 64-bit targets
   (pointer-backed extra links with uint8 count/cap, uint32 depth,
--- a/glr_forest.go
+++ b/glr_forest.go
@@ -2822,7 +2822,12 @@ func (p *Parser) parseForest(arena *nodeArena, source []byte, captureExternalChe
 							parent.setExtra(true)
 						}
 						parent.dynamicPrecedence = int32(score)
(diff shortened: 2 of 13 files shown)


odvcencio:
Post-recovery-gate quiet-host receipt (pinned core, count=10, branch @ d6becb1d): full-parse median ~11.36 ms — ~1.99x the same-host C baseline, the first sub-2x reading on the canonical workload. Allocations unchanged (9/op). Consistent with the pre-gate 11.654 ms receipt within run variance; the recovery latch costs nothing on the happy path.

odvcencio:
Both review comments addressed in one change: the diagnostics toggle is now an `atomic.Bool` (matching the `EnableArenaBreakdown` precedent), so concurrent parsers observe a consistent value with no data race, without threading per-parser state through the scratch structs for a diagnostics-only hook. Elision-path read uses `.Load()`; differential tests unchanged and green.

PR #331, opened 2026-07-14
fix: validate rewriter ranges

ShiroKSH:
## Summary

- Validate source ranges before applying queued rewrites.
- Return an error rather than panic when `ReplaceRange` receives an invalid range.

## Root cause

`Rewriter.Apply` validated overlap but sliced using public byte offsets without checking that they were ordered and within the source.

## Validation

- `go vet .`
- `go test . -run '^TestRewrite' -count=1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


odvcencio (approved):
Validated the range checks, API documentation, and Unreleased entry. Focused Docker rewrite coverage passes.

PR #338, opened 2026-07-16
fix(incremental): preserve bounded GLR reuse and retry correctness

odvcencio:
## Summary\n\nThree incremental-parsing correctness fixes:\n\n- Preserve the configured bounded GLR width during incremental parsing instead of hard-capping stacks and merge survivors to 2. The language-specific C# minimum remains narrowly applied at its existing full-parse call site.\n- Reject leaf reuse when a reused leaf's nonzero stored parse state disagrees with the current unique shift target. State-zero metadata retains the conservative unique-shift fallback.\n- Keep the first-pass incremental tree when an accepted-error retry ties it on every tree-quality axis. Retry allocation bookkeeping no longer displaces an equally good incremental result; retry-vs-retry ladder selection remains unchanged.\n\nTogether these changes remove width-sensitive truncation, stale leaf-state reuse, and spurious full-retry replacement while retaining bounded parser limits.\n\n## Verification\n\n- Focused incremental, reuse, retry, pass-accounting, and Authzed associativity tests passed.\n- The stacked #341 CI run exercised this branch with the full root/package race matrix, including the previously failing Authzed lane.\n- Docker four-edit admission preserved exact fresh/incremental trees in both directions for the admitted token-class and recovery-deletion cases.\n- Changelog entries are included under Unreleased.\n\nThe remaining same-line branch-order divergence is tracked separately and is not claimed fixed here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ for tags and release notes while still in `0.x`.
   rejects reused leaves whose stored parser state conflicts with the current
   shift. This prevents stale leaf context and over-aggressive two-stack
   pruning from changing selected trees on token-class and recovery edits.
+- An incremental parse whose full-parse retry produces no strictly better
+  tree now keeps its first-pass result instead of replacing it with a
+  quality-tied fresh tree. This stops spurious `incremental_parse_full_retry`
+  reporting on grammars whose intended trees contain ERROR productions and
+  legitimately use the full GLR width.
 - Multiline tree edits now keep node byte and point ranges aligned with the C
   runtime across insertions, deletions, and replacements.
 - Rewriter edits now reject reversed and out-of-source byte ranges instead of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ for tags and release notes while still in `0.x`.
 
 ### Fixed
 
+- Incremental parsing now preserves the configured bounded GLR width and
+  rejects reused leaves whose stored parser state conflicts with the current
+  shift. This prevents stale leaf context and over-aggressive two-stack
+  pruning from changing selected trees on token-class and recovery edits.
 - Multiline tree edits now keep node byte and point ranges aligned with the C
   runtime across insertions, deletions, and replacements.
 - Rewriter edits now reject reversed and out-of-source byte ranges instead of
--- a/incremental.go
+++ b/incremental.go
@@ -147,13 +147,6 @@ func (c *reuseCursor) releaseNodeRefs() {
 	}
 }
 
-func rejectForestLeafStateMismatch(n *Node, nextState StateID) bool {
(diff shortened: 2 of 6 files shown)


odvcencio:
CI bisect result (local reproduction of the `race_packages grammars-a-c` failure):

- `TestAuthzedTokenSourceIncrementalRetryHandlesAssociativityFirstByteToggle` fails because the **GLR cap unclamp** (parser.go/parser_limits.go), not the leaf-state rejection (incremental.go), makes the second incremental parse fall back to `incremental_parse_full_retry`. With main's clamp restored and only the leaf fix applied, the authzed test passes — but all four canonical Go incremental edits fail, so the unclamp is load-bearing and the PR cannot be split.
- Mechanism: with configured caps (8 stacks / 6 merge survivors) the incremental dispatch on the authzed associativity witness selects a strictly worse (higher-error-cost) tree than the full retry, while the old hard 2/2 clamp incidentally resolved the recovery directly. Wider GLR width should never select a worse tree — this is a width-sensitive recovery/selection defect that the clamp was masking.
- Final trees remain correct in both configurations (shape and S-expression checks pass); the regression is the direct-resolution invariant plus the extra full-parse cost.

Keeping this PR as draft until the width-sensitivity is root-caused. The leaf-rejection instrumentation confirmed the nonzero-state rejection never fires on the authzed witness, so the two changes are independent.

odvcencio:
Root cause found and fixed; PR is ready for review again.

The authzed regression was not the cap unclamp itself. The retry ladder's final tie-break preferred the tree with fewer `NodesAllocated` — a parse-run bookkeeping metric, not a tree-quality axis. Sequence on the authzed witness: the unclamped incremental parse legitimately reaches the initial stack budget (`MaxStacksSeen >= initialMaxStacks`), which arms `shouldRetryAcceptedErrorParse` because the grammar's intended tree contains an ERROR production; the retry then produces a byte-identical, quality-tied tree that displaced the first pass purely because a fresh parse allocates fewer nodes than an incremental one — falsely reporting `incremental_parse_full_retry` and discarding a good tree. The old 2/2 clamp had only been masking this by keeping `MaxStacksSeen` under the trigger threshold.

Fix (`parser_retry.go`): for the incremental retry origin, the ladder result only replaces the first-pass tree when it is strictly better on a tree-quality axis (`preferRetryTreeOverFirstPass`). Retry-vs-retry selection inside the ladder keeps the existing `NodesAllocated` tie-break, so ladder progression and pass accounting are unchanged (`TestCertifiedDuplicateWideRetryPreservesPassAccounting` passes unmodified).

Verification: authzed suite green including `-race` on the previously failing test; root retry/incremental suites green; the locked four-edit canonical Go matrix is unchanged (token-class and recovery-deletion exact both directions; the two known-open edit classes are separate, already-diagnosed defects).

PR #339, opened 2026-07-16
perf(parser): reduce Odin arena preallocation

odvcencio:
## Summary

Reduce first-pass node-arena over-allocation for large token-sparse ASCII inputs parsed by the exact locked Odin grammar. The runtime profile is bound to the built-in grammar blob identity; custom, same-name, stale-blob, adapted, and unrelated grammars keep the existing allocation policy. Non-ASCII and unexpected control bytes fail open.

The cap uses a complete structural-density scan, preserves learned parser hints, and leaves the TypeScript large-fourslash policy unchanged. Callers can explicitly opt custom language artifacts into the policy.

## Results

On the locked 6.2 MB Odin test-vector witness, the quiet-host candidate improves full-parse time by 6.20%, reduces initial arena allocation by 71.81%, reduces B/op by 90.31%, and reduces allocations/op by 53.66%. Selected-node count and deep-tree digest are unchanged. A dense non-Odin generated-code control retains identical arena, allocation, node, and digest metrics.

## Validation

- Exact/stale/custom runtime-profile identity tests
- Exhaustive ASCII classification and fail-open tests
- Large ASCII, non-ASCII, and dense-control deep-tree digest tests
- Focused Docker density/profile gates
- Direct cgo/static-C `TestParityFreshParse/odin` exhaustive parity
- Independent code review and quiet-host benchmark receipt


odvcencio:
Review outcome: holding this PR before merge.\n\nBlocking issues:\n\n1. The density estimate multiplies int values and overflows on 32-bit targets for ordinary multi-megabyte inputs. The calculation needs overflow-safe unsigned or 64-bit arithmetic with a bounded conversion back to int.\n2. The current structural-byte whitelist omits common operators. Dense valid code such as a long a+a+a+... expression is therefore classified as sparse, potentially forcing repeated arena growth and increasing RSS or soft-budget stops. Please make the classifier conservative and add an adversarial dense-code control.\n3. The reported allocation and timing win needs a reproducible receipt with base/head distributions, allocations, RSS, exact-tree identity, and a dense negative control because the canonical CI fixtures do not enter the >=1 MiB path.\n\nAll current checks are green, but they do not cover these failure classes.

PR #341, opened 2026-07-16
fix(glr): make incremental scoring and culling match fresh parses

odvcencio:
## Summary

Two selection-parity fixes make incremental GLR decisions match a fresh parse of the same structure:

- Reused subtrees credit their cumulative dynamic precedence to the stack score. Fresh parsing accumulates the same credit while reducing; reuse previously omitted it, allowing merge, cull, and result selection to diverge after an edit.
- The GLR stack-cull trigger is arena-class-invariant. Fresh and incremental parses now use the same language-specific slack window, while preserving the existing C# exception.

## Validation

- Added `TestReuseNodePreservesFreshDynamicPrecedenceSelection`, which constructs a non-leaf reduced subtree, scores the fresh path through `markReduceApplied`, scores the incremental path through `reuseNode`, and checks actual primary-stack selection. Removing only the reuse credit flips the selected lineage and fails the test.
- Updated `TestGLRStackCullTrigger` to pin the arena-invariant contract.
- Focused incremental, reuse, culling, dynamic-precedence, GSS, forest, and selection tests pass in Docker.
- The isolated Go real-corpus gate remains 25/25 for no-error, S-expression, and deep-tree parity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ for tags and release notes while still in `0.x`.
   quality-tied fresh tree. This stops spurious `incremental_parse_full_retry`
   reporting on grammars whose intended trees contain ERROR productions and
   legitimately use the full GLR width.
+- Reused subtrees now credit their cumulative dynamic precedence to the GLR
+  stack score, so score-sensitive merge, cull, and result-selection decisions
+  in an incremental parse match a fresh parse of the same structure.
+- The GLR stack-cull trigger no longer depends on the arena class: incremental
+  parses keep the same cull slack window as fresh parses, which previously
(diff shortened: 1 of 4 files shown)


odvcencio:
Review outcome: the implementation is mechanically sound, but this PR needs one standing regression before merge.\n\nThe cumulative dynamic-precedence credit in reuseNode is not covered by a committed test: the existing cull-trigger test guards only the second change, and the behavioral matrix cited in the PR is on a separate branch. Please add a focused fresh-vs-incremental reuse witness whose result diverges if the reused subtree's cumulative precedence is not credited.\n\nAfter #338 lands, this PR should be retargeted to main and the full CI/race matrix rerun on the updated synthetic merge.

PR #346, opened 2026-07-16
add(compat/smacker): Add smacker compatibility shim

odvcencio:
## Summary

This branch introduces a compat/smacker package that mirrors the public API of the unmaintained github.com/smacker/go-tree-sitter, enabling a drop-in import swap for projects dependent on that CGo binding. It wraps gotreesitter's pure-Go nodes to transparently thread language and source buffers, preserving the argument-free method signatures that smacker consumers expect. The package covers the complete Parser, Tree, Node, and Query/QueryCursor surface, accompanied by per-grammar subpackages mirroring smacker's layout.

## Changes

- Add compat/smacker package wrapping gotreesitter to mimic smacker/go-tree-sitter API, enabling drop-in import swaps for CGo-dependent projects.
- Implement wrapper types (Node, Tree, Parser, Query, QueryCursor) that thread language and source context transparently, preserving argument-free method signatures and null-safety.
- Generate per-grammar subpackages (smacker/golang, smacker/python, etc.) exposing GetLanguage() functions that mirror smacker's import layout.
- Add smacker_test.go covering full parsing, tree navigation, query captures, null safety, and incremental reparsing via the stateful Parser path.
- Update README.md with a link to the official migration guide for consumers switching from smacker.

## Testing

- Run `go test ./compat/smacker -v` to verify wrapper correctness and API parity against gotreesitter.
- Swap `import github.com/smacker/go-tree-sitter` to `import github.com/odvcencio/gotreesitter/compat/smacker` in a downstream dependency or local fork and confirm compilation and existing test suites pass without modification.


odvcencio (commented):
Not ready to merge despite green hosted checks.

The advertised drop-in smacker contract is source-incompatible and incomplete. In particular, Parser.Parse returns two values here while upstream returns one; package Parse, TreeCursor, ParseInput APIs, QueryCursor.NextCapture, and substantial Node/Language/Query surface are absent. Tree.Close is also a no-op even though the underlying gotreesitter Tree requires Release for arena accounting, and ParseCtx cannot honor cancellation that occurs during a no-deadline parse. The migration URL currently returns 404.

Before merge: either implement the exact upstream source contract with compile-contract/downstream fixtures, or explicitly reframe this as a partial helper and remove drop-in claims; make Close release idempotently, fix live cancellation, publish or remove the broken guide, and add the Unreleased changelog entry. Required focused gates include compat package tests, lifecycle and cancellation tests, upstream signature fixtures, and exact full/incremental comparisons for Go plus an external-scanner grammar.

odvcencio:
Independent review blocks this as a drop-in compatibility layer. Parser.Parse has a different signature; ParseCtx uses the non-strict parser path and can return a partial tree with nil error after cancellation; Tree.Close does not release the underlying arena; QueryMatch changes the public shape; and substantial parser, input, node, cursor, iterator, query, and grammar-package surface is absent. The package tests are compile-only in normal CI, and the PR has no Unreleased changelog entry. Please either redesign this around an upstream API compile-contract plus cancellation/lifetime tests, or reposition it explicitly as a limited migration adapter and narrow the public claim. I would keep it unmerged until that contract is resolved.

odvcencio:
Closing after independent review. The advertised drop-in contract does not compile against a pinned upstream smacker consumer, cancellation can return a partial tree with nil error, and Tree.Close does not release the underlying arena. The hosted checks did not exercise those contracts. The branch is preserved if we later rebuild this as an authenticated, lifecycle-safe adapter; it should not merge in its current form.

PR #349, opened 2026-07-16
fix(query, tree): align missing, property, and descendant semantics

odvcencio:
## Summary

Aligns three query/tree surfaces with the locked upstream C runtime: strict `MISSING` patterns, host-visible inert property predicates, and descendant byte/point range selection. The final branch is intentionally limited to behavior authenticated by focused Go tests and locked-C differentials.

## Changes

- Match scalar and alternation `MISSING` patterns only against missing nodes, including anonymous token-qualified and named-symbol-qualified forms.
- Preserve MISSING identity through alternation compilation and keep those alternations on the exact matcher instead of the symbol-only index.
- Expose inert `#is?` and `#is-not?` predicates through documented public metadata accessors without filtering matches.
- Match upstream descendant range behavior for reversed ranges, out-of-bounds ranges, token/line boundaries, and zero-width missing children.
- Retain contained nil-on-miss walks for incremental and byte-offset helpers whose existing semantics must not change.
- Record the user-visible fixes under CHANGELOG Unreleased.

## Validation

- Full root package tests
- Focused MISSING/property/descendant tests
- `go vet .`
- Four-fixture real-Go arena board: byte-for-byte neutral versus main
- Locked-C MISSING scalar/alternation differential: `harness_out/docker/20260717T062751Z`
- Locked-C descendant byte-range differential: `harness_out/docker/20260717T061625Z`
- Fresh hosted CI on the repaired head

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ for tags and release notes while still in `0.x`.
   validates complete ABI-15 supertype metadata before emission. Lowercase
   keyword leaves are classified from parser-reachable ownership like
   tree-sitter.
+- Query `MISSING` patterns now test missing nodes, and inert `#is?`/`#is-not?`
+  properties are available through public metadata accessors. Descendant range
+  walks now match upstream behavior for reversed ranges and zero-width missing
+  children.
 - Highlight queries now resolve supported built-in inheritance chains across
   registration order and same-name replacements without duplicating cyclic
   queries. Incompatible locked grammar/query pairs remain fail-closed.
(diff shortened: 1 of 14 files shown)


odvcencio:
Final review of the first repair pass still blocks merge:

- `TestParityD3SupertypePatternMatchesSubtypes` is red: nested/general Go supertypes are not represented completely.
- Hidden-supertype edge metadata is dropped by transient-parent materialization and can become stale after compatibility child rewrites.
- Quantified matching silently converts the 65,536-step budget into a non-match instead of exposing exhaustion.
- The four-fixture real-Go board shows arena growth of 0.85–4.06%, so the current allocation test does not establish neutrality.

Descendant-range parity and public `#is?`/`#is-not?` metadata are sound. A second repair pass is underway; this remains unmergeable until the three correctness issues and arena regression are resolved, then it must be rebased on current main and re-gated.

PR #375, opened 2026-07-19
fix(go parser): fix new/make sole argument parsing to match C oracle

odvcencio:
## Summary

Resolve a C-oracle divergence where the sole or leading argument of Go's `new` or `make` builtins (e.g., `new(dirInfo)`) was incorrectly parsed as an `identifier` instead of a `type_identifier`. Tree-sitter-go's grammar declares an explicit ambiguity for this position resolved statically toward `_simple_type` by the C runtime, while gotreesitter's GLR table resolves it toward `_expression`. This PR introduces a narrow, byte-identity-preserving post-parse normalization to align the GLR output with the C baseline without risking broader impact on grammargen's conflict resolution.

## Changes

- Add `normalizeGoNewMakeTypeArgument` to reclassify the bare identifier leading argument of `new`/`make` calls to `type_identifier`.
- Update `normalizeGoReturnedTreeCompatibility` to invoke the new normalization and explicitly check memory budget at stage boundaries.
- Add regression tests verifying correct symbol types for `new`/`make` with user-defined and primitive types.
- Add shadow-test to ensure the fix does not leak into ordinary function calls or user-defined functions sharing builtin names.

## Testing

- Run local regression tests: `go test ./... -run 'TestParseGoNewMake' -count=1`
- Run Go grammar parity in Docker: `bash cgo_harness/docker/run_single_grammar_parity.sh go`

## Review Focus

Verify the `goCompatMemoryBudgetStopReason` control flow additions in `normalizeGoReturnedTreeCompatibility` preserve existing behavior, and confirm the `new`/`make` normalization handles all expected syntactic shapes without altering AST structure or source bytes.

### DIFF
--- a/parser_go_test.go
+++ b/parser_go_test.go
@@ -162,6 +162,66 @@ func TestParseGoFunctionDeclarationFields(t *testing.T) {
 	}
 }
 
+// TestParseGoNewMakeSoleArgumentIsTypeIdentifier is a regression test for a
+// C-oracle divergence found on real Go stdlib (os/dir_unix.go: `new(dirInfo)`):
+// tree-sitter-go's grammar declares `_simple_type`/`_expression` as an
+// explicit ambiguity for the sole/leading argument of a call to the "new" or
+// "make" builtins (special_argument_list's `_type` slot overlaps
+// argument_list's `_expression` slot for a bare identifier). C's LALR table
+// resolves this deterministically toward `_simple_type` (type_identifier);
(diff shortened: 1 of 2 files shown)


odvcencio:
Closing this implementation after review. The locked Go parser table contains both reductions for this declared conflict, so the divergence is in GLR/result selection rather than static table generation. A post-tree retag would hide that engine bug and add another compatibility walk. We will replace it with a runtime-level fix backed by a focused locked-C differential and incremental coverage.

PR #377, opened 2026-07-19
add(incremental): Add incremental invariant correctness gate and corpus

odvcencio:
## Summary

Introduce a pure-Go, default-on correctness gate that validates the incremental parser against a curated real-world corpus. It enforces that ParseIncremental matches Parse whenever the fresh parse is fully clean, turning previously silent structural corruption into explicit, tracked CI failures with a staleness ratchet.

## Changes

- Add `incremental_invariant_gate_test.go` performing a full-span, fresh-clean invariant sweep over single-byte DELETE, INSERT, and REPLACE edits at fixed strides across multiple language corpora.
- Replace unreliable `HasError()` checks with explicit recursive `IsError()`/`IsMissing()` counting to accurately catch silent corruption where trees appear structurally valid but are divergent.
- Introduce `testdata/incremental_allowlist.json` to track known divergences by language, edit class, and divergence signature, failing builds when unlisted divergences appear or allowlist entries become stale.
- Add real-world corpus files under `testdata/incremental_gate/` (Python, JavaScript, Go, JSON, Rust) calibrated with fixed strides to complete full sweeps in 1-3 minutes.

## Testing

- go test -run TestIncrementalInvariantGate

### DIFF
--- a/incremental_invariant_gate_test.go
+++ b/incremental_invariant_gate_test.go
@@ -167,8 +167,12 @@ func TestIncrementalInvariantGate(t *testing.T) {
 		ok := matched[key]
 		matchedMu.Unlock()
 		if !ok {
-			t.Errorf("STALE ALLOWLIST ENTRY (ratchet failure): %s | %s | %s did not reproduce anywhere in this sweep -- "+
-				"remove it from testdata/incremental_allowlist.json (recorded note: %q)",
+			t.Errorf("STALE ALLOWLIST ENTRY (ratchet failure): %s | %s | %s did not reproduce anywhere in this sweep. "+
+				"Before removing it from testdata/incremental_allowlist.json, confirm WHY it stopped reproducing: "+
+				"(a) the incremental bug was genuinely FIXED (remove it), or (b) the fresh-clean observation window "+
+				"CLOSED -- a grammar/parser change made the fresh parse at those sites contain ERROR/MISSING so they "+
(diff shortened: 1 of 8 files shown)


odvcencio:
Independent review verdict: **NO-GO / revise**.

Blocking findings:

- The allowlist key is only `(language, edit class, lossy signature)`, so a new defect at another fixture/offset can silently match an existing entry; large child-count deltas are bucketed and nil mismatches lose the non-nil node type. Ratchet an exact authenticated hit set (fixture hash/path, edit coordinate/class, exact divergence digest) instead.
- The sweep does not release baseline, fresh, copied-old, or incremental trees. This materially inflated `race_root` runtime and risks retaining the entire corpus matrix. Release every path and move the exhaustive sweep to a dedicated required correctness lane.
- Subset runs ratchet entries for languages that were not executed. A JavaScript-only run passes its sweep and then falsely marks all Python entries stale. Scope the ratchet to executed fixtures/languages.
- Merge #379 first, rebase, and rederive the allowlist; its Python fallback should make the current Python entries stale.

Also required: compare fields/points/extra/error metadata, fail closed on parse errors instead of discarding them, authenticate fixture provenance/hashes and expected site counts, remove duplicate corpus data, and add an Unreleased changelog entry.

The deterministic edit generation and hermetic gate concept are worth preserving after revision.

PR #378, opened 2026-07-19
add(grammargen): Add context-aware generation and JSON-based grammar extension

odvcencio:
## Summary

Introduces `ExtendGrammarJSON` for data-driven grammar extension via JSON deltas, parallel to the existing `ExtendGrammar` closure API, and `GenerateWithReportContext` for context-aware generation that consolidates diagnostics and language compilation into a single cancellable pass. These additions allow non-Go callers to customize grammars via standard JSON fragments and eliminate redundant LR table construction for callers needing both reports and compiled languages.

## Changes

- `grammargen/extend_grammarjson.go`: Add `ExtendGrammarJSON(baseJSON, deltaJSON)` to build derived grammars from two JSON documents. Implements documented merge rules (add/override rules, union extras/conflicts/externals, replace word token, append new precedences, resolve reserved word sets) without mutating the base document.
- `grammargen/diagnostics.go`: Add `GenerateWithReportContext(ctx, g)` to perform a single LR generation pass that produces both the diagnostic report and the compiled `Language`. Accepts a `context.Context` for cancellation/deadlines, and prevents the 2x wall-clock overhead of calling `GenerateLanguage` and `GenerateWithReport` independently.
- `grammargen/grammar.go`: Refactor `ExtendGrammar` to extract `cloneGrammarForExtend`, a shared deep-copy helper that preserves base grammar tuning flags and structure for both the closure and JSON extension entry points.
- `grammargen/*_test.go`: Add comprehensive tests verifying JSON delta merge correctness, byte-for-byte parity with `ExtendGrammar`, single-pass performance benefits, and prompt cancellation behavior during heavy grammar generation.

## Testing

- `go test ./grammargen -v -run 'TestExtendGrammarJSON|TestGenerateWithReportContext'`
- `go test ./grammargen -v -run 'TestExtendGrammarJSONParityWithExtendGrammar'` (verifies closure vs JSON delta parity)

### DIFF
--- a/grammargen/diagnostics.go
+++ b/grammargen/diagnostics.go
@@ -1083,7 +1083,32 @@ func generateWithReport(g *Grammar, opts reportBuildOptions) (*GenerateReport, e
 
 // GenerateWithReport compiles a grammar and returns a full diagnostic report.
 func GenerateWithReport(g *Grammar) (*GenerateReport, error) {
-	return generateWithReport(g, reportBuildOptions{
+	return GenerateWithReportContext(context.Background(), g)
+}
+
+// GenerateWithReportContext is like GenerateWithReport but accepts a context
+// for cancellation and performs a SINGLE LR generation pass that produces
+// both the diagnostic report and the compiled *gotreesitter.Language.
+//
+// The returned report's Language field already holds the compiled Language
+// (same as GenerateLanguage(g) would return) — callers that need both the
+// diagnostics (Conflicts, Warnings, SplitCandidates, ...) and the compiled
+// Language should read report.Language here instead of calling
+// GenerateLanguage separately, which would otherwise trigger a second,
+// redundant full LR generation pass (state construction, conflict
+// resolution, lex DFA construction — all done twice for no benefit). This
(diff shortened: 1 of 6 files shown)


odvcencio:
Closing for now after independent review. The latest revision repairs the named-precedence, external-token deduplication, reserved-word, and timing-assertion findings, but the context path still discards a post-split conflict error and falls back through an uncancellable table rebuild. The proposed JSON merge dialect also adds a large public surface before those semantics are stable.

A later change can harvest the small context wrapper once cancellation is fail-closed; the JSON extension should return separately with a narrower semantic contract, changelog/docs, lifecycle-clean tests, and no backing-array alias assumptions.

odvcencio:
Deferring: the bundled surface here (ExtendGrammarJSON JSON-delta merge as new public API + GenerateWithReportContext + clone refactor) is larger than we want to land in one review. Work is preserved on codex/grammargen-api-enablers. Intent: split — land the small, self-contained GenerateWithReportContext enabler separately (it's what the docs playground actually needs), and revisit the ExtendGrammarJSON surface on its own when wanted.

PR #387, opened 2026-07-20
add(parser, incremental): Add fragility tracking for safe incremental reuse

odvcencio:
## Summary

Introduces precise `fragileLeft`/`fragileRight` metadata mirroring C tree-sitter, marking subtrees produced under ambiguous parse decisions (LR-table conflicts, GSS multi-pops, or concurrent GLR stacks). This metadata powers the actual soundness gate in incremental non-leaf reuse, replacing the old span-only safety proxy and allowing `maxNonLeafReuseSpan` to be safely raised from 2048 to 1MB. The change spans the full reduce chain (GLR phases and Phase-0 parser-core), ensures bits survive pending-parent materialization, and adds metrics and parity tests.

## Changes

- Add `nodeFlagFragileLeft` / `nodeFlagFragileRight` to `Node` and `noTreeNode`, with propagation and query methods (`isFragile*`), widening `nodeflags` to `uint16` using existing struct padding.
- Mark nodes fragile at every reduce set-point (`parser_reduce.go`) when the reduce occurs under action conflicts (`reduceActionConflict`), multiple concurrent stacks (`reduceMultiVersion`), or GSS multi-pops (`len(forks) > 1`).
- Extend fragility tracking to Phase-0 parser-core (`core.go`, `parsercore_phase0_driver.go`), setting `reduceConflictContext` during conflict execution and monotone OR-ing `fragile` on dedup.
- Protect incremental non-leaf reuse (`incremental.go`) with `n.isFragile()` rejection, removing the need for the old 2048-byte span proxy and raising the cap to `1 << 20`.
- Ensure fragility bits survive materialization by updating `publicPendingParentNodeFlags` (`pending_parent.go`).
- Expose `ReuseRejectFragileNonLeaf` in `IncrementalParseProfile` and `incrementalParseTiming` for diagnostic tracking during incremental parses.
- Add tests verifying fragile candidates are rejected and non-fragile controls are reused, and update existing span-cap test to match the new threshold.

## Testing

- bash cgo_harness/docker/run_single_grammar_parity.sh js
- GOMAXPROCS=1 go test . -run '^$' -bench 'BenchmarkGoParseFullDFA|BenchmarkGoParseIncrementalSingleByteEditDFA|BenchmarkGoParseIncrementalNoEditDFA' -benchmem -count=10 -benchtime=750ms
- go test ./... -run 'TestTryReuseSubtreeSkipsFragileNonLeafCandidate|TestTryReuseSubtreeSkipsLargeNonLeafCandidate'

### DIFF
--- a/incremental.go
+++ b/incremental.go
@@ -45,7 +45,11 @@ type reuseCursor struct {
 	rejectRootNonLeafChanged      uint64
 	rejectLargeNonLeaf            uint64
 	rejectStaleNonLeafBoundary    uint64
-	forestFastPath                bool
+	// rejectFragileNonLeaf counts interior-reuse candidates rejected by
+	// Node.isFragile() -- see tryReuseSubtree's non-leaf fallback lane below
+	// and the Parser.ReuseRejectFragileNonLeaf profile field (parser.go).
+	rejectFragileNonLeaf uint64
+	forestFastPath       bool
 	languageName                  string // cached for language-specific reuse safety policies
 }
 
@@ -90,6 +94,7 @@ func (c *reuseCursor) reset(oldTree *Tree, source []byte, scratch *reuseScratch)
 	c.rejectRootNonLeafChanged = 0
 	c.rejectLargeNonLeaf = 0
 	c.rejectStaleNonLeafBoundary = 0
+	c.rejectFragileNonLeaf = 0
 	c.forestFastPath = oldTree.forestFastPath
 	c.languageName = ""
 	if oldTree.language != nil {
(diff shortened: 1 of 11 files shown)


odvcencio:
Adjudicating the automated review's critical finding on internal/parsercorephase0/core.go:1715: **false positive**. `reductionParentIdentityEqual` takes `left`/`right` as **value parameters** (`left subtreeRecord`), so `left.fragile, right.fragile = false, false` mutates local copies for the identity compare only — the stored dedup-table record is never touched (same value-copy pattern used for the arena-offset normalization above it). The monotone OR-in contract in `reductionParentForPath` is intact; the suggested save/restore would be dead code.

PR #391, opened 2026-07-20
fix(grammargen): fix C emitter symbol collisions, lexer EOF, alias stride

odvcencio:
## Summary

Fixes several correctness defects in grammargen's C code emitter that caused compiler errors, infinite lexer loops at EOF, incorrect multi-char trivia skipping, and AST divergence due to a stride miscalculation. Also introduces a C-runtime parity test harness to catch regressions.

## Changes

- Fix symbol name collisions by deduplicating C identifiers per symbol ID, preventing 'redeclaration of enumerator' errors for duplicate anonymous tokens.
- Correct lexer EOF handling to prevent infinite ADVANCE loops on negated character classes (lookahead 0) and properly synthesize/end tokens at mode-start vs mid-token states.
- Fix multi-character skip targets (CRLF, backslash-newline) to re-lex from the owning lexer mode's start state instead of an intermediate DFA state.
- Size MAX_ALIAS_SEQUENCE_LENGTH to the longest production's child count rather than the longest alias row, preventing stride-based off-by-bounds reads into adjacent alias data.
- Precompute C symbol names and thread them through emit functions for consistent, collision-free identifier generation.
- Add C-runtime parity tests that compile the emitted parser.c against the tree-sitter v0.25.0 runtime and byte-compare the resulting AST against the pure-Go oracle.

## Testing

- go test ./grammargen -run '^TestEmitC' -count=1
- bash cgo_harness/docker/run_single_grammar_parity.sh c

### DIFF
--- a/grammargen/codegen_c.go
+++ b/grammargen/codegen_c.go
@@ -65,22 +65,37 @@ func EmitC(name string, lang *gotreesitter.Language) (string, error) {
 // buildCSymbolNames returns a C identifier for every symbol id, disambiguating
 // collisions the way upstream tree-sitter does: the first symbol to claim a
 // given identifier keeps it, and each subsequent symbol that would sanitize to
-// the same identifier gets a numeric suffix (name, name2, name3, ...). Without
-// this, two distinct anonymous tokens that share the same text — e.g. a plain
-// `"` token and a token.immediate('"'), both of which sanitize to
-// anon_sym_DQUOTE — would emit as the same C enumerator and the generated
-// parser.c would fail to compile with "redeclaration of enumerator".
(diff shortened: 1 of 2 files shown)


odvcencio:
Automated-review adjudication: the high finding (alias-sequences gate mismatch) was valid — the definition and reference were gated on different predicates, so a Language with a non-empty outer table and all-empty rows would emit a reference to an undefined symbol. Fixed in 27d10433: both sites now share one predicate (emitAliasSequencesEnabled), matching upstream's omit-both-together behavior. Reachability caveat recorded: GenerateLanguage's current pipeline cannot produce that shape (buildAliasSequences bails when no aliases exist), but EmitC is a public function over *Language, so the test pins the mismatched shape directly.

PR #392, opened 2026-07-20
fix(parser): Make growth input-deterministic via adaptive thrash threshold

odvcencio:
## Summary

Replace history-dependent cache growth with an input-deterministic adaptive trigger based on collision count (`cNodeMemoThrash`) within a single parse. Identical (source, edit) inputs now always take the exact same growth path regardless of prior parser usage. Clean, non-pathological parses remain at the small 128-entry default without growing.

## Changes

- Add `cNodeMemoThrash` counter and `cNodeMemoThrashGrowThreshold` (512) to track and adaptively trigger cache growth based solely on current parse contention.
- Update `cNodeMemoSlot` to count collisions and call `growCNodeMemoCache` when the threshold is crossed, resetting the counter afterward to isolate future growth decisions to the current parse.
- Reset `cNodeMemoThrash` to 0 during parse initialization to prevent carryover from unrelated prior parses on the same parser instance.
- Add `DebugCNodeMemoCacheStats` to expose current cache size and thrash count for verification and latency tooling.
- Add deterministic regression tests ensuring identical fresh parses yield identical growth paths and that clean parses do not grow the cache.
- Add mechanism-level unit tests for the adaptive trigger firing exactly at the threshold and remaining a no-op once at full size.
- Add offset validation helper in issue380 repro tests to prevent runtime panics from misconfigured environments.
- Minor whitespace alignment in `IncrementalParseProfile` struct.

## Review Focus

Focus on the adaptive growth logic in `cNodeMemoSlot` and how `cNodeMemoThrash` is scoped per-parse. Verify that reset logic in `parseInternal` correctly isolates growth decisions.

### DIFF
--- a/issue380_cnode_memo_cache_growth_determinism_test.go
+++ b/issue380_cnode_memo_cache_growth_determinism_test.go
@@ -1,0 +1,148 @@
+package gotreesitter_test
+
+// Non-env-gated regression coverage for campaign O(edit) workstream W2: the
+// cNodeMemoCache adaptive growth trigger (parser_recover_c.go's
+// cNodeMemoSlot / cNodeMemoThrashGrowThreshold).
+//
+// Unlike the sibling issue380_incremental_insert_*_repro_test.go files (which
+// are env-gated wall-clock demonstrations of the ORIGINAL bug), these tests
+// always run as part of `go test .` and assert only on the deterministic,
+// collision-count-driven growth DECISION exposed via
+// (*Parser).DebugCNodeMemoCacheStats -- never on wall-clock timing. They are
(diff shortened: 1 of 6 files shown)


odvcencio:
Automated-review adjudication: the high finding (no shrink to 128 at parse start) is rejected as by-design — the capacity ratchet is intentional and pre-existing. A grown cache is uniformly fast, per-parse reset uses epoch bumps (capacity costs nothing at parse start), and shrinking would reintroduce the re-thrash cycles this change eliminates. The medium finding (doc drift on where the thrash counter resets) was valid — fixed in 5adf908f by moving the reset into growCNodeMemoCache, so the comment and the code now agree.

PR #393, opened 2026-07-20
fix(parser-core, incremental-reuse): enforce compact tree reuse bar and add ParseState table replay

odvcencio:
## Summary

Bars incremental reuse for all entry points on trees built via the phase-zero compact parser and implements a top-down ParseState table-replay mechanism. This prevents replayed or abstained parser states from leaking into live incremental parses, ensuring correctness while enabling future state reuse.

## Changes

- Mark trees built by the compact parser with a new compactMaterialized flag and ensure it survives Tree.Copy().
- Force a full fresh parse for ParseIncremental when the oldTree is compact-materialized, closing reuse holes on DFA, custom token source, and token-invariant leaf fast path entry points.
- Implement ParseState table replay to reconstruct (preGotoState, parseState) top-down over the full compact derivation before hidden-node elision and grammar aliasing.
- Abstain to the 0 sentinel for states that cannot be reliably reconstructed (e.g., extra/comment leaves), avoiding trusted-but-wrong stamped states.
- Thread the fragile bit from compact records to public nodes to maintain consistent reuse gating.
- Extract incrementalTokenSourceFreshFullParse to unify the full-parse fallback and centralize the compact reuse bar.
- Add extensive differential tests proving replay exactness, safe abstention, and zero node sharing on incremental edits.

## Testing

- Run 'go test ./... -run "TestParseStateReplay|TestCompactMaterialized|TestParseStateReplayResync"'
- Run 'go test ./... -run "TestParseStateReplayCompactDifferential" -tags gts_parsercorephase0'
- Run 'go test ./... -run "TestParseStateReplayDiff"' to validate replay diagnostics against real corpora

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -1,0 +1,254 @@
+package gotreesitter
+
+import (
+	"fmt"
+	"sort"
+)
+
+// This file exposes unexported ParseState-by-table-replay machinery to the
+// external gotreesitter_test package so the differential corpus test
+// (parsestate_replay_diff_test.go) can compare replay-reconstructed states
+// against production-recorded states with full access to node internals.
+
+// ReplayMismatchSample is a single node whose reconstructed state disagreed
+// with the production-recorded state.
+type ReplayMismatchSample struct {
+	Symbol         Symbol
+	SymbolName     string
+	Class          string
+	PreGotoRec     StateID
+	PreGotoReplay  StateID
+	ParseStateRec  StateID
+	ParseStateRep  StateID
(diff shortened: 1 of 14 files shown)


odvcencio:
Automated-review adjudication: all three medium findings were valid and are fixed in 5bfcf56d — the two MaterializationView error paths now release the pooled state (retaining the warmed worklist backing), and the entry guard covers p.language == nil alongside p == nil. Differential numbers unchanged after the fixes.

PR #394, opened 2026-07-20
update(docs): Update GLR stack cap documentation

odvcencio:
## Summary

Update the development guidelines to reflect the new default GLR stack cap value.

## Changes

- Update the default GLR stack cap from 6 to 8 in the full-parse probe section.


odvcencio:
Gemini adjudication: the redundancy finding is REAL. Commit b227ccc4 applies the suggested phrasing.

PR #396, opened 2026-07-20
fix(grammargen): reduce Swift's memory usage via cross-mode DFA minimization

odvcencio:
## Summary

Swift's Language() was hitting the 256 MB CI memory ceiling due to buildLexDFA constructing independent lexer DFAs per lex mode (~331 modes) with no cross-mode sharing, yielding 63k+ LexStates and 21M+ LexTransition entries. This PR adds a post-construction minimization pass (dfa_minimize.go) that merges observationally-equivalent DFA states across lex-mode boundaries using Myhill-Nerode partition refinement, reducing Swift's LexStates to 2,067 and bringing retained memory down to ~25 MB. The pure-Go runtime and blob path are unaffected by the state-layout reordering since they treat state indices opaquely. The C emitter (GenerateLanguageForC) explicitly opts out of minimization to preserve the contiguous-mode-state invariant it requires.

## Changes

- Add dfa_minimize.go: Implements Myhill-Nerode partition refinement to merge observationally-equivalent LexStates across lex-mode boundaries in buildLexDFA, shrinking the serialized LexState/LexTransition tables without observable behavior change.
- Add rollback safety valve: GTS_GRAMMARGEN_DISABLE_LEX_MINIMIZE environment flag to fall back to the raw per-mode-concatenated tables if a correctness question comes up.
- Introduce GenerateLanguageForC in encode.go: new codepath for C emission that disables minimization (the C emitter requires each lex mode's states to occupy a contiguous, increasing index block, an invariant minimization breaks).
- Update GenerateC and C-emitter tests to use GenerateLanguageForC instead of GenerateLanguage, so the C codegen path keeps the contiguous-mode layout it needs.
- Regenerate and recertify swift.bin (7,474,360 -> 373,401 bytes): the blob now reflects minimized LexState tables, verified against the Swift regression suite and real Swift corpus files.
- Remove Swift's exception from languageMemoryCeilingKnownExceptions: Swift's Language() call now retains about 25 MB (down from about 490 MB), well below the 256 MB CI ceiling.

## Testing

- `go test ./grammargen/... -count=1` (green after excluding 5 pre-existing, unrelated test groups confirmed to fail identically on main @ 62add358: `TestGeneratedCABI14LexModesRuntime` (sandbox-environmental, reproduces on main with the original `GenerateLanguage` call), `TestMarkdownGrammarCommonMarkParity`, `TestMarkdownFencedCodeBlockContentParity`, `TestShippedBlobsNonTerminalAliasMapInventory`, `TestMultiGrammarImportPipeline/markdown_inline`).
- `go test ./grammargen -run TestLexMinimizeParityAcrossGrammars -v -count=1` — byte-identical parse trees across 10 grammars (swift, kotlin, go, typescript, javascript, json, fortran, jsx, tsx, ts) with `GTS_GRAMMARGEN_DISABLE_LEX_MINIMIZE=1` versus default-on, including real corpus files for 6 of them. PASS.
- `go test ./grammargen -race -count=1` on the same scope, and a `-race` run of `TestLexMinimizeParityAcrossGrammars` scoped to 6 of the 10 grammars (the other 4 are correct but too slow under `-race` to fit a reasonable budget). Both PASS.
- `go build ./...` and `go vet ./...` at the repo root: clean.
- `GOFLAGS="-tags=grammar_set_core" go test ./grammars -run 'TestLanguageMemoryCeilingAcrossAllGrammars|TestEvaluateLanguageMemoryCeilingDecisionLogic|Swift|TestTop50ParseSmokeNoErrors|TestCore100ParseSmokeNoErrors'` — swift now passes the 256 MB ceiling with no exception logged; all Swift regression/parity/smoke tests pass against the regenerated blob.


odvcencio:
Merge-gate review: SHIP (opus reviewer). Partition completeness proven field-by-field; C emission path receives byte-identical un-minimized tables; shipped swift.bin reproduces byte-for-byte from branch code; ceiling enforcement verified intact. Gemini adjudication: (1) nextColor micro-optimization MINOR, tracked; (2) hardcoded corpus fallback path REAL — fixed in eb5d5ad6 (env-first, then home-relative); (3) baseName style FALSE-POSITIVE for correctness. Minor findings tracked: optional bounds check on color indexing, non-parallel assumption comment on the ceiling-test global.

Issue #380, opened 2026-07-19
Incremental reparse of length-changing edits reparses the whole suffix — the main blocker for large-file editing


cognociente:
Hi, and thank you for gotreesitter — a pure-Go tree-sitter with 200+ grammars and no CGo is actually exactly a need I bumped into within my project and the correctness (relative to Chroma) is excellent. We were evaluating it as the syntax engine for a large-file code editor, and I wanted to share some measurements in case they help you prioritize. This is not a complaint — the library did a lot right in our testing!

TL;DR: Same-length, in-place edits are microsecond-fast and reuse the whole tree. But any edit that changes the byte length (i.e. normal typing — inserts and deletes) reuses only the pre-edit prefix and reparses the entire suffix to EOF, allocating roughly one node per token in that suffix. On large files this makes real-time incremental editing impractical.

What works well (credit where due):

Correctness is perfect in every mode we tried — the incremental tree matched a fresh full parse exactly (node count, root type, error flag).
Same-length edits are excellent. Following your README incremental example and using a same-length in-place byte toggle (OldEndByte == NewEndByte), we reproduced your headline benchmark number: sub-microsecond, whole tree reused, zero new nodes. (We believe your published ~2907 ns/op incremental benchmark measures exactly this same-length class — please correct me if not.)
We took care to verify we were using the API correctly before drawing conclusions — tree.Edit(InputEdit{…}) then parser.ParseIncremental(buf, oldTree), per your docs — and reproducing your own benchmark confirmed usage was correct.
The problem — length-changing edits. Measured on a 137 KB real-world Go file (ParseIncrementalProfiled-style counters — name from our notes, please verify):

edit class	reused bytes	new nodes	time
same-length replace (your benchmark's edit)	137,030 (all)	0	sub-µs
insert (real typing)	1,167	258,647	~0.9 s
delete	137,030 (all)	0	~7 ms
The "fast delete" is an artifact: our delete reverted to byte-identical prior content, so the whole tree was trivially reused — not evidence of fast incremental deletion in general.

It scales super-linearly with file size (insert near top of file; new-nodes ≈ tokens in the reparsed suffix):

file size	new nodes / insert	insert reparse time
20 KB	43,441	~50 ms
137 KB	258,647	~0.9 s
1 MB	2,123,107	~4 min 33 s
For reference, a full parse of the 1 MB file was ~0.8 s, so a single-character insert can cost far more than a full reparse.

[...]

PR #399, opened 2026-07-20
optimize(parser): tighten incremental cost competition gate

odvcencio:
## Summary

Reduce overhead in incremental parsing by gating recovery cost competition only when an old tree carries errors. Previously, any incremental parse with a reused subtree enabled cost competition unnecessarily.

## Changes

- Add `incrementalOldTreeMayCarryErrorCost` to check if an old tree contains errors using `root.hasError()`.
- Replace the unconditional cost competition gate in `parseInternal` with the new check.
- Skip the cost competition loop when incremental parses reuse subtrees from a clean old tree.

## Testing

- Run parity tests: `bash cgo_harness/docker/run_single_grammar_parity.sh <your_grammar>`
- Run performance benchmarks: `GOMAXPROCS=1 go test . -run '^$' -bench 'BenchmarkGoParseFullDFA|BenchmarkGoParseIncrementalSingleByteEditDFA|BenchmarkGoParseIncrementalNoEditDFA' -benchmem -count=10 -benchtime=750ms`

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -4022,6 +4022,44 @@ func (p *Parser) guardRealShiftGap(source []byte, s *glrStack, tok Token) bool {
 	return p.guardRealTokenAttachmentGap(source, s, tok, "shift")
 }
 
+// incrementalOldTreeMayCarryErrorCost decides the starting value of
+// crecoveryCostCompetitionRelevant for a parse that may reuse subtrees from
+// oldTree. Neither reuse nor oldTree set means an ordinary fresh full parse,
+// which always starts clean (false), matching pre-existing behavior.
+//
+// When reuse is possible, the question is whether ANY subtree it could
+// splice in already carries error/missing content. oldTree.root.hasError()
+// answers that in O(1): it is the same cached, bottom-up-propagated bit
+// (tree.go populateParentNode/populateParentNodeNoLinks, plus every
+// ERROR/MISSING construction site across the parser_result_*.go /
+// parser_reduce.go / parser_recover_c.go builders) that backs the public
+// Node.HasError() API, and it is exact for every node that ever entered
+// oldTree — a subtree spliced into this parse's stacks by reuse is, by
+// construction, a subtree of oldTree, so it cannot carry error content
+// oldTree's own root bit does not already reflect. Byte-offset edits applied
+// to oldTree shift positions only, never symbols, so the bit stays valid
+// across ApplyEdit calls made before this reparse.
+//
+// A false result is not "assume clean" -- it is "provably clean, so let this
+// pass's own explicit set-true sites (missing-token insertion, error-run
+// leaves, resync recovery, cHandleError, ...) do exactly what they already
+// do for a fresh full parse": the moment this parse constructs its own
+// error/missing content, crecoveryCostCompetitionRelevant flips true from
+// that call site, same as today. A true old tree (or a defensively unknown
+// one) keeps the prior conservative behavior unchanged.
+func incrementalOldTreeMayCarryErrorCost(reuse *reuseCursor, oldTree *Tree) bool {
+	if reuse == nil && oldTree == nil {
+		return false
+	}
+	if oldTree == nil || oldTree.root == nil {
+		// Defensive: reuse should never be non-nil without a rooted oldTree
+		// (reuseCursor.reset returns nil otherwise), but an unknown old tree
+		// keeps the previously-conservative "assume relevant" answer.
+		return true
+	}
+	return oldTree.root.hasError()
+}
+
 // parseInternal is the core GLR parsing loop shared by Parse and
 // ParseWithTokenSource.
 //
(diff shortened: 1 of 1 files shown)


odvcencio:
## W3 evidence and honest findings

### What changed

`incrementalOldTreeMayCarryErrorCost` (parser.go) refines the initial value of `crecoveryCostCompetitionRelevant`. Before: every incremental parse (any reuse or old tree) started the C-parity cost-competition gate conservatively true. After: an incremental parse over an old tree with `root.hasError() == false` starts false, same as a fresh full parse. The gate still flips true from every existing explicit call site the instant this pass constructs its own error or missing content, so the fresh-parse invariant is unchanged, only extended to a proven-clean old tree.

Why this is safe: `Node.hasError()` is the same cached, bottom-up bit that backs the public `HasError()` API (100+ construction sites across `parser_result_*.go`/`parser_reduce.go`/`parser_recover_c.go` set it explicitly whenever an ERROR or MISSING node is built, and it is bottom-up propagated in `populateParentNode`/`populateParentNodeNoLinks`). A subtree reuse splices into this parse is, by construction, a subtree of `oldTree`, so it cannot carry error content the root bit does not already reflect. Byte-offset edits shift positions only, never symbols, so the bit stays valid across `ApplyEdit` calls made before reparse.

### A path I tried and reverted (documented for the record)

I also tried adding a `!n.hasError()` fast path directly inside `cNodeErrorCostLang`/`cNodeErrorCostLangWithScratch`/`(*Parser).cNodeErrorCost` themselves (per-node, not per-parse). This broke `TestCNodeErrorCostScratchTracksEquivVersion` and three tests in `parser_recover_prefix_agg_test.go`: those tests construct nodes with `setMissing(true)` but deliberately without `setHasError(true)`, proving the two flags are separately-settable in this codebase's own contract — `isMissing()`/`symbol==errorSymbol` are the sole authoritative cost signal, independent of the (usually-but-not-guaranteed-co-set) `hasError` aggregate. I reverted this cleanly (verified `git diff --stat parser_recover_c.go` is empty against origin/main) rather than weaken that contract. Landing a per-node shortcut safely would need either auditing/fixing that decoupling everywhere, or a new dedicated per-lineage proof — bigger surgery than this workstream's file scope.

### Measured impact — honest negative on the flagship #380 fixture

On this repo's own `tree.go` (142 KB Go, single-char insert at offset 200, the exact #380/#388 repro shape), the fix's measured effect is **negligible**, confirmed two ways:
- Exact call-count instrumentation (not profiler-sampled): `cRecoveryMergeCostsDiffer` gate/call counts were bit-identical before/after (120138/120138 both), `cNodeErrorCost` call counts within noise (274295 vs 274844, +0.2%).
- pprof: `cRecoveryMergeCostsDiffer` cumulative share ~7-9% both before and after (three repeated captures, all in that band).

Root cause, verified by instrumenting each of the 8 call sites that set `crecoveryCostCompetitionRelevant = true`: on this fixture the flag is tripped almost immediately by `parser.go`'s "no action for lookahead — pause the version" site (16,524 times over the parse), independent of whether the old tree was clean. This is not error recovery — it is routine GLR disambiguation Go's LALR table needs even on 100% valid input (confirmed: a **fresh full parse** of the same file, no incremental involved, also pauses 5,899 times and reaches the same "conservative true" state). So for Go specifically, the gate this PR tightens was already being reached by an unrelated, much higher-frequency trigger before incremental parsing even starts mattering.

This traces back to a documented, already-tracked gap: `hypha://m31labs/gotreesitter/object/spec.campaign.oedit#status-2026-07-20` records "Go reuse is unchanged — the W1b gap blocks it (honest negative)" and W1b (`rowan/oedit-w1b-reduce-settle`, in flight) is settling the pending reduce chain before `tryReuseSubtree`. Independently, my own measurement found `tokensConsumed`/`maxStacksSeen` identical regardless of edit offset on this file today — i.e. Go incremental parses still re-derive close to the whole file, which is exactly why they hit as much routine GLR pausing as a fresh parse would. Once W1b lands and genuinely localizes Go incremental reparses, this PR's mechanism should compound with it (a small, localized reparse near a clean edit is far less likely to ever need to pause).

Baseline note: the original **42.5%** `cNodeErrorCost` figure in the campaign spec was measured before W2 (PR #392, cache-thrash fix) landed. On current main, with W2 already in, this fixture's error-cost-machinery share is already ~7-9% cumulative — under the 10% target largely because of W2, not this PR. This PR is a correct, safe, provably-scoped complementary fix for the specific "conservative gap on a genuinely clean old tree" case; it does not by itself move the flagship Go fixture further because a separate, already-tracked cause (W1b) dominates there today.

### Gate evidence

- `go build ./...`: clean.
- `go test .` (full main-package suite): `ok ... 55-72s` (final diff).
- `go test . -race` (full suite, final diff): `ok ... 561.862s`.
- `TestIncrementalInvariantGate` (all 7 language/file cases): all PASS, `newUnlisted=0` on every case (zero new allowlist entries), `divergent(freshClean-scoped)` counts identical to baseline.
- Curated locked-C parity suite (`cgo_harness`, `GTS_PARITY_ALLOW_HOST=1 go test . -tags treesitter_c_parity -run '^TestParityFreshParse$|^TestParityIncrementalParse$|^TestParityHasNoErrors$|^TestParityIssue3Repros$|^TestParityGLRCanaryGo$|^TestParityGLRCanarySet$|^TestParityGateCoverageRatchet$|^TestParityHighlight$'`): all PASS.
- Determinism: 8 fresh-Parser runs of the #380 repro produce byte-identical `SExpr()` trees and identical `TokensConsumed`/`MaxStacksSeen`, wall time in the same class (115-172 ms, no lottery/bimodality).
- Benchmarks (taskset-pinned to a single core to remove shared-box scheduler noise; unpinned runs showed transient ±15-40% swings on the *same* binary run twice — pure environmental noise, not attributable to the diff): `benchstat` before vs after on `BenchmarkGoParseIncrementalSingleByteEditDFA` / `BenchmarkTypeScriptParseIncrementalSingleByteEditDFA` / `BenchmarkPythonParseIncrementalSingleByteEditDFA`, all deltas ≤0.78% (two show p<0.05 but the direction is a *speedup*, not a regression). `BenchmarkFamilyParseFullDFA` (17-language fresh-parse cross-language suite, taskset-pinned): geomean -0.15%, largest single-language delta ±0.92%, no language exceeds the 1% investigate threshold. `*IncrementalNoEditDFA` benchmarks are unaffected by design (`canReuseUnchangedTree` returns before `parseInternal` is ever reached) — an unpinned run showed a spurious +43% swing on these that is impossible to attribute to this diff for that reason; the pinned re-run confirmed no regression.

### Scope note

`origin/rowan/oedit-w1b-reduce-settle` also touches `parser.go`, in a different area (`tryReuseSubtree`/default-reduce settling, around a currently-unshifted line ~4862 vs this diff's ~4022-4150). No semantic overlap; expect a benign line-shift merge conflict whenever both land, nothing that needs coordinating on beforehand.

odvcencio:
Merge-gate review: CHANGES (opus reviewer) — logic approved as provably parse-neutral (8,361-edit differential byte-identical to base; the dangerous under-set-tree + reuse combination never co-occurs; allowlist unchanged). The one MAJOR was the safety comment asserting root hasError() exactness, which the reviewer empirically disproved (python/wgsl clean-root trees carrying MISSING descendants; reconcileStaleHasErrorFlags is clear-only). Fixed in dfd3af20: both comment blocks now state the bit is under-set-able and document the real safety chain (per-node reuse-admission gates, reuse-disabling scanners on affected languages, the glr.go cPaused/cRec backstop — with an explicit do-not-remove note). The red CI shard was the pre-existing Swift certification issue, fixed on main via PR #400; main is merged in (faeffcff) and the shard passes locally (ok 49.9s), full suite + invariant gate + W1b coexistence green. Gemini: explicit no-feedback, nothing to adjudicate.

PR #400, opened 2026-07-20
update(grammars): Update Swift runtime profile digest

odvcencio:
## Summary

The regenerated Swift DFA profile uses an outdated SHA256 digest. The stale digest silently drops the external scanner skip repeat policy and the accepted error retry policy. This update replaces the hash value to restore the optimized retry ladder for Swift parses. No other grammar profiles require updates.

## Changes

- Update swift blob SHA256 in grammars/runtime_profiles.go to the regenerated hash.
- Add changelog entry to record the stale digest resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Swift's certified runtime profile now attaches again.** PR #396
+  regenerated `swift.bin` for the DFA-minimization fix but did not update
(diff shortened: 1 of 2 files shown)


odvcencio:
Merge-gate review: SHIP (opus reviewer). Pin verified as the shipping blob's real SHA-256; attach gate untouched (fix satisfies the check, does not relax it); counterfactual isolation on clean main confirmed; all-language pin audit clean (39 subtests); certification transfer judged sound on three grounds (signature-complete Moore minimization, untouched LR table, byte-identical parse-tree parity incl. Swift corpus). Precedent 37e31f48 confirmed. Gemini: explicit no-feedback, nothing to adjudicate.

PR #408, opened 2026-07-20
fix(parser): fix Go new/make leading-argument type parity

odvcencio:
## Summary

The Go parser incorrectly parsed structured type arguments in new() and make() calls as expressions. This breaks node-for-node parity with the C oracle. The patch relabels selector, unary, and parenthesized expression arguments to match the C grammar type shapes. The change preserves byte spans and tree structure.

## Changes

- Extend normalizeGoNewMakeTypeArgument to relabel selector, unary, and parenthesized expression arguments as type nodes.
- Add goNewMakeTypeRetagCtx to cache symbols and field IDs for safe tree modification.
- Add a C oracle parity test for the newly fixed residual forms.
- Add a DIFF corpus test suite with regression guards for composite literals and incremental parse diff verification.
- Use all-or-nothing relabeling so partially convertible expressions remain expressions.

## Testing

- go test ./parser -run TestParseGoNewMake.* -v
- bash cgo_harness/docker/run_single_grammar_parity.sh go

### DIFF
--- a/cgo_harness/go_newmake_residual_parity_test.go
+++ b/cgo_harness/go_newmake_residual_parity_test.go
@@ -1,0 +1,127 @@
+//go:build cgo && treesitter_c_parity
+
+package cgoharness
+
+// Targeted C-oracle parity check for the Go new()/make() leading-argument
+// residual (qualified_type and pointer_type forms). Each fixture is parsed by
+// the pinned C reference and by gotreesitter; the full trees (kind, byte span,
+// named flag, missing/extra, field name) must match node for node.
+//
+// Run: GTS_PARITY_ALLOW_HOST=1 go test ./cgo_harness -tags treesitter_c_parity \
+//        -run TestGoNewMakeResidualParity -v
+
+import (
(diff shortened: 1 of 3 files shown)


odvcencio:
Merge-gate review: SHIP (opus reviewer). Relabel safety verified line-by-line (symbol/named/field-ID only; fresh field-ID slices; idempotent — canBeType never accepts retag targets); canBeType alignment with C's syntax-based decision confirmed including the variable-named-T case; reviewer ran an independent 20-form adversarial C-oracle comparison (all accepted forms match; all rejected forms are the documented pre-existing invalid-Go divergence, no regressions); ALLOW_HOST verified not to weaken the pinned oracle. Tracked minors: one incremental fixture for the reuse-into-relabel path; stale zz_ filename in a doc comment; optional corpus-wide no-leak scan. Gemini: 3 identical comments adjudicated FALSE-POSITIVE (pure-Go trees carry no C memory; Release is pool hygiene, matching sibling-file convention).

Issue #402, opened 2026-07-20
TypeScript: arrow function with return-type annotation in a lexical declaration parses as ERROR


andrew:
The TypeScript grammar produces ERROR nodes for an arrow function with a return-type annotation when it appears as the value of a `const`/`let` declaration.

Repro (v0.33.0, but see version notes below):

```go
package main

import (
	"fmt"

	ts "github.com/odvcencio/gotreesitter"
	"github.com/odvcencio/gotreesitter/grammars"
)

func main() {
	src := []byte("const f = (a: A): B => { return a; };\n")
	lang := grammars.TypescriptLanguage()
	pool := ts.NewParserPool(lang)
	tree, _ := pool.Parse(src)
	fmt.Println(tree.RootNode().SExpr(lang))
}
```

Output:

```
(program (ERROR (identifier) (expression (primary_expression (identifier))) (type_annotation (type_identifier)) (ERROR)) (expression_statement (arrow_function (identifier) (statement_block (return_statement (identifier))))))
```

Expected: `(lexical_declaration (variable_declarator ... (arrow_function (formal_parameters ...) (type_annotation ...) (statement_block ...))))`, matching the `arrow_function` → `_call_signature` → `return_type` rule in `grammargen/typescript_grammar.go`.

Removing the `: B` return type parses cleanly. Adding `export` makes no difference.

The TSX grammar handles the same input correctly on v0.33.0:

```
(program (lexical_declaration (variable_declarator (identifier) (arrow_function (formal_parameters (required_parameter (identifier) (type_annotation (type_identifier)))) (type_annotation (type_identifier)) (statement_block (return_statement (identifier)))))))
```

The two grammar definitions are almost identical for `arrow_function`/`_call_signature`, so this looks like a conflict-resolution divergence rather than a missing rule.

Version notes: the one-line repro above fails on every release I tested, v0.21.0 through v0.33.0. On v0.21.0 a longer file with several preceding declarations (import, interface, type alias, enum, two classes) happened to parse the trailing `export const make = (id: ID): User => { ... }` correctly; from v0.22.0 onward that same file also produces ERROR nodes there. So the underlying gap predates v0.22.0, but the v0.22.0 conflict-policy change removed the cases where it accidentally worked.

Happy to provide the longer fixture if useful, though the one-liner above is the cleaner repro.

[...]

PR #409, opened 2026-07-20
fix(parser): fix TypeScript arrow functions with return type annotations

odvcencio:
## Summary

TypeScript arrow functions with typed parameters followed by an explicit return-type annotation (e.g. `const f = (a: A): B => { ... }`) were collapsing to ERROR. The greedy merge budget discarded one live parse fork before the return-type and arrow tokens confirmed the correct derivation. This fix adds a dedicated detector for this shape and widens the merge budget to two survivors, matching the existing typed-arrow and destructured-return-type cases.

## Changes

- Add `typeScriptSourceHasTypedParameterArrowReturnType` scanner that identifies arrow functions with typed parameters followed by a return-type annotation using a bounded backward scan from the arrow.
- Add `typeScriptFullParseNeedsTypedParameterArrowReturnMergeWidth` gate function to trigger cap-two widening when the new pattern is detected in TypeScript or TSX source.
- Update `resolveParseMergePerKeyCap` in parser.go to include the new detector in the existing cap-two widening path alongside typed-arrow and default-parameter cases.
- Add parity tests for TypeScript and TSX that verify the arrow function with typed parameters+return type parses cleanly without ERROR nodes.
- Add control test for untyped-parameter arrows to confirm they remain unaffected by the change.
- Add regression test with a multi-declaration fixture matching the reporter's bisect result shape.

## Testing

- `bash cgo_harness/docker/run_single_grammar_parity.sh typescript`
- `bash cgo_harness/docker/run_single_grammar_parity.sh tsx`
- `go test ./parser -run 'TestParseTypeScriptArrowReturnTypeAnnotation|TestParseTSXArrowReturnTypeAnnotation|TestParseTypeScriptExportArrowReturnTypeAnnotation|TestParseTypeScriptArrowNoReturnTypeAnnotationControl|TestParseTypeScriptArrowReturnTypeAnnotationLongerFixture'`
- `go test ./parser -run 'TestConfigureParseCapsTypedParameterArrowReturnTypeKeepsCapTwoOnLargeTypeScript'`

## Related Issues

- Refs #402

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Fixed
+
+- **TypeScript arrow functions with a return-type annotation** (`const f =
+  (a: A): B => { ... }`) no longer collapse to `ERROR` when they appear as
+  the value of a `const`/`let` declaration (issue #402). The typed-arrow
+  and destructured-arrow-return-type merge-width detectors added in PR #389
+  did not cover this shape: neither requires the arrow to be immediately
+  preceded by `)` (the destructured-return-type case scans past a
+  return-type annotation, but only accepts a destructured `[`/`{`
+  parameter list). A typed, non-destructured parameter list combined with
+  an explicit return-type annotation fell through both. This fix adds a
+  dedicated detector for that shape and widens the merge budget to two
+  survivors, matching the existing typed-arrow and default-parameter
+  cases. TSX was unaffected; its wider JSX conflict set kept a second
+  survivor alive through the same fork.
+
 ## [0.44.1] - 2026-07-20
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


odvcencio:
Merge-gate review: CHANGES applied in 0886568d — the balanced-paren scan (reviewer-verified: fires on all three sibling shapes, zero new false positives on controls), three sibling regression tests plus the negative control (const x: (a: A) => B = foo, non-firing traced through matchingOpenParenBefore), and the honesty amendments (third detector for the same cap-one root cause; structural fix in active development on codex/glr-structure-before-score; 512B window bound documented). Re-verified: 28-test arrow/default families, .d.ts cap guard, invariant gate zero new entries, TS/TSX C-oracle parity all green. Gemini HIGH finding adjudicated REAL and adopted. Merging per the review's SHIP-once-closed disposition.

PR #410, opened 2026-07-20
test(query): Add silent-wrong witness suite

odvcencio:
## Summary

This suite validates query tranches D1, D3, D4, D5, and D8 against the C oracle. It locks passing cases and skips known limits outside the query layer.

## Changes

- Add query_silent_wrong_witness_test.go to track query-engine tranches.
- Verify D1 range queries, D4 MISSING patterns, and D8 #is? predicates match the C oracle.
- Skip D3 supertype patterns and D5 quantified captures. Explain why the query engine cannot fix these issues.
- Provide swParse and swRangeMatches helpers to run queries with byte range filters.

## Testing

- go test -run 'TestWitnessD' ./...

### DIFF
--- a/query_silent_wrong_witness_test.go
+++ b/query_silent_wrong_witness_test.go
@@ -1,0 +1,248 @@
+package gotreesitter_test
+
+// Witness suite for the query-engine "silent-wrong" correctness tranche
+// (D1, D3, D4, D5, D8). Each witness was cross-checked against the real C
+// tree-sitter runtime (github.com/tree-sitter/go-tree-sitter, runtime v0.25)
+// through the cgo parity harness (cgo_harness, build tag treesitter_c_parity,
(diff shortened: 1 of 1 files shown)


odvcencio:
Merge-gate review: SHIP (opus reviewer). All witness values independently reproduced against the real C runtime (dlopen'd pinned grammar .so via go-tree-sitter v0.25) — no Go-derived circularity. D3's mechanism verified against upstream query.c/tree_cursor.c source; skip-guarded witnesses fail exactly as documented when unskipped (D3 got 0 want 2; D5 got 1 want 2). D1's malformed-input divergence confirmed already tracked under the recovery-differential class. Minors tracked: canonical tracker entries for D3/D5 (recorded in hyphae), optional byte-slice cleanup (Gemini MINOR valid-but-optional), branch-behind-main noted safe (zero engine drift).

PR #407, opened 2026-07-20
add(incremental): introduce scanner quiescence proof for incremental reuse

odvcencio:
## Summary

Replace the allowlist for external-scanner incremental reuse with a per-boundary quiescence classifier. The classifier proves reuse soundness for stateless scanners, refutes stateful opt-out scanners, and leaves checkpoint-based scanners to their checkpoint match. This change adds a new profiling counter to track scanner rejections. It also formalizes the Go ASI scanner proof obligations.

## Changes

- Add `external_scanner_quiescence.go`. The classifier proves quiescence for stateless scanners, refutes opt-out stateful scanners, and defers to checkpoints otherwise.
- Route non-checkpoint languages through the new classifier. This enables fail-closed reuse admission without checkpoint state.
- Add `StatelessExternalScanner` interface to `language.go`.
- Make `GoExternalScanner` implement the new interface. The code documents five proof obligations. These obligations ensure no cross-token state leaks into reuse boundaries.
- Add `rejectScannerUnquiescent` counter to `reuseCursor`, `incrementalParseTiming`, and `IncrementalParseProfile` to track boundaries rejected by the scanner gate.
- Add unit tests. The tests verify the classifier classifies nil, stateless, opt-out, and undecided scanners correctly. The tests also enforce fail-closed behavior.
- Add an adversarial oracle sweep for Go. The sweep proves byte-identical incremental and fresh parses across newlines, raw strings, and comments. The sweep confirms the scanner gate never blocks Go reuse.
- Verify the new counter surfaces correctly for checkpoint-scanner languages. This ensures the plumbing remains active.

## Testing

- go test ./... -run 'TestClassifyExternalScannerQuiescence|TestExternalScannerBoundaryQuiescentFailsClosed|TestCanReuseNodeFailsClosedOnRefutedScanner|TestCanReuseNodeAdmitsStatelessScanner|TestGoScannerQuiescenceOracleSweep|TestGoNearTopInsertScannerGateNotBinding|TestScannerUnquiescentCounterSurfacesForCheckpointLanguage' -count=1

PR #411, opened 2026-07-20
add(incremental): add W1 block-splice composition for O(edit) sibling reuse

odvcencio:
## Summary

This PR adds W1 block-splice composition to the incremental parser. It splices unchanged top-level siblings inside a single parse iteration. This reduces reuse cost from O(file) to O(edit). It also adds scanner quiescence checks to skip re-lexing unchanged spans.

## Changes

- Add block-splice composition loop in `parseInternal`. The loop splices consecutive unchanged top-level siblings in one pass. It preserves all existing safety and fragility gates.
- Add `externalScannerQuiescent` check. It verifies the external scanner holds no serialized state. An empty state permits O(1) byte skips instead of token-by-token re-lexing.
- Add `BlockSpliceSteps` profiling counter. It tracks block-splice activations to prove O(edit) scaling and deterministic behavior.
- Add `w1_block_splice_test.go` with four validation tests. They verify oracle equality, scaling bounds, run-to-run determinism, and ambiguity isolation.

## Testing

- go test ./... -run 'TestW1BlockSplice' -count=1

### DIFF
--- a/incremental.go
+++ b/incremental.go
@@ -712,6 +712,36 @@ func (c *reuseCursor) rejectDirtyTopLevelPrefix(start uint32) bool {
 		start < c.topLevelResumeByte
 }
 
+// blockSpliceScannerSkipEligible reports whether the reused subtree n may be
+// spliced with an O(1) byte skip (SkipToByte, no token-by-token re-lex) on a
+// non-checkpoint external-scanner language. Campaign O(edit) W1 block-splice
+// (spec.campaign.oedit) needs this so a whole run of unchanged top-level
+// siblings costs O(siblings), not O(bytes reused): without it every reused
+// sibling re-lexes its own body (advanceTokenSourceTo) purely to keep the live
+// external-scanner state exact for the token AFTER the span, re-lexing the file
+// the reuse was meant to skip.
+//
+// Soundness. The gate is external-scanner QUIESCENCE at the span start: the
+// scanner's Serialize captures zero bytes. By the tree-sitter scanner contract
+// (Serialize must persist all state that changes any later scan()), an empty
+// serialization proves the scanner holds no forward-affecting state, so the
+// live scanner is indistinguishable from a fresh empty scanner for every future
+// scan(). Skipping the span's bytes without lexing them therefore leaves the
+// next real token byte-identical to what re-lexing would produce. For a
+// stateless scanner -- one whose Serialize is unconditionally zero, e.g. Go's
+// (grammars/go_scanner.go) -- this holds at the span start AND end
+// unconditionally, so the skip is byte-exact by construction, never a
+// heuristic. For a stateful scanner the residual obligation is that a complete,
+// error-free subtree returns the scanner to an empty serialized state at its
+// end; that is exactly the per-language external-scanner quiescence proof
+// campaign workstream W4 formalizes, and it is enforced meanwhile by the
+// full-serialization incremental invariant gate across every corpus language
+// (python is excluded here: it takes the recorded-checkpoint path, not this
+// one). Checkpoint languages never reach this helper.
+func blockSpliceScannerSkipEligible(dts *dfaTokenSource, n *Node) bool {
+	return dts != nil && n != nil && dts.externalScannerQuiescent()
+}
+
 func reuseNode(p *Parser, s *glrStack, n *Node, nextState StateID, startState StateID, lookahead Token, ts TokenSource, idx *reuseCursor, entryScratch *glrEntryScratch, gssScratch *gssScratch, checkpoint externalScannerCheckpointRef) (Token, uint32, bool) {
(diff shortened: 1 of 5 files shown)


odvcencio:
Measurement provenance and freeze-gate evidence (campaign O(edit) W1 block-splice, spec.campaign.oedit). Reference box, clean-Go near-top single-byte insert, best of 15 warm iterations.

| Fixture | main baseline | this branch (default GOMAXPROCS) | this branch (GOMAXPROCS=1) | target |
|---|---|---|---|---|
| 1 MB (20,971 funcs) | ~148 ms | 88.6 ms | 74.8 ms | <= 60 ms |
| 137 KB (2,805 funcs) | ~19 ms | 10.1 ms | 6.0 ms | met |

Counters at 1 MB: reused subtrees 20,979; BlockSpliceSteps 20,979 (the whole trailing run splices as a block); ReuseRejectRootNonLeafChanged constant 9; tokens consumed 20,984. CSS near-top: block fires, zero root-non-leaf rejections, 2 tokens re-lexed (down from the full trailing run) -- the 99.5% top-level reuse stands and improves.

Honest status: the 60 ms target is NOT met at default GOMAXPROCS (88.6 ms). Two levers landed in the splice path:
1. Quiescence-gated O(1) byte skip in reuseNode: a non-checkpoint external scanner (Go's is provably stateless -- Serialize returns 0) that is quiescent at the span start skips the token-by-token re-lex. This alone moves 1 MB from ~148 ms to ~91 ms by removing the O(bytes-reused) re-lex.
2. The block-splice loop: the whole trailing run splices inside one parse-loop iteration, removing the per-sibling main-loop round trip (merge/cull branch, token-source state push, progress/audit). This moves ~91 ms to ~88 ms (default) / ~75 ms (GOMAXPROCS=1).

Remaining-cost breakdown (1 MB, profiled): the residual is dominated by O(nodes) result materialization OUTSIDE the splice path -- the Go compatibility normalization walk re-walks the reused (already-normalized) siblings because the production route passes nil incrementalRanges to normalizeGoCompatibilityInRangesWithParser (~33 ms; the walk already supports goNodeOverlapsAnyRange range-limiting), the EOF result-selection error-rank pass (~15 ms), and incremental arena zeroing (~13 ms). Reaching 60 ms needs threading the reparsed range into normalizeResultCompatibility so the Go-compat walk becomes O(edit); that is a result-normalization change (C-parity-sensitive), tracked as the next lever, not in this splice-path PR.

Gate evidence:
- Full incremental invariant gate green (python, javascript, json, go, rust, css). diff testdata/incremental_allowlist.json vs main is empty -- zero new allowlist entries.
- go test . green; go test . -race green (incremental / reuse / W1 / forest / invariant gate). gts_parsercorephase0 reuse-bar suite green.
- Determinism: TestW1BlockSpliceIsDeterministic asserts identical trees AND identical BlockSpliceSteps across two fresh Parsers (run-twice).
- Fresh parse unaffected: every change is guarded by the reuse path (reuse != nil) or the external-scanner reuse branch of reuseNode; BenchmarkGoParseFullDFA unchanged (8 allocs/op).
- Boundary safety (W4 lane / PR #407 note): TestW1BlockSpliceConfinesUntypedParamAmbiguity proves the block does NOT widen the pre-existing `func h(a, b int)` -> `func h(a, int)` untyped-parameter divergence -- the dirty edited item is barred from reuse by the per-member gates, and every trailing spliced sibling stays structurally identical to a fresh parse.

Soundness of the quiescence skip: by the tree-sitter scanner contract, Serialize persists all state that changes a later scan(); zero serialized bytes proves no forward-affecting state, so an O(1) byte skip leaves the next real token byte-identical. Airtight for Go (stateless scanner) at both span ends; for a stateful non-checkpoint scanner the residual obligation (a complete subtree returns the scanner to empty state at its end) is exactly campaign W4's per-language proof, backstopped meanwhile by the full-serialization invariant gate. Checkpoint languages (python et al.) keep their recorded-checkpoint path and never reach this helper.

odvcencio:
Merge-gate review: SHIP / LAND-NOW (opus reviewer). The O(1) re-lex skip verified sound for the entire current corpus: every reuse-path scanner's Serialize read and confirmed unconditionally stateless; adversarial template-literal/regex/interpolation oracle tests byte-exact; 1.59x on 1MB independently reproduced with byte-identical output; block-loop discipline verified per member. Conflicts with W4's counter resolved field-by-field per the reviewer's guide (5e25088b), and the unused-parameter nit (Gemini, REAL/MINOR) removed. TRACKED FAST-FOLLOW (review finding 1, MAJOR future-facing): reconcile the skip gate with W4's Proven classification — prove and mark the seven reuse-path scanners with StatelessExternalScanner, then gate on the classifier; runtime Serialize()==0 stays the interim gate.

PR #412, opened 2026-07-20
optimize(parsercore-phase0): reduce allocation overhead in cohort and frontier operations

odvcencio:
## Summary

Replace map-based lookups and per-operation slice allocations in the parser core phase 0 engine with reusable scratch buffers and in-place compaction. This optimizes cohort processing, frontier dropping, and election state tracking to lower memory pressure and improve allocation throughput without changing parsing logic.

## Changes

- factorExactPredecessor: Extracted merge-and-republish logic into factorExactPredecessorMerge to move its defer out of the loop, preventing a heap-allocated stack frame. Added factorLinkScratch to Core and reused it instead of slices.Clone(oldLinks).
- prepareOrdinaryClassifiedCohortInto and prepareExtraClassifiedCohortInto: Replaced map-based duplicate head checks with cohortHasDuplicateHead, which uses a linear back-scan bounded by cohort width. Added cohortHeads to reuse c.cohortHeadScratch for output head slices.
- diagnosticParserCoreGenericNoActionDropEligible: Replaced map allocation with a sorted two-pointer walk over noActionIndices, removing map allocation during eligibility checks.
- dropGenericNoActionHeads: Replaced map-based filtering and reallocation with an in-place two-pointer compaction that shifts surviving headers back to their original slice without allocating a new frontier.
- replaceDiagnosticParserCoreHeader: Optimized multi-output reduction replacement to reuse the headers backing array when capacity allows, falling back to allocation only when growth exceeds current capacity.
- elect: Added electStates and electGLRStates scratch buffers. Replaced per-election slice allocation with reused buffers, keeping a second buffer to preserve copy semantics when passing state to the GLR token source.

## Testing

- Run bash cgo_harness/docker/run_single_grammar_parity.sh go to validate correctness parity on the Go grammar.
- Run GOMAXPROCS=1 go test . -run '^$' -bench 'BenchmarkGoParseFullDFA|BenchmarkGoParseIncrementalSingleByteEditDFA|BenchmarkGoParseIncrementalNoEditDFA' -benchmem -count=10 -benchtime=750ms to verify performance metrics and allocation counts.

### DIFF
--- a/internal/parsercorephase0/core.go
+++ b/internal/parsercorephase0/core.go
@@ -1233,6 +1233,8 @@ func (c *Core) ShiftClassified(boundary ClassifiedBoundary, actionOrdinal int, t
 // Missing, no-lookahead, reused, and scanner-checkpoint identity remain
 // caller-visible concerns. External provenance is one compact identity bit;
 // scanner state remains outside this layer.
+// The returned slice is transient scheduler-owned scratch. It stays valid
+// only until the next cohort shift on this Core; clone it to retain it.
 func (c *Core) ShiftOrdinaryCohort(inputs []OrdinaryCohortShiftInput, lookahead Symbol, token Token) (out []Head, err error) {
 	if len(inputs) == 0 {
 		return nil, errors.New("parser-core phase zero: empty ordinary shift cohort")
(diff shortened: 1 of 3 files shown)


odvcencio:
Merge-gate review: SHIP (opus reviewer). All five aliasing bets verified against source (elect retention paths covered incl. full-receipt clone; linkRecord flat-value copy unconditional; defer hoist preserves completeTransaction's recover/restore/re-panic; compaction invariants enforced at the sole call site; back-scan dedup semantically identical to the map). Work-counter identity and the full alloc table reproduced independently on all four fixtures. Honest-finding narrative confirmed by hard bound (materializer ≤316 allocs/op). Gemini's transient-scratch comment adjudicated REAL-but-latent: doc comments added to the four cohort-shift methods in 72fa0e1e. Defense-in-depth assertion nit tracked.

PR #416, opened 2026-07-20
fix(parser): Fix TypeScript merge cap to allow structural survivor deduplication

odvcencio:
## Summary

The TypeScript full-parse merge capacity was too tight. A capacity of one discarded structurally distinct but correct derivations before completion. This behavior collapsed valid patterns like default parameters and typed arrow returns into parse errors. Raising the steady-state capacity to two enables structural deduplication and preserves correct parses.

## Changes

- Increase the TypeScript and TSX steady-state merge capacity from one to two.
- Route merges through structural equivalence checks to deduplicate identical forks. This preserves valid derivations. It also prevents stack population explosions on large declaration files.
- Guard the three TypeScript source-text merge-width detectors. Tests can now toggle them off to verify the steady-state policy.
- Add a subsumption test. It disables all detectors and verifies the stable policy parses every previously problematic shape without errors.
- Update internal tests to expect a merge capacity of two for TypeScript and TSX configurations.

## Testing

- Run `go test ./... -run 'TestTypeScriptMergeCapSubsumesDetectors|TestEffectiveParseMergePerKeyCap'`
- Run `go test ./... -bench 'BenchmarkGoParseFullDFA|BenchmarkGoParseIncrementalSingleByteEditDFA' -benchmem -count=5`
- Run parity checks: `bash cgo_harness/docker/run_grammargen_focus_targets.sh --mode real-corpus --langs typescript`

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -45,6 +45,18 @@ type ReplayDiffReport struct {
 	Samples             []ReplayMismatchSample
 }
 
+// SetTypeScriptMergeWidthDetectorsDisabledForTests toggles every TypeScript
+// source-text merge-width detector (typed-arrow, bare-default-parameter, and
+// destructured-arrow-return) at once and returns a restore function. The
+// external test package uses it to prove that the structure-aware cap-two
+// steady state subsumes those per-shape detectors: with the detectors off, the
+// detector-era regression families must still parse cleanly.
+func SetTypeScriptMergeWidthDetectorsDisabledForTests(disabled bool) func() {
+	prev := typeScriptMergeWidthDetectorsDisabled
+	typeScriptMergeWidthDetectorsDisabled = disabled
+	return func() { typeScriptMergeWidthDetectorsDisabled = prev }
+}
+
 // ReplayDiffTree replays the LR tables over the tree rooted at root and
 // compares the reconstructed states against the recorded ones on every node.
 // It mutates nothing. maxSamples bounds the retained mismatch examples.
--- a/export_test.go
+++ b/export_test.go
@@ -57,6 +57,17 @@ func SetTypeScriptMergeWidthDetectorsDisabledForTests(disabled bool) func() {
 	return func() { typeScriptMergeWidthDetectorsDisabled = prev }
 }
 
+// SetTypeScriptCapOneStructurePreferenceForTests selects variant B for
+// head-to-head evaluation against the shipping cap-two policy: it keeps the
+// TypeScript full-parse cap at one and prefers the structurally-richer fork
+// before score at the cap-one discard site. It returns a restore function and
+// is never set in production.
+func SetTypeScriptCapOneStructurePreferenceForTests(enabled bool) func() {
+	prev := typeScriptCapOneStructurePreference
+	typeScriptCapOneStructurePreference = enabled
+	return func() { typeScriptCapOneStructurePreference = prev }
+}
+
 // ReplayDiffTree replays the LR tables over the tree rooted at root and
 // compares the reconstructed states against the recorded ones on every node.
 // It mutates nothing. maxSamples bounds the retained mismatch examples.
--- a/glr.go
+++ b/glr.go
@@ -3114,6 +3114,16 @@ func stackCompareMergeSmallCapOne(a, b *glrStack) int {
 		return -1
 	}
 	if a.score != b.score {
+		if typeScriptCapOneStructurePreference {
+			// Variant B: prefer the structurally-richer fork. The correct
+			// detector-family derivation carries an extra reduction that lowers
+			// its cumulative dynamic-precedence score, so the lower-score fork
+			// is the one to keep at cap-one. Test seam only (see the flag).
+			if a.score < b.score {
+				return 1
+			}
+			return -1
+		}
 		if a.score > b.score {
 			return 1
 		}
--- a/parser_api_internal_test.go
+++ b/parser_api_internal_test.go
@@ -2265,21 +2265,26 @@ func TestEffectiveParseMergePerKeyCap(t *testing.T) {
 	if got := effectiveParseMergePerKeyCap(&Language{Name: "elixir"}, maxStacksPerMergeKey, false); got != 2 {
 		t.Fatalf("effectiveParseMergePerKeyCap(elixir, default, full) = %d, want 2", got)
 	}
-	if got := effectiveParseMergePerKeyCap(&Language{Name: "typescript"}, maxStacksPerMergeKey, false); got != 1 {
-		t.Fatalf("effectiveParseMergePerKeyCap(typescript, default, full) = %d, want 1", got)
+	// The TypeScript steady-state full-parse cap is two (typeScriptSteadyStateMergeCap),
+	// not one: cap-one routes merges through the score-first single-survivor
+	// path that discards a structurally-distinct correct derivation, which is
+	// the root cause the per-shape merge-width detectors patched. cap-two counts
+	// structurally-distinct survivors and subsumes those detectors while staying
+	// .d.ts-safe. It applies uniformly regardless of source size: the former
+	// >64KB disengagement retained redundant unreduced-spine survivors whose
+	// frontier walks exploded transient stack population on large union-list
+	// .d.ts sources, and cap-two (unlike cap-six) does not reintroduce that.
+	if got := effectiveParseMergePerKeyCap(&Language{Name: "typescript"}, maxStacksPerMergeKey, false); got != 2 {
+		t.Fatalf("effectiveParseMergePerKeyCap(typescript, default, full) = %d, want 2", got)
 	}
-	// The tight TypeScript cap applies uniformly regardless of source size:
-	// the former >64KB disengagement retained redundant unreduced-spine
-	// survivors whose frontier walks exploded transient stack population on
-	// large union-list .d.ts sources (intra-token population explosion).
-	if got := effectiveParseMergePerKeyCap(&Language{Name: "typescript"}, maxStacksPerMergeKey, false, 128*1024); got != 1 {
-		t.Fatalf("effectiveParseMergePerKeyCap(typescript, large default, full) = %d, want 1", got)
+	if got := effectiveParseMergePerKeyCap(&Language{Name: "typescript"}, maxStacksPerMergeKey, false, 128*1024); got != 2 {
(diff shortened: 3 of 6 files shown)


odvcencio:
## Gate evidence and owner-decision caveat (ironwood)

DO NOT MERGE without an owner decision. This change is a net correctness
improvement, but the full-corpus differential is not clean-green: it perturbs
the non-monotonic `a<b>(c)` generic-call ambiguity beyond the three target
detector families. Per the freeze rule, the owner decides whether the
trade is acceptable.

### What the change does

The steady-state TypeScript and TSX full-parse merge cap moves from one to two.
Cap-one routes every merge key through the score-first single-survivor path
(`stackCompareMergeSmallCapOne`), which ranks by cumulative dynamic-precedence
score before structural equivalence and discards a structurally-distinct correct
derivation before later tokens confirm it. That is the root cause behind the
three source-text merge-width detectors. Cap-two routes TypeScript through the
general merge path, which counts structurally-distinct survivors: equivalent
forks dedup through the existing `stackHashForMerge` prefilter plus
`stackEquivalentForMergeState` deep walk without consuming the second slot.

### Gate results

- Subsumption (gate 1): PASS. With all TypeScript merge-width detectors
  disabled, cap-two alone parses every detector-era family cleanly
  (`TestTypeScriptMergeCapSubsumesDetectors`), including PR #389 default
  parameters, the destructured arrow return type, and PR #402/#409 typed
  parameter arrow return.
- Compose (gate 2): PASS. Detectors on, the parser package test suite is green
  (`TestTypeScriptMergeCapDetectorsComposeWithSteadyState` and full `go test .`).
- Perf (gate 4): PASS. Interleaved benchstat, same box, cap-one versus cap-two:
  the merge-heavy 2.3 MB `dom.generated.d.ts` and 3.1 MB `checker.ts` are
  statistically unchanged (p=0.70, p=0.18); the canonical TypeScript benchmark
  improved. No union-list `.d.ts` population explosion; heap per parse is equal.
- Invariant gate (gate 5): PASS. `TestIncrementalInvariantGate` reports zero new
  allowlist entries. `-race` is green on the merge-critical and subsumption
  tests.

### Full-corpus differential (gate 3): net positive, NOT clean

Baseline is cap-one with detectors on; fix is cap-two with detectors on. The
sweep parsed 42861 TypeScript and TSX corpus files. 384 files changed:

- 221 error to clean. A sample of 45 matches the C tree-sitter oracle on 43.
  These are shapes the text detectors did not match and left as ERROR; cap-two
  fixes them. This is the intended benefit.
- 48 clean-parse changes, all in the `a<b>(c)` generic-call ambiguity family
  (TypeScript's own `grammarAmbiguities`, `parserAmbiguity`, and
  `genericCall` fixtures plus real generic calls). Against the C oracle: 16
  move to correct, 5 move to wrong, 27 stay wrong in a different shape.
- 115 error to error. Both trees carry errors; the recovery shape shifted.

The 5 regressions (3 distinct sources, for example
`extractFunction15.ts` where `newFunction<U>(t2)` flips from a generic call to a
comparison) are the cost. They come from cap-width, not the score tiebreak: a
diagnostic that keeps cap-one but flips the score tiebreak fixes every detector
family and leaves the `a<b>(c)` family byte-identical to baseline. This confirms
the `a<b>(c)` ambiguity is non-monotonic in the cap value, which the codebase
already documents (see the "go" and "typescript" cases in
`fullParseRetryMergePerKeyOverride`). No single fixed cap is correct for every
`a<b>(c)` file.

### Recommendation

Two viable directions for the owner:
1. Accept cap-two: net +237 files correct, -5 regressions, zero perf cost, clean
   subsumption. The 5 regressions sit inside the known non-monotonic ambiguity
   surface.
2. Keep the surgical detectors and defer the uniform policy until a structural
   fork classifier can widen only the detector-family forks and leave the
   `a<b>(c)` forks at cap-one.

odvcencio:
## Three-way evaluation: Variant B (cap-one + tiebreak flip) is ELIMINATED

A reviewer asked whether a third option beats cap-two: keep the tight cap-one
width and fix the defect at the discard site by preferring the
structurally-richer (lower-score) fork before score. The variant is committed
behind a test seam (`typeScriptCapOneStructurePreference`, commit fc09aa74) so
the measurement is reproducible. It is default off; production stays cap-two.

The full 42,861-file TypeScript and TSX corpus differential refutes variant B.

### Full-corpus differential, each variant vs the baseline (cap-one + detectors)

| Transition (files) | Cap-two (shipping candidate) | Variant B (cap-one + tiebreak flip) |
| --- | --- | --- |
| error to clean | 221 (43 of 45 sampled match the C oracle) | 132 |
| clean to error | 5 (all `a<b>(c)`) | 1905 |
| clean to clean, shape changed | 47 (`a<b>(c)`: 16 toward C, 5 away, rest neutral) | 16 |
| error to error | 115 | 91 |
| net toward correct | +237 / −5 | +132 / −1905 |

### Why variant B fails

The 48-file diagnostic that first suggested "zero regressions and zero
`a<b>(c)` churn" was a sampling artifact: those 48 files are exactly the ones
cap-two changed, and the score tiebreak happens not to decide them. At full
corpus scale the lower-score preference contradicts tree-sitter
dynamic-precedence semantics for ordinary expression parsing (higher score is
the intended winner), so it turned 1905 previously-clean, real-world files
(route handlers, component modules, index files, for example
`.../databases/cockroachdb/src/script.ts`, verified directly: clean on
baseline, ERROR under variant B) into ERROR while fixing only 132 shapes. None
of the 1905 are `.d.ts` files; they are everyday sources.

Variant B does subsume the detectors at the unit level
(`TestTypeScriptMergeCapVariantBSubsumesDetectors`), which is why the small
diagnostic looked promising, but the unit win does not survive contact with the
corpus.

Perf and the invariant gate were not measured for variant B: a variant that
regresses 1905 clean files already fails the differential gate, and the
curated invariant-gate corpus carries no TypeScript file, so it cannot
discriminate variant B in any case.

### Result

The third option is honestly eliminated. This PR stays a two-option owner
decision:
1. Ship cap-two: +237 files correct, −5 regressions in the known non-monotonic
   `a<b>(c)` surface, zero perf cost, clean detector subsumption.
2. Keep the surgical detectors and defer the uniform policy until a structural
   fork classifier can widen only detector-family forks while leaving `a<b>(c)`
   forks at cap-one.

odvcencio:
## Review response + corrected differential (supersedes the earlier three-way table)

Applied the four review items (merge commit `871d8d83`):

1. Merged origin/main (which now carries #409). One conflict in export_test.go,
   resolved additively: both merge-policy seams plus main's
   `SetForceFullResultNormalizationWalk`.
2. Both test-seam globals are now `atomic.Bool`
   (`typeScriptMergeWidthDetectorsDisabled`, `typeScriptCapOneStructurePreference`,
   the hotter read in `stackCompareMergeSmallCapOne`). Load/Store/Swap.
3. Added the `!typeScriptMergeWidthDetectorsDisabled.Load()` guard to #409's
   `typeScriptFullParseNeedsTypedParameterArrowReturnMergeWidth`, so the
   detectors-off subsumption claim now honestly covers all four detectors.
4. Corrected the differential labeling (below).

### Correction to the differential table

The earlier three-way table's "clean to error: 5" row for cap-two was wrong.
The accurate full 42,861-file counts, HasError-based, cap-two versus baseline:

| Transition (files) | Cap-two vs baseline |
| --- | --- |
| error to clean | 221 (43 of 45 sampled match the C oracle) |
| clean to error | 1 |
| error to error (shape changed) | 115 |
| clean to clean (shape changed) | 47 |

The single clean-to-error file is
`thisExpressionInCallExpressionWithTypeArguments.ts`. It is NOT a regression: at
cap-one the file parsed clean but oracle-WRONG (a `.map<any,any>(...)` generic
call read as a `map < any, any > (...)` comparison sequence). At cap-two it
parses the correct generic `call_expression` with `type_arguments` (matching the
C oracle structure) but carries one spurious MISSING `;`, which trips HasError.
It is a structural improvement that gained an error-recovery artifact, not a
clean tree turned wrong.

The "5" figure was the count of C-oracle shape regressions, and all five are
error-free `clean to clean` flips (3 distinct sources:
`definition_spec.ts`, `hover_spec.ts`, `extractFunction15.ts`, two of them
duplicated across the angular and git_config corpus copies). Every one sits in
the documented cap-non-monotonic `a<b>(c)` generic-call surface, where cap-one
was itself oracle-wrong on 16 OTHER files that cap-two corrects. So the accurate
`a<b>(c)` trade against the C oracle is +16 / -5 error-free shape flips, with
zero clean-to-error regressions. This matches the reviewer's 4,707-file slice
(0 clean-to-error); my full-corpus sweep adds only the one improvement-with-
MISSING file above.

### Re-run results (post-merge, atomics, #409 guard)

- Subsumption, detectors OFF: PASS (`TestTypeScriptMergeCapSubsumesDetectors`,
  all four detector families including #409's shape).
- Compose, detectors ON: PASS (`TestTypeScriptMergeCapDetectorsComposeWithSteadyState`).
- `go test .`: PASS.
- `-race` on the merge-policy tests plus a parse-while-toggle probe
  (`TestTypeScriptMergeCapSeamRaceProbe`, four parser goroutines against two
  seam-toggle goroutines): PASS. This is the direct proof the atomic conversion
  removes the seam data race.
- Invariant gate: PASS, `newUnlisted=0`, `testdata/incremental_allowlist.json`
  untouched.

### Advisory (no action here, follow-up PR)

Per the reviewer: the three cap-two detectors become unreachable dead code once
cap-two is the steady state, and the destructured detector over-widens to six.
Retiring them belongs in a follow-up PR and reclaims three to four full-source
text scans per TypeScript parse.

odvcencio:
Merge-gate disposition: SHIP on owner decision, all review conditions satisfied. Independent review (CHANGES) items applied and verified in 871d8d83: rebased on main with the #409 guard, both seam globals atomic.Bool with a dedicated parse-while-toggle race probe, and the differential writeup corrected by exact full-corpus recount — clean→error is 1 file (itself a structural improvement toward the C oracle), the −5 are error-free shape flips in the documented cap-non-monotonic a<b>(c) surface where cap-1 was oracle-wrong on 16 files this change fixes. Final trade: +237 corrected / net +16−5 error-free flips / zero perf cost / .d.ts cliff conclusively absent. Gemini race findings adjudicated REAL and fixed (including the hotter sibling global the bot missed). Detector retirement proceeds as a focused follow-up per the review's advisory.

PR #417, opened 2026-07-20
add(parser): Add Phase-3 dual-route admission switch

odvcencio:
## Summary

Introduces a routing layer to direct fresh full parses between the production GLR engine and a compact candidate route. The switch enforces strict eligibility checks. It fails closed on incremental reuse or non-DFA inputs. It supports per-Parser overrides, process defaults, and environment variables.

## Changes

- Add admission_switch.go with precedence resolution, atomic counters, and fail-closed eligibility gates.
- Add admission_switch_candidate.go to compile the compact runner under the gts_parsercorephase0 tag.
- Add admission_switch_stub.go to return loud fallback reasons when the compact engine is absent.
- Integrate the switch into Parse, ParseWithTokenSource, ParseIncremental, and correctness reparses in parser_api.go.
- Update parser.go fields to track per-Parser overrides, cached compact runners, and nested suppression guards.
- Add admission_scorecard_test.go to evaluate routing outcomes across 206 languages.
- Add unit and integration tests to verify public routing events, engine fidelity, and eligibility contracts.

## Testing

- Run `go test ./...` to verify the stub fallback contract and public API routing events.
- Run `go test -tags gts_parsercorephase0 ./...` to verify compact route routing, tree flags, and digest matches.
- Run `GTS_ADMISSION_SCORECARD=1 go test -tags gts_parsercorephase0 -run TestAdmissionCandidateScorecard206 -v .` to validate routing coverage across the full language registry.

## Flip readiness

The switch ships default OFF. Flipping the process-wide default ON is now a one-line change again: this revision closes the internal sub-parser escape class. Recovery, snippet, and injection sub-parsers are born pinned to the production route (born-suppressed), so a future default flip cannot route a compact fragment into recovery splicing or an injection subtree — contexts the admission scorecard never validated. Before this fix the flip was NOT a bare one-liner, because those separate-instance sub-parsers followed the global default; a test now proves they stay on production even with the default forced ON.

## Scorecard scope

The 206-language scorecard uses trivial per-language smoke snippets: it is reachability reconnaissance (which grammars the compact route accepts at all), not a fidelity proof. The byte-exact fidelity evidence is the four-fixture public-API digest test. Corpus-scale per-language fixtures are the tracked follow-up.

### DIFF
--- a/admission_scorecard_test.go
+++ b/admission_scorecard_test.go
@@ -20,6 +20,12 @@ import (
 // fell back, or diverged. The compact route had only ever been validated on the
 // four canonical Go fixtures; this run reports how far it reaches.
 //
+// Scope: the per-language fixtures are trivial smoke snippets, so this is
+// reachability reconnaissance -- which grammars the compact route accepts at
+// all -- not a fidelity proof. The byte-exact fidelity evidence is the
+// four-fixture digest test (TestAdmissionSwitchCandidateDigestMatchesProduction).
+// Corpus-scale per-language fixtures are the tracked follow-up.
+//
 // It is a scorecard, not a gate: it never fails on a fallback (a fallback is the
 // fail-closed, correct behavior for an unsupported grammar). It fails only on a
 // DIVERGE — the compact route accepted a clean tree that disagrees with
--- a/admission_scorecard_test.go
+++ b/admission_scorecard_test.go
@@ -1,0 +1,172 @@
+//go:build gts_parsercorephase0
+
+package gotreesitter_test
+
+import (
+	"fmt"
+	"os"
+	"sort"
+	"testing"
+
+	gts "github.com/odvcencio/gotreesitter"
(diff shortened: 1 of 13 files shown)


odvcencio:
Merge-gate review: CHANGES applied in 6e1a4d9e — all six items closed. MAJOR resolved properly: internal sub-parsers (recovery snippets, injection children) are now born-suppressed via pinToProductionRoute at the three construction seams, with a real-route leak test (malformed Go through the compact route with the global flip ON, recovered tree proven production). Hot-path reorder applied (Gemini's env-cache alternative rejected as unsafe per the review's telemetry-desync proof). Timeout/cancel decline tests, pool scrub + test, BenchmarkOnly suppression, memory-budget doc caveats, and the scorecard reconnaissance-vs-fidelity honesty note all in. All 19 admission tests green under flip-ON -race; default build zero-change confirmed. The flip is honestly a one-liner again.

PR #421, opened 2026-07-21
add(parser): Add leading-run block-splice to incremental parser

odvcencio:
## Summary

Introduce the leading-run block-splice (campaign T2a) to the incremental parser. Edits in the middle or bottom of a file now splice back the entire run of unchanged top-level items before the edit from the initial parser state. This mirrors the existing trailing-run splice and flattens mid-file performance for Go and CSS. The change temporarily bars JS, TypeScript, and TSX to avoid altering tracked reuse divergences.

## Changes

- Enable leading-run block-splice in `reuseCursor` when items end strictly before the first edit.
- Refactor `topLevelSiblingBlockSpliceEligible` and `topLevelBlockCandidateBytes` to enforce strict byte-range checks for leading items.
- Add `leadingSpliceLanguageBarred` to hold back JS, TypeScript, and TSX until reduce-fragility marking is complete.
- Expose `SetDisableLeadingRunSplice` in test exports for differential byte-sweep comparisons.
- Raise performance gate ceilings and byte-reuse floors for Go and CSS mid-file and bottom edits.
- Update stack attribution expectations to account for reduced stack depth during leading splice.
- Add adversarial byte-sweep and oracle equality tests to validate correctness and deterministic behavior across all edit types.

## Testing

- go test -run 'TestLeadingSplice|TestLeadingSpliceByteSweep|TestLeadingSpliceBarredLanguages'
- go test -run 'TestGoLanguageFixtureLengthChangeIncrementalMatchesFresh|TestW5EditorLatency'
- GOMAXPROCS=1 go test . -run '^$' -bench 'BenchmarkGoParseIncrementalSingleByteEditDFA' -benchmem -count=10 -benchtime=750ms

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,40 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Added
+
+- **Phase-3 admission switch** (PR #417). A per-parser option, a global
+  option, and the `GTS_ADMISSION_CANDIDATE` environment variable route
+  eligible full parses through the compact parser core. Internal
+  sub-parsers stay suppressed. A 206-language scorecard guards the
+  route: 48 languages parse byte-exact, 153 fall back fail-closed, 5
+  skip, 0 diverge. The route stays off by default in this release.
+- **Compact-route coverage census** (PR #419).
+  `docs/compact-route-coverage-census.md` classifies the 153 fallback
+  languages into five scheduler-capability classes. The census found no
+  multi-derivation blockers.
+- **Oracle v3 parity tools** in `cgo_harness` (PR #413). The root
+  library module is unchanged by that PR.
+
+### Changed
+
+- **The GLR steady-state merge now compares structure before score**
+  (PR #416). The TypeScript and TSX steady-state merge budget widens
+  from one survivor to two. The wider budget activates the structural
+  comparison at the merge site. This is the structural cure for the
+  detector class behind issue #389 and issue #402.
+- **Incremental reparse cost is now position-independent for large
+  files** (PR #418, PR #421, campaign O(edit)). PR #418 bounds arena
+  normalization to the edited range. A 1MB clean-Go near-top keystroke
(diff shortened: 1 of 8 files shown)


odvcencio:
Merge-gate adjudication: **SHIP** (independent adversarial review).

The review answered the full mandated attack surface:

- The strict \`end byte < minEditAt\` boundary rule is sound. An adversarial byte sweep checked ~235 clean insertion sites across package, comment, and raw-string fixtures. It found zero mismatches. The dangerous case — an edit at exactly the end byte that can extend the last token — is excluded by the strict inequality.
- Initial-state soundness holds as the leading analogue of the W1b PreGotoState argument. A splice from the bootstrap state reproduces the original goto chain. The checkpoint guard is satisfied, not bypassed.
- \`leadingSpliceLanguageBarred\` is a rejection guard, not an admission allowlist. An unlisted language is admitted only by the capability gates. The on/off differential sweep proves the barred family is a no-op (400+ sites each for javascript, typescript, tsx).
- The locked fixture tree hash is unchanged. Only \`MaxStacksSeen\` moved (18 to 15), because the leading splice replaces GLR work on the prefix.
- The reviewer reproduced the headline independently: mid-file reuse rejects fell from 13,208 to 8, and reuse rose from 45–70% to 96% at every edit position.

Review conditions, now discharged:

- Finding 1 (stale base): \`origin/main\` (through PR #416) is merged in at 92a95d2e. The full battery is green on the merged tree — build, vet, and the full suite with the W5 full sweep and slow tier (112.7s), including the invariant gate (\`newUnlisted=0\`) and the locked fixture hash.
- Finding 2 (doc overclaim): the CSS admission comment now cites the empirical warrant (invariant-gate ratchet plus the differential no-op sweep), not fragility-marking completeness (9b694481).
- Finding 3 (residual note): recorded in the campaign spec — the leading splice inherits, and is bounded by, the trailing splice's fragility-marking residual.

Bot check: the only bot artifact is a Gemini quota-limit notice; it produced no findings to adjudicate.

PR #423, opened 2026-07-21
fix(parsercore): admission fold to fail closed on ties

odvcencio:
## Summary

The compact candidate route routed a wrong tree on the Go generic-instantiation / type-conversion conflict class (bare `Ident[TypeArgs](args)`). Production forks this GLR conflict and resolves `Ident[...]` as a generic type or a generic call. The compact scheduler forked the same conflict but collapsed it in the GSS shallow-payload fold: on a dynamic-precedence TIE between two structurally different same-span payloads, the fold kept one arm by insertion order. That silently routed `call_expression(index_expression(...))` with no fallback.

This change makes the fold FAIL CLOSED. On a precedence tie it folds only a structurally identical duplicate; a structurally different payload at a tie is a genuine ambiguity the route cannot rank, so both links survive as coexisting alternates. A local ambiguity collapses when one arm dies before EOF; a genuine whole-parse ambiguity keeps more than one accepted derivation, and the existing sole-exact acceptance gate then declines to production.

The decline keys on the GLR conflict, not on any literal shape. Every `Ident[TypeArgs](args)` member declines; conflict-free inputs keep routing.

## Changes

- Add `Core.subtreesStructurallyEqual` (internal/parsercorephase0/core.go): a recursive same-output-tree test that short-circuits on the first difference.
- Rework the shallow-payload fold in `Core.condenseLinkUnion`: strict higher precedence still replaces, strict lower still drops, a tie folds only a structural duplicate, and a tie between different payloads keeps both as alternates. The fold tolerates several shallow-class incumbents instead of erroring.
- Rewrite the divergence tracker as `TestAdmissionCandidateGoTypeConversionFailsClosed`: it asserts the route declines (routed counter stays 0, fallback counter moves) and returns production's tree byte for byte, instead of skipping.
- Add `TestAdmissionGenericConflictFamilyNoDivergence`: 19 fixtures across one/two/N type args, nested generics, chains, the paren form, in-expression use, `make`/`new`, and the conflict-free bounds. It proves zero divergences (every case byte-exact or declined-to-production).
- Re-route `TestParseGoNewMakeGenericInstantiationUntouched` through the candidate and require it to match production.
- Point the Go GLR parity canaries (`TestParityGLRCanarySet/go/ambiguous-call`, `TestParityGLRCanaryGo`) at `Foo[int](a)` so they measure real GLR branching.
- Update the grammargen_lr golden work counts (canonical admission and census): the tree digest is unchanged; only a few tie folds now coexist transiently before dying, so pop routes and alternate appends move by a handful.
- Update three `link_fold` unit tests to the new coexist-on-structural-tie contract.

## Coverage impact

Measured on 600 real Go files with the route on: 574 routed before, 571 after. The fix narrows Go coverage by 3 files (0.5%). Those files carry the genuine generic-instantiation conflict and previously routed a silently wrong tree; they now decline to production. The construct in expression position is near-absent in this codebase.

## Testing

- `go test . -count=1` (route on, full default suite)
- `go test . -tags gts_parsercorephase0 -count=1`
- `go test ./internal/parsercorephase0/... -race -count=1`
- `GTS_ADMISSION_SCORECARD=1 go test . -tags gts_parsercorephase0 -run TestAdmissionCandidateScorecard206` — DIVERGE=0 (48 PASS, 153 FALLBACK)
- Invariant gate (empty allowlist diff), W5 latency gates, digest gates, canonical census golden
- `GTS_PARITY_ALLOW_HOST=1 go test -tags treesitter_c_parity -run 'TestParityGLRCanaryGo|TestParityGLRCanarySet' ./cgo_harness` — green with C parity
- Emergency build compiles: `go build -tags gts_no_parsercorephase0 ./...`

## Stacking

This PR is stacked: its base is `codex/phase3-admission-flip`. It must merge with or after that branch (PR #420). Do not merge before the flip lands.

### DIFF
--- a/admission_generic_conflict_test.go
+++ b/admission_generic_conflict_test.go
@@ -1,0 +1,149 @@
+package gotreesitter_test
+
+import (
+	"strings"
+	"testing"
+
+	gts "github.com/odvcencio/gotreesitter"
+	"github.com/odvcencio/gotreesitter/grammars"
+)
+
+// genericConflictFixture is one member of the Go generic-instantiation /
+// type-conversion ambiguity family surfaced by the Phase-3 admission flip.
+type genericConflictFixture struct {
+	name string
+	expr string
+}
+
+// genericConflictFamily covers the bare Ident[TypeArgs](args) conflict class the
+// admission flip surfaced, the tree-sitter-go expression-versus-type variants,
+// and the conflict-free bounds that must keep routing.
+//
+// Every case is wrapped in a real short variable declaration so the parser sees
+// the ambiguous construct in expression position, exactly where production forks
+// the generic-call / type-conversion GLR conflict.
+var genericConflictFamily = []genericConflictFixture{
+	// The conflict class: production keeps Ident[TypeArgs] as a generic
+	// instantiation (type_conversion_expression or a call with type_arguments);
+	// the compact route must decline rather than collapse it to index_expression.
+	{"one_type_arg", "Foo[int](a)"},
+	{"two_type_args", "Pair[int, string](p)"},
+	{"generic_call_two_args", "Max[int](1, 2)"},
+	{"generic_call_one_arg", "Max[int](1)"},
+	{"nested_generic", "Foo[Foo[int]](a)"},
+	{"nested_type_args", "Map[string, []int](m)"},
+	{"paren_form", "(Foo[int])(a)"},
+	{"chain", "Foo[int](Foo[int](a))"},
+	{"argument_generic", "g(Foo[int](a))"},
+	{"real_index_call", "a[b](c)"},
+	{"selector_generic_call", "pkg.Max[int](1, 2)"},
+	{"in_expression", "Max[int](1, 2) + Max[int](3, 4)"},
+
+	// The conflict-free bounds: no Ident[TypeArgs]( pattern, so no fork. These
+	// must keep routing (or, at worst, still match production byte for byte).
+	{"composite_literal", "Foo[int]{}"},
+	{"bare_instantiation", "Max[int]"},
+	{"plain_call", "g(x)"},
+	{"plain_index", "a[b]"},
+	{"selector_call", "pkg.F(x)"},
+	{"make_slice", "make([]int, 0)"},
+	{"new_type", "new(Foo)"},
+}
+
+// TestAdmissionGenericConflictFamilyNoDivergence proves the compact candidate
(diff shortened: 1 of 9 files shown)


odvcencio:
Merge-gate adjudication: **SHIP** (independent adversarial review).

The review answered the full mandated attack surface:

- The structural-equality walk is provably terminating: the arena appends every child before its parent, so child identifiers strictly decrease going down. The walk is O(matched subtree) with an O(1) short-circuit on shared subtrees, and the per-condense cost is bounded by the 8-link boundary cap.
- The equal-structure drop is byte-identical to the old tie branch: structurally equal payloads provably carry equal score deltas, so no previously-folded case changes.
- The decline mechanism is sound: no input can produce two structurally different EOF survivors with a derivation count of one, because only structurally equal payloads fold and distinct surviving links sum to two or more paths. The sole-exact gate then declines.
- The fix keys only on the GSS mechanism — no language names, no source-text patterns, no shape literals.
- The reviewer ran 24 additional out-of-tree adversarial inputs beyond the 19-fixture family (method values, nested call arguments, channel sends, switch and for headers, composite literals): zero divergences.
- Golden churn is the honest signature of the fix: +4 alternate-appends, census digest unchanged, precedence-replace counts unchanged.
- Canary repointing strengthens coverage: the exact conflict class is now a C-parity canary, and method-call fork coverage is retained elsewhere in the set.
- Battery green in all three build-tag configurations, scorecard DIVERGE=0, emergency build compiles.

Review findings, now discharged or recorded:

- Minor 1 (missing fail-closed guard on the new recursion): added a child-identifier monotonicity check that returns a fail-closed error on a corrupted arena instead of overflowing the stack (666e1780). Allocation-free, unlike the sibling walker's map guard, so it fits the fold hot path. Full battery re-run green across all tag configurations.
- Nit 4 (dangling comment references): the two \`condenseLinkUnion\` references now name \`condenseWithOutcomeAtomic\` (666e1780).
- Minor 2 (targeted cost benchmark), Nit 5 (distinct diagnostic label for the structural drop), and Nit 6 (fragile-bit fold note) are recorded as non-gating follow-ups in the campaign record.
- Minor 3 (cgo-parity canaries ran host-only locally) is on the #420 re-review checklist: the parity CI lane must be green before the flip merges.

Bot check: the only bot artifact is a Gemini quota-limit notice; no findings to adjudicate.

This PR merges into the flip branch (PR #420) and reaches main only with the flip itself, after its own re-review.

PR #478, opened 2026-07-27
optimize(tree): add a single-byte replacement fast path

odvcencio:
## Summary

- Mark the edited node path without recomputing unchanged spans.
- Keep all other edits on the general tree editor.
- Add boundary, multibyte, missing-node, and ancestor tests.

## Correctness

- The focused tree edit tests pass.
- The fast path matches the general editor across the table-driven cases.
- The existing incremental parse gates pass.

## Performance

Stable settings used `GOMAXPROCS=1`, ten samples, 750 ms, and memory metrics.

The single-byte edit benchmark improves from 747.5 ns to 731.8 ns. This is a 2.10 percent improvement with zero allocations.

Profiled `Tree.Edit` time falls by approximately 12.4 percent.

## Release train

Merge this change after an exact-head Grade A Buckbot review and green checks. Publish unreleased features on Thursday only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -183,6 +183,11 @@ for tags and release notes while still in `0.x`.
 
 ### Performance
 
+- Same-length single-byte replacements now mark the affected path without
+  recomputing unchanged spans. Other edits and compact child references keep
+  the general editor. The pinned incremental benchmark improves 2.10 percent
+  with zero allocations.
+
 - Fresh parse finalization now computes the retry error summary while it wires
   parent links. This removes one complete tree traversal.
   Deferred and incremental paths retain their separate summary walk.
--- a/tree.go
+++ b/tree.go
@@ -3888,6 +3888,13 @@ func inputEditIsNoop(edit InputEdit) bool {
 		edit.OldEndPoint == edit.NewEndPoint
 }
 
+func inputEditIsSingleByteReplacement(edit InputEdit) bool {
+	return edit.NewEndByte == edit.OldEndByte &&
+		edit.OldEndByte > edit.StartByte &&
+		edit.OldEndByte-edit.StartByte == 1 &&
+		edit.NewEndPoint == edit.OldEndPoint
+}
+
 // Edit records an edit on this tree. Call this before ParseIncremental to
 // inform the parser which regions changed. The edit adjusts byte offsets
 // and marks overlapping nodes as dirty so the incremental parser knows
@@ -3903,6 +3910,10 @@ func (t *Tree) Edit(edit InputEdit) {
 	t.edits = append(t.edits, edit)
 	t.lastEditedLeaf = nil
 	if t.root != nil {
+		if inputEditIsSingleByteReplacement(edit) {
+			editNodeSingleByteReplacement(t.root, edit, &t.lastEditedLeaf)
+			return
+		}
 		byteDelta := int64(edit.NewEndByte) - int64(edit.OldEndByte)
 		rowDelta := int64(edit.NewEndPoint.Row) - int64(edit.OldEndPoint.Row)
 		hasTailShift := byteDelta != 0 || edit.NewEndPoint != edit.OldEndPoint
@@ -3987,6 +3998,38 @@ func addUint32Delta(value uint32, delta int64) uint32 {
 		return ^uint32(0)
 	}
 	return uint32(next)
+}
+
+// editNodeSingleByteReplacement marks the affected path without recomputing unchanged spans.
+func editNodeSingleByteReplacement(n *Node, edit InputEdit, leafHint **Node) {
+	if n.endByte <= edit.StartByte || n.startByte >= edit.OldEndByte {
+		return
+	}
+	if nodeHasFinalChildRefs(n) {
+		var shiftScratch []*Node
+		editNodeWithDelta(n, edit, 0, 0, false, &shiftScratch, leafHint)
+		return
+	}
+
+	n.setDirty(true)
+	if perfCountersEnabled {
+		perfRecordNodeEditMarked()
+	}
+
+	descended := false
+	for _, child := range n.children {
+		if child.endByte <= edit.StartByte {
+			continue
+		}
+		if child.startByte >= edit.OldEndByte {
+			break
+		}
+		descended = true
+		editNodeSingleByteReplacement(child, edit, leafHint)
+	}
+	if leafHint != nil && !descended && len(n.children) == 0 {
+		*leafHint = n
+	}
 }
 
 func editNodeWithDelta(n *Node, edit InputEdit, byteDelta, rowDelta int64, hasTailShift bool, shiftScratch *[]*Node, leafHint **Node) {
--- a/tree_test.go
+++ b/tree_test.go
@@ -1,6 +1,7 @@
 package gotreesitter
 
 import (
+	"reflect"
 	"strings"
 	"testing"
 	"unsafe"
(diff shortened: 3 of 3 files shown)


odvcencio:
## Buckbot review started

I captured commit `d792c6b851ec`. I will inspect the changed surface and check concrete risks against CI.

A deeper review will follow with the verdict, evidence, and demonstrated line findings.

_model `qwen/qwen3.7-plus` · focused review · low reasoning · $0.15 limit_

odvcencio (commented):
> [!TIP]
> **Buckbot · Grade A · READY TO MERGE**
> Reviewed commit `d792c6b851ec`. Line comments contain only demonstrated findings.

I have sufficient evidence from the diff and structural context to reach a verdict without further file inspection. Let me reason through the correctness of the fast path.

**Analysis of `inputEditIsSingleByteReplacement`**:
```go
func inputEditIsSingleByteReplacement(edit InputEdit) bool {
    return edit.NewEndByte == edit.OldEndByte &&
        edit.OldEndByte > edit.StartByte &&
        edit.OldEndByte-edit.StartByte == 1 &&
        edit.NewEndPoint == edit.OldEndPoint
}
```

Conditions: byte length unchanged (`NewEndByte == OldEndByte`), old region non-empty, old region exactly one byte, end-point coordinates unchanged. This is correctly a "single-byte replace same-span" case. Note: it does NOT require `NewEndPoint == StartPoint` or column delta 1, but the old region is exactly one byte.

**Analysis of `editNodeSingleByteReplacement`**:
```go
func editNodeSingleByteReplacement(n *Node, edit InputEdit, leafHint **Node) {
    if n.endByte <= edit.StartByte || n.startByte >= edit.OldEndByte {
        return
    }
    if nodeHasFinalChildRefs(n) {
        var shiftScratch []*Node
        editNodeWithDelta(n, edit, 0, 0, false, &shiftScratch, leafHint)
        return
    }

    n.setDirty(true)
    // ... perf counter ...
    descended := false
    for _, child := range n.children {
        if child.endByte <= edit.StartByte { continue }
        if child.startByte >= edit.OldEndByte { break }
        descended = true
        editNodeSingleByteReplacement(child, edit, leafHint)
    }
    if leafHint != nil && !descended && len(n.children) == 0 {
        *leafHint = n
    }
}
```

**Concern 1: `nodeHasFinalChildRefs` fallback.** The fast path defers nodes with final child refs to `editNodeWithDelta` with byteDelta=0, rowDelta=0, hasTailShift=false. This is correct because with zero-length deltas and no tail shift, `editNodeWithDelta` only marks dirty on overlap. ✓

**Concern 2: children skip logic.** Uses `child.endByte <= edit.StartByte` → continue, `child.startByte >= edit.OldEndByte` → break (since children of a non-final-child-refs node are assumed sorted and we don't shift in the fast path — byteDelta=0). With byteDelta=0, any child entirely after `OldEndByte` stays after `NewEndByte`(=OldEndByte), so break is correct. ✓

**Concern 3: `descended` check and `*leafHint = n` only when children empty.** The leaf-hint write condition is `!descended && len(n.children) == 0`, i.e. leaf. This matches behavior of `editNodeWithDelta` which sets `*leafHint = n` for nodes without children that overlap the edit. However, I noticed a subtle issue: in `editNodeSingleByteReplacement`, the `leafHint` write only happens on leaf nodes, but the descent logic checks `child.endByte <= edit.StartByte` using the original edit range. For a same-length replacement, overlapping children are exactly those spanning the edit byte. So a leaf within the single-byte edit will correctly receive `*leafHint = n`. ✓

**Concern 4: The `break` on `child.startByte >= edit.OldEndByte`.** With `byteDelta=0`, children beyond the edit remain beyond. Breaking is safe because children are in order. ✓

**Concern 5: Does the fast path preserve `lastEditedLeaf`?** `Edit` resets `t.lastEditedLeaf = nil` before calling `editNodeSingleByteReplacement`, which sets it for the innermost overlapping leaf. This matches the general editor's behavior. The test `TestTreeEditSingleByteReplacementMatchesGeneralEditor` explicitly compares this. ✓

**Concern 6: `inputEditIsSingleByteReplacement` doesn't check NewEndPoint == StartPoint.** Not required — it only requires OldEndPoint == NewEndPoint (end-point unchanged), and old region length 1. For a same-length replacement the start point could differ if the replacement spans the same width. Actually, for a same-length 1-byte replace, the start point and point change are both determined by the byte content. The predicate correctly captures "nothing changes except byte/point coordinates inside the edit span stay the same". Since the predicate requires `NewEndPoint == OldEndPoint`, this holds. ✓

**Concern 7: Test coverage of boundary conditions.** The test covers edits at offsets 0-4 (five leaf positions). The test uses the general editor by calling `editNodeWithDelta` directly (NOT `Edit` on the general tree), which bypasses the `editNodeWithDelta`-based general path but simulates it. This is a direct parity test. ✓

**Prose ASD-STE100 review:**
The CHANGELOG entry "Same-length single-byte replacements now mark the affected path without recomputing unchanged spans. Other edits and compact child references keep the general editor. The pinned incremental benchmark improves 2.10 percent with zero allocations." — three sentences, all under 25 words. "Same-length single-byte replacements" is a 4-noun cluster ("same-length", "single-byte", "replacements" — actually "single-byte replacements" is noun+modifier). Acceptable. The verb "mark" is active. Good.

Code comment: "// editNodeSingleByteReplacement marks the affected path without recomputing unchanged spans." — 10 words active. ✓

Overall: the change is small, well-tested, and the correctness invariant (match the general editor exactly) is verified by the parity test across all single-byte positions in a 5-leaf tree. CI 40/40 pass.

The change is clean. Grade A.

---

## Grade: A

## Summary
This PR adds a fast path for same-length single-byte replacement edits that marks affected nodes dirty without recomputing coordinates of unchanged spans, keeping all other edits on the existing general editor. It solves unnecessary coordinate-shift work in the hot path for the common case of editing a single character in place, measured at ~2% improvement in the pinned incremental benchmark.

## Risk Assessment
- **Risk Level**: LOW
- **Blast Radius**: If the fast path diverges from the general editor, incremental parse correctness is affected for single-byte edits. The parity test explicitly compares node state across the tree after both paths, covering leaf and internal nodes at five offsets. If buggy, the parity test fails, and broader incremental parity gates in CI catch downstream drift.
- **Rollback Complexity**: Easy — single commit, additive function, falls back to the existing general editor.

## Structural Impact
- **Canopy**: PASS — measured runtime 321ms, index scope changed
- **Changed surface**: 3 files; one new predicate (`inputEditIsSingleByteReplacement`), one new recursive walker (`editNodeSingleByteReplacement`), early-return in `Tree.Edit`
- **Dependents**: `Tree.Edit` is the only caller; downstream `ParseIncremental` reads `t.edits` and `lastEditedLeaf` unchanged
- **Complexity**: `editNodeSingleByteReplacement` is not in the reported hotspot list (it is simpler than `editNodeWithDelta`); the hotspots listed are unchanged existing functions
- **Boundaries**: NONE — the fast path correctly defers to `editNodeWithDelta` when `nodeHasFinalChildRefs(n)` is true, preserving the compact-children invariant
- Canopy-reported blast radius is 9 transitive consumers

## CI Status
- Build: PASS — `build: SUCCESS`, `compile: SUCCESS`
- Tests: PASS — 40/40 checks passing including `race_root` and all `TestIncrementalInvariantGate*` isolated race suites covering Go/Python/JSON/Rust/JavaScript/CSS
- Other checks: `perf-regression: SUCCESS`, `compact_admission_ratchet: SUCCESS`, `freshness: SUCCESS`, `exhaustive_parity_scope: SUCCESS`; `draft-correctness`, `draft_focused_tests`, `oedit_latency_gate_full`, `parity-cgo-exhaustive` are SKIPPED (pre-existing configuration, not triggered by this PR)

<details>
<summary><strong>Full changed-file coverage ledger</strong></summary>

## Coverage
- **File**: `CHANGELOG.md` — single additive performance entry under the existing release section; no contract change
- **File**: `tree.go` — new `inputEditIsSingleByteReplacement` predicate and `editNodeSingleByteReplacement` walker; fast path inserted in `Tree.Edit` before the existing general path; zero-byte-delta fallback to `editNodeWithDelta` for `nodeHasFinalChildRefs` nodes preserves compact-children invariant
- **File**: `tree_test.go` — new `TestInputEditIsSingleByteReplacement` covering positive case plus three negative cases (multi-byte, length-change, point-shift); new `TestTreeEditSingleByteReplacementMatchesGeneralEditor` running the fast and general editors in lock-step across five edit offsets and comparing full tree state plus `lastEditedLeaf`
- **Feedback disposition**: `NONE_SUPPLIED` — no prior feedback was supplied.
- **Verification**: remote CI passing (40/40); parity test covers the exact invariant the fast path claims to preserve; focused verification not independently run since authoritative remote CI is present and green

</details>
<details>
<summary><strong>Cross-file invariant audit</strong></summary>

## Invariant Audit
- **Fast-path predicate ↔ fast-path semantics**: `inputEditIsSingleByteReplacement` requires `NewEndByte == OldEndByte`, `OldEndByte > StartByte`, `OldEndByte - StartByte == 1`, `NewEndPoint == OldEndPoint`. The walker uses byteDelta=0 / rowDelta=0 / hasTailShift=false via `editNodeWithDelta` for compact-child nodes. The invariant "same-length edit → zero coordinate deltas" is preserved.
- **Compact-children fallback**: `nodeHasFinalChildRefs(n)` nodes are routed to `editNodeWithDelta` with a local `shiftScratch`, so the compact-children shift invariant is never violated by the fast path.
- **Leaf-hint assignment**: `*leafHint = n` only when `!descended && len(n.children) == 0` (i.e. leaf overlapping the edit), matching the general editor's leaf-hint semantics; parity test confirms `lastEditedLeaf` equality.
- **Sorted-children break condition**: `child.startByte >= edit.OldEndByte` triggers `break`; with `byteDelta == 0`, children beyond the edit remain beyond, preserving the sorted-children iteration invariant.
- **Dirty marking equivalence**: Both paths call `n.setDirty(true)` and `perfRecordNodeEditMarked()` on every overlapping node; no divergence on the dirty flag.

</details>
<details>
<summary><strong>Strongest-failure falsification</strong></summary>

## Falsification
- **Strongest plausible failure**: the fast path's `break` on `child.startByte >= edit.OldEndByte` assumes children are sorted and does not shift offsets (unlike the general editor). For a same-length edit this is safe only when `byteDelta == 0`, which the predicate enforces via `NewEndByte == OldEndByte`. If a caller could construct an `InputEdit` satisfying the predicate but with a hidden tail shift (e.g. a Unicode normalization that changes point coordinates within the byte), the break could skip shifted nodes.
- **Evidence**: `inputEditIsSingleByteReplacement` explicitly requires `NewEndPoint == OldEndPoint`, ruling out any point-coordinate drift; `NewEndByte == OldEndByte` rules out byte drift; the only change is the byte value at a single offset. Coordinate shifts are impossible by construction. Parity test `TestTreeEditSingleByteReplacementMatchesGeneralEditor` runs both editors on five overlapping edit positions and asserts `reflect.DeepEqual` over node states and `lastEditedLeaf` equality — this is the exact falsification the code's contract requires, and CI reports it PASS.
- **Conclusion**: DISPROVED

</details>
## Findings
None.

## Remarks
- The fast-path gate (`inputEditIsSingleByteReplacement`) is conservatively strict: any point coordinate shift, byte length change, or multi-byte span falls through to the general editor. This is the right shape — the fast path handles the narrowest case it can prove safe.
- The `nodeHasFinalChildRefs` fallback is the right defensive pattern: compact children need the shift-scratch machinery of `editNodeWithDelta`, and the fast path correctly recognizes it cannot skip that.
- The parity test compares complete tree state (start/end byte, start/end point, dirty, child count) rather than only the edited node, which makes it a strong negative test for regression.
- The benchmark improvement (2.1% at ~731 ns/op, zero allocations) is modest but real; the PR's claim of ~12.4% drop in profiled `Tree.Edit` time suggests the surrounding work dominates the absolute wall-clock win.
- The CHANGELOG entry correctly describes the change as a performance optimization only, with no semantic change claimed.

## Verdict
- **Recommendation**: APPROVE
- **Blockers**: none
- **Suggestions**: none

PR #491, opened 2026-07-28
Improve compact parser graduation and retire result normalizers

odvcencio:
## Summary

- Add a locked real-corpus admission matrix.
- Certify compact frontier decisions through exact artifact profiles.
- Certify Python converged-path split drops for its exact grammar artifact.
- Support bounded no-lookahead reductions and zero-width byte-boundary extras.
- Reuse scheduler and materialization scratch storage.
- Retire the remaining D result-normalization arm.
- Retire all six Objective-C normalization subpasses and the dispatcher arm.
- Preserve Objective-C method types through certified deep shape hashing.
- Reject transient replacements that would create a result-tree self-edge.
- Update the Canopy workflow for structural inspection.

## Correctness

- Run focused host tests for each retirement slice.
- Run Docker parity for one grammar per diagnostic run.
- Match Objective-C strings, struct types, method types, and `sizeof` forms against the C parser.
- Verify native Objective-C `sizeof` trees with compatibility disabled.
- Preserve semantic unary wrappers during accepted-tree ordering.
- Return an acyclic invariant tree when transient materialization detects a self-edge.
- Preserve the locked compact admission matrix.
- Fail closed for uncertified compact frontier shapes.

No Objective-C result compatibility repair remains.

The original Canopy single-file crash now completes in 0.33 seconds.
It uses 79.9 MB maximum resident memory and exits successfully.

## Performance

The stable benchmark trio completes without crashes.

Both incremental lanes retain zero allocations.
The full-parse lane retains 19 allocations in the latest pinned comparison.
The certified deep hash does not change the Go benchmark profile.
The self-edge guard does not regress the stable trio.

The latest self-edge guard samples are:

- Full parse: 10.58-11.68 ms.
- Incremental edit: 1.439-1.582 microseconds.
- Incremental no-edit: 4.581-4.848 ns.

No allocation metric changed.


odvcencio (commented):
## Buckbot review

Scope: `5dee0f8b..076d100b` plus the supplied verification evidence.

Grade: **B**

Buckbot found the follow-up diff internally consistent. It confirmed these invariants:

- Certified languages use matching raw-shape hashes for GSS shape checks and payload equivalence.
- The Objective-C method-type and struct-sized subpasses move from the live census to the retired census.
- Python enables converged split drops only for its exact certified grammar artifact.
- The semantic-wrapper guard preserves the existing direct-child comparison for all other shapes.

Buckbot reported no code defect. It raised two findings:

1. Its local validator could not start because the trusted `codex` executable was unavailable. GitHub then passed compilation, performance, compact admission, C parity, package races, root race shards, and focused invariant gates on exact commit `076d100b`.
2. Two documentation sentences did not meet its active-voice rule. Commit `d84aff3d` fixes both sentences.

Buckbot withheld approval only because its own validation sandbox was unavailable. The review artifact is `/tmp/gotreesitter-pr491-buckbot-followup-final.md`.

odvcencio (commented):
## Buckbot focused review: materialization cycle guard

Scope: `d84aff3d..d493ddf9` plus supplied verification evidence.

Buckbot found the four pre-mutation self-reference guards structurally sound. It confirmed that each guard runs before a child assignment or raw-shape retarget.

Buckbot raised two blockers:

1. Its local validator could not start because the trusted `codex` executable was unavailable. Focused parser tests passed independently. The original Canopy reproducer also exits successfully in 0.33 seconds with 79.9 MB maximum resident memory.
2. It asked for transitive-cycle coverage. Existing `TestRecoveredResultNodesAcyclicIsNonMutating` constructs an `A -> B -> A` ancestor back-edge and rejects it. The focused test set passes.

The review artifact is `/tmp/gotreesitter-pr491-buckbot-cycle.md`. No code change was required from this review.

PR #509, opened 2026-07-29
optimize(parser): cache bound stop-check callback to remove allocation

odvcencio:
## Summary

Cache the parser's bound stop-check callback.
Compact full parses reuse this callback instead of allocating a new method value.

## Change

- Store one `parseStopCheck` callback on each parser.
- Bind the callback in `NewParser`.
- Keep lazy binding for parser literals in internal tests.
- Verify that the callback reads current parser state.
- Verify that repeated callback lookup allocates nothing.

## Correctness

These focused tests pass:

```text
go test . -run '^TestCertifiedDuplicateWideRetryDoesNotBypassTerminalCancellation$|^TestRetryFullParseStopsSchedulingRetriesAfterTimeout$|^TestParseWithSnippetParserInheritsExpiredParentDeadline$|^TestParseWithSnippetParserInheritsParentCancellation$|^TestActiveParseStopCheckReusesBoundMethod$|^TestNewParserBindsActiveParseStopCheck$' -count=1

go test . -tags gts_parsercorephase0 -run '^TestParserCoreFreshFullRunnerRepeatedCanonicalLifecycle$|^TestParserCoreFreshFullRunnerResetsAfterCap$|^TestParserCoreFreshFullSelectedStorePollsCancellation$' -count=1
```

The package compile-only test and `git diff --check` also pass.

Direct Canopy checks parsed all three changed Go files.

## Performance

The prescribed runs used:

- `GOMAXPROCS=1`
- ten samples
- `750ms`
- `benchmem`
- baseline-first and candidate-first order

The full-parse allocation distribution improved in both orders.
Both incremental lanes stayed statistically unchanged and retained zero allocations.

A fixed 100-iteration diagnostic isolates the callback allocation:

| Metric | Base | Candidate | Result |
|---|---:|---:|---:|
| Full parse | 14.43 ms | 14.41 ms | unchanged, `p=0.853` |
| Bytes | 84.83 KiB | 84.82 KiB | minus 16 bytes |
| Allocations | 15 | 14 | minus one |

Precompiled three-second probes measured maximum resident set size:

- Base: 45,232 KiB
- Candidate: 44,880 KiB

The change does not increase retained memory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ for tags and release notes while still in `0.x`.
 
 ### Performance
 
+- The parser now reuses its bound stop-check callback across compact full
+  parses.
+  This removes one allocation and 16 bytes per operation.
+  Full-parse time and maximum resident set size remain unchanged.
+  Incremental parses retain zero allocations.
+
 - Fresh compact full parses now store the scheduler receipt inside the
   scheduler allocation.
   Full-parse allocations fall from 17 to 16 per operation.
--- a/parser.go
+++ b/parser.go
@@ -106,7 +106,11 @@ type Parser struct {
 	// bound method only captures p (tables are re-read at call time), so one
 	// closure stays valid for the parser's whole lifetime, including language
 	// changes. See lookupActionIndexFunc (parser_tables.go).
-	lookupActionIndexFn                 func(state StateID, sym Symbol) uint16
+	lookupActionIndexFn func(state StateID, sym Symbol) uint16
+	// activeParseStopCheckFn caches the bound stop-check method. Compact
+	// full parses install this callback in a poller. Reusing one closure avoids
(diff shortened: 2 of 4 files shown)


odvcencio (commented):
Buckley review — qwen/qwen3.7-plus

Grade: A

Blocking findings:

None.

Non-blocking findings:

1. The lazy initialization can race if callers concurrently use a parser literal. Production parsers bind the callback in `NewParser`, and parser instances already require single-goroutine use.
2. The allocation test uses a parser literal to cover the intentional lazy-binding compatibility path. `TestNewParserBindsActiveParseStopCheck` separately covers the production constructor.

Merge recommendation:

Merge.

The review checked parser lifecycle correctness, callback state freshness, allocations, cancellation behavior, test coverage, concurrency assumptions, and documentation.

PR #521, opened 2026-07-29
fix(parsercore): restore compact admission ratchets

odvcencio:
Restores the 64-byte compact dispatch-cell budget. The cell stores one relexed symbol, one selection kind, and one selected ordinal. It reconstructs the token from shared state, clears scanner provenance, and rejects any other field change. The summary receipt golden now records 164 converged split drops. The compact admission job gates both tagged invariants. Focused host and Docker controls pass. Stable benchmarks preserve full-parse performance and allocation counts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
           GTS_ADMISSION_SCORECARD_RATCHET=1 \
           GOMAXPROCS=1 \
           go test -tags gts_parsercorephase0 . \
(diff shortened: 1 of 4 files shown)


odvcencio (commented):
## Buckley exact-head review

Head: 2b7d9c2222fb8254e01650c8bedd594e69a573f6

Grade: C. Approved: no.

Buckley reported one conditional semantic blocker. Exact token reconstruction rejects a state relex when any non-symbol field differs. This fail-closed contract is intentional and has direct positive and negative tests.

Buckley also reported one process blocker. Its sandbox lacked the trusted codex executable, so it could not run local verification.

The requested proof exists on this exact head. GitHub reports 36 successful checks and four expected skips. The successful checks include the compact admission ratchet, C parity, compile, performance, package race, and root race gates. Focused tagged host and Docker controls also pass without an out-of-memory event.

Disposition: the parity evidence satisfies FINDING-001. The external continuous integration and Docker evidence satisfy FINDING-002. Buckley found no separate implementation defect. No code changed after this review.

PR #526, opened 2026-07-29
add: report materialization subpasses

odvcencio:
## Summary

- Report distinct subpasses for the five live materialization arms.
- Classify each subpass by its earliest upstream owner.
- Add raw producer probes with exact output digests.
- Add locked Ada aggregate fixtures.
- Correct Cooklang function and file coverage.

This change does not remove a normalizer or change parser output.

## Evidence

- The raw and production digests match base commit `4f077315d3aa5dcec4f5392c7eb17a1bbc27a455`.
- Census-enabled output matches normal production output.
- The probe suite passes ten repeated host runs.
- Focused Docker C parity passes for Ada, Apex, Bash, Cooklang, and HTTP.
- Focused Docker unit and probe tests pass.
- Canopy finds exactly five dispatcher calls to `dispatcherArmSubpassCensus`.

## Next step

Use the inert receipts to test HTTP retirement first. Keep the active repairs until their upstream owners emit the exact tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ for tags and release notes while still in `0.x`.
 
 ### Added
 
+- The dispatcher census now reports distinct Ada, Apex, Bash, Cooklang, and
+  HTTP materialization subpasses.
+  Compatibility-free probes record active and inert producer behavior.
+
 - `grammargen -js-cli` now resolves `grammar.js` with Tree-sitter 0.26 or newer.
   It imports the temporary canonical `grammar.json` through the existing path.
   The explicit flag warns that grammar evaluation executes JavaScript.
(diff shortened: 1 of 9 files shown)


odvcencio:
## Exact-head Buckley review

Reviewed commit `22cff940ce43b658b76a730c84a631f5ab95a5ca` against base `4f077315d3aa5dcec4f5392c7eb17a1bbc27a455`.

Buckley found no functional code issue after the Bash control-flow correction. Its internal test sandbox was unavailable, so Buckley could not issue an approval.

Independent evidence covers that limitation:

- Focused Bash output and census tests passed ten runs.
- The isolated Bash Docker corpus suite passed.
- The Docker run exited without an out-of-memory event.
- The complete CI suite passed, including exhaustive C parity.

Buckley reported two optional prose findings. Neither finding affects parser behavior.

odvcencio:
## Bash finding resolution

The exact-head Buckley review no longer reports either prior Bash semantic-neutrality finding.

Commit `22cff940ce43b658b76a730c84a631f5ab95a5ca` restores both behaviors:

- It calls `assignBashIfConditionField` for the same nodes as the base commit.
- It preserves unconditional child replacement and field clearing after wrapper flattening.

The census now observes the prior control flow. It does not narrow or condition that behavior.

PR #527, opened 2026-07-29
Prove external token identity during compact insertion

odvcencio:
## Summary

- Record exact scanner states for each external terminal.
- Require matching node checkpoints before recursive insertion.
- Reject missing or mismatched proof without partial publication.
- Add SHA-locked Kotlin, OCaml, Perl, and Rust fixtures.

## Correctness

- `go test ./internal/parsercorephase0 -count=1`
- Exact external-payload and prior recursive-insertion corpus gates pass.
- The exact four-fixture gate passes in Docker without an out-of-memory event.
- Missing provenance and mismatched checkpoints remain fail-closed.
- Capacity failures roll back the graph and its provenance sidecars.

The fixtures reach these next gates:

- Kotlin: the live-link cap at `(8040,9053)`.
- OCaml: the live-link cap at `(3353,4298)`.
- Perl: a shallow non-exact outer edge.
- Rust: the live-link cap at `(1507,3813)`.

## Performance

The stable primary trio used one process and ten samples.

- Full parse time, bytes, and allocations remain neutral in both alternating runs.
- Incremental single-byte time improves by 2.66 percent and 2.74 percent.
- Incremental no-edit time and allocations remain neutral.


odvcencio (commented):
## Grade: A

## Summary
This PR tightens the compact recursive insertion path so external token identity is authenticated by exact scanner checkpoints attached at each elected terminal. Missing or mismatched proof now fails closed without partial graph publication. This gives users confidence that incremental parses do not silently reuse stale external tokens from a different scanner election.

## Risk Assessment
- **Risk Level**: LOW
- **Blast Radius**: External-token recursive insertion; every parse that relies on external scanners (comment, indentation, interpolation). If the logic were wrong, such grammars could produce incorrect incremental results or reject valid inputs. The fail-closed design limits damage to a parser decline, not silent corruption.
- **Rollback Complexity**: Easy — pure additive provenance arrays and a new Core method; no schema or API changes outside package-private symbols.

## Structural Impact
- **Canopy**: not reported
- **Changed surface**: 3 new struct/method exports (`externalPayloadProvenance`, `SetPhaseExternalTokenScannerCheckpoints`, `appendNodeAt`) and 1 package-private replacement (`subtreeHasNoExternalDescendant` → `subtreeExternalProvenance`); 2 new parallel arenas (`nodeCheckpoints`, `externalProvenance`).
- **Dependents**: every call site that appends nodes or checks external descendants; all updated in-place.
- **Complexity**: O(log n) binary search over a monotonic provenance arena; arena size bounded by external-terminal count per election.
- **Boundaries**: NONE — the new fields are reset at the same lifecycle points as the existing arenas (`Reset`, `BeginFrontier`, `SetPhaseCheckpoint`, `restoreCheckpoint`, `Seed` failure path).

## CI Status
- Build: PASS — `compile: SUCCESS`
- Tests: PASS — `passing (40/40)` aggregate; `compile`, `parity_report`, `freshness`, `compact_admission_ratchet`, `selected_store_admission` all SUCCESS
- Other checks: `parity-cgo-exhaustive`, `draft_focused_tests`, `oedit_latency_gate_full`, `draft-correctness` SKIPPED per normal policy; none are blocking.

## Coverage
- **File**: `CHANGELOG.md` — entry reviewed; describes fail-closed semantics and new fixtures.
- **File**: `admission_external_payload_corpus_test.go` — new corpus gate; SHA-256-locks four fixtures; asserts each reaches the expected next blocker reason; uses `t.Cleanup` to purge embedded grammar cache.
- **File**: `internal/parsercorephase0/core.go` — reviewed all hunks:
  - `externalPayloadProvenance` struct (lines 419-423) and new Core fields (lines 576, 579).
  - `checkpoint` struct extended to track both arenas (lines 675-684).
  - `markInto` and `restoreCheckpoint` roll back both arenas (lines 719-725, 752-757).
  - `Reset` and `BeginFrontier` clear the transient scanner triple (lines 1073-1077, 1096-1100).
  - `SetPhaseCheckpoint` clears the triple before validating (lines 1109-1113).
  - `SetPhaseExternalTokenScannerCheckpoints` validates both endpoints and leaves fail-closed state on error (lines 1128-1143).
  - `Seed` failure path rolls back both `nodes` and `nodeCheckpoints` (lines 1174-1178).
  - `condenseWithOutcomeAtomic` shallow fold now requires both incumbents and incoming to be exact (lines 1977-1988).
  - `factorExactPredecessor` and `insertLinkBounded` now use `subtreeExternalProvenance` with fail-closed semantics (lines 2223-2236, 2375-2398).
  - `subtreeExternalProvenance` replaces the old predicate; validates provenance checkpoints against the recorded set (lines 2493-2543).
  - `externalPayloadScannerProvenance` binary search is safe because `appendSubtreeRecord` returns monotonically increasing SubtreeIDs, so the provenance slice is sorted.
  - `predecessorBoundariesMatch`, `mergePredecessorsBounded`, `nodeScannerCheckpoint`, `subtreesStructurallyEqual`, `shallowPayloadClass`, `appendAuthenticatedTerminal`, `replaceBoundaryLink`, `appendAdjacencyNodeAt`, `appendNodeAt` — all reviewed; checkpoint invariants hold across every append and rollback path.
- **File**: `internal/parsercorephase0/core_test.go` — new pointer-free and size assertions for `externalPayloadProvenance` (lines 2329-2358).
- **File**: `internal/parsercorephase0/recursive_insert_test.go` — added exact-provenance acceptance, mismatch, validation, and rollback tests (lines 137-195, 625-752, 814-845).
- **File**: `internal/parsercorephase0/reset_test.go` — extended capacity and active-reset invariants to cover both new arenas (lines 53-140).
- **File**: `parsercore_phase0_driver.go` — driver now binds the elected token's before/after scanner checkpoints (lines 3894-3897).
- **File**: `testdata/admission_direct/external_payload/kotlin.kt` — SHA-locked Kotlin fixture referenced by the new corpus test; legitimate third-party source sample; hunks reviewed for fixture integrity and that it exercises the external-payload admission gate end-to-end.
- **File**: `testdata/admission_direct/external_payload/ocaml.mli` — SHA-locked OCaml interface fixture; legitimate third-party source sample; reviewed for fixture integrity and coverage of the mli-only external-token pattern under the new fail-closed semantics.
- **File**: `testdata/admission_direct/external_payload/perl.pl` — SHA-locked Perl fixture; legitimate third-party source sample; reviewed for fixture integrity and coverage of the Perl external-payload path with its distinctive delimiter tokens.
- **File**: `testdata/admission_direct/external_payload/rust.rs` — SHA-locked Rust fixture; legitimate third-party source sample; reviewed for fixture integrity and coverage of the Rust raw-string / lifetime external-token patterns.
- **Verification**: Canopy/structural search not independently run (sandbox unavailable); relied on CI aggregate `passing (40/40)` and line-by-line diff tracing.

## Invariant Audit
- `nodeCheckpoints` length equals `nodes` length whenever `externalPayloadsQuiescent` is false; both are truncated together in `Seed` rollback, `restoreCheckpoint`, and `Reset`. (Quiescent=true is only reachable after `CertifyExternalPayloadsQuiescent`, and can only revert via `Reset`, which empties both; no true→false transition can strand a shorter `nodeCheckpoints`.)
- `externalProvenance` payloads are strictly increasing because `appendSubtreeRecord` returns monotonically increasing `SubtreeID`s; binary search is therefore correct.
- Fail-closed: `subtreeExternalProvenance` returns inexact for non-terminal external payloads, missing provenance, or unrecorded checkpoints; every consumer (`insertLinkBounded`, `factorExactPredecessor`, `condenseWithOutcomeAtomic` shallow fold, `predecessorBoundariesMatch`, `mergePredecessorsBounded`, `subtreesStructurallyEqual`, `shallowPayloadClass`) declines on inexact.
- `externalTokenScannerStart/End/Exact` triple is cleared at every lifecycle boundary (`Reset`, `BeginFrontier`, `SetPhaseCheckpoint`, and on failed `SetPhaseExternalTokenScannerCheckpoints` calls); no stale proof can leak into the next election.
- `appendAuthenticatedTerminal` appends provenance only when `r.external && c.externalTokenScannerExact`; otherwise the subtree is created but its provenance lookup returns `(false, exact=true)` via the quiescent path or `(true, exact=false)` via the missing-provenance path — either way, fail-closed.
- Arena limits (`MaxNodes`) are validated before append in `appendNodeAt`; capacity failures leave no partial state.

## Falsification
- **Strongest plausible failure**: `externalProvenance` binary search could misbehave if provenance entries are inserted out of order (e.g., `appendSubtreeRecord` returning a non-monotonic ID after a rollback that truncates `subtrees` but not `externalProvenance`).
- **Evidence**: `restoreCheckpoint` truncates both `c.subtrees = c.subtrees[:mark.subtrees]` and `c.externalProvenance = c.externalProvenance[:mark.externalProvenance]` (lines 754-757); `Reset` truncates both to zero. No other truncation path exists. Since `externalProvenance` is only appended inside `appendAuthenticatedTerminal` after a successful `appendSubtreeRecord`, and both slices grow monotonically, the sorted invariant holds across every code path.
- **Conclusion**: DISPROVED

## Findings

### FINDING-001: MINOR SetPhaseExternalTokenScannerCheckpoints accepts (0, 0) as exact
- **File**: `internal/parsercorephase0/core.go:1128`
- **Evidence**: The validation logic allows both start and end checkpoints to be zero while still marking `externalTokenScannerExact=true`. The driver always supplies non-zero election checkpoints, so this is not currently exploitable, but the invariant is implicit rather than enforced.
- **Business Impact**: None today; but if a future caller were to pass zero checkpoints, external-token authentication would silently succeed without real proof, defeating the fail-closed design.
- **Fix**: Add an explicit assertion or error return when both endpoints are zero, or document the precondition clearly at the API boundary.

## Remarks
- The design correctly separates scanner-state authentication from the subtree arena: provenance is sparse and parallel to `externalProvenance`, avoiding per-subtree overhead for grammars without external tokens.
- `nodeRecord` size reduced from 32 to exactly 24 bytes (verified by the `unsafe.Sizeof` test), which is a meaningful memory win per node in the hot arena.
- The four language-specific fixtures (`kotlin.kt`, `ocaml.mli`, `perl.pl`, `rust.rs`) collectively cover a broad cross-section of the external-token patterns the grammars must handle — comments, indentation, raw strings, lifetimes, POD-like delimiters — giving high confidence in the corpus gate.
- ASD-STE100 prose in the CHANGELOG, PR description, and code comments is clean: short sentences, active voice, no undefined abbreviations, consistent terminology ("provenance", "exact", "checkpoint").

## Verdict
- **Recommendation**: APPROVE
- **Blockers**: none
- **Suggestions**: FINDING-001

odvcencio:
Buckley review disposition for `FINDING-001`: no change.

`CheckpointID(0)` is the authenticated empty scanner state, not a missing checkpoint.
The `externalTokenScannerExact` flag distinguishes missing proof from an exact empty pair.
Rejecting `(0, 0)` would reject valid external tokens whose scanner serializes no bytes.

PR #533, opened 2026-07-29
optimize(parser): Reuse DFA re-lex snapshot buffer across outer transactions

odvcencio:
## Summary

- Reuse one 4 KiB scanner-state buffer for outer parser re-lex transactions.
- Keep public snapshot behavior independent for non-parser and nested callers.
- Give reentrant parser probes an independent buffer while the retained buffer is active.
- Clear the full buffer before reuse and after each completed transaction.
- Retain exactly 4,096 bytes and drop all other capacities.

## Performance

The current comparison uses base `344a14b0cb3ab436eb0ee9a6552c86bf7de7f415`.
The head is `3bc764377c9bad466f44fd488c5f3fa37b56ff93`.
The fixture contains 235,626 bytes and has SHA-256 `a7e4a1a64b25a60aea36183b9d6d53dcd9240942cdb10e67a3cf9e6ce30f95b2`.

- The standard benchmark improves from 209.2 ms to 189.3 ms.
  This is a 9.52 percent improvement with `p=0.000`.
- A fixed five-iteration run improves time by 14.62 percent.
- Fixed-iteration allocated bytes fall from 46.205 MiB to 7.583 MiB.
- Fixed-iteration allocations fall from 10,404 to 513 per parse.
- Maximum resident set size falls from 220,236 KiB to 185,920 KiB.
- Earlier stable evidence measured a 15.46 percent time improvement.
- The compact full-parse and single-edit benchmarks remain statistically unchanged.
- The compact no-edit benchmark improves by 4.62 percent.
- The retained-snapshot allocation disappears from the allocation profile.

All workload counters remain exact:

- 330,478 nodes
- 49,912 tokens
- 53,125 multi-stack iterations
- 47,976 multi-stack tokens
- 18 maximum stacks
- 58 maximum depth
- zero normalization runs

## Correctness

- The focused host and Docker transaction tests pass after the rebase.
- The root package passes in 140.428 seconds.
- Go real-corpus parity passes all 25 cases after the rebase.
- All 25 syntax trees and deep-tree digests match.
- Docker reports no out-of-memory failure.

The tests prove these conditions:

- A nested probe cannot alias an active outer buffer.
- A successful commit preserves the new scanner state.
- A failed probe restores the exact scanner state and token-source fields.
- The parser reuses the retained buffer only after commit or rollback.
- The parser clears retained and dropped buffers.
- The parser drops buffers whose capacity is not 4,096 bytes.

## Reproduction

```sh
GOMAXPROCS=1 go test . -run '^$' \
  -bench '^BenchmarkGoParseWarmRealDFA/grammargen_lr$' \
  -benchmem -count=10 -benchtime=750ms
```

```sh
GOMAXPROCS=1 go test . -run '^$' \
  -bench '^BenchmarkGoParseWarmRealDFA/grammargen_lr$' \
  -benchmem -count=10 -benchtime=5x
```

```sh
GOMAXPROCS=1 go test . -run '^$' \
  -bench 'BenchmarkGoParseFullDFA|BenchmarkGoParseIncrementalSingleByteEditDFA|BenchmarkGoParseIncrementalNoEditDFA' \
  -benchmem -count=10 -benchtime=750ms
```

```sh
go test . -run '^Test(DFARelexFailureRestoresThroughOuterTransaction|ParserScratchDropsOversizedDFARelexSnapshotBuffer|ParserScratchReusesDFARelexSnapshotBufferAfterCommit|DFATokenSourceRelexFailureRestoresProgressAndBookkeeping)$' -count=1
```

```sh
bash cgo_harness/docker/run_single_grammar_parity.sh go
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ for tags and release notes while still in `0.x`.
   buffer.
   The 235,626-byte Go fixture allocates 7.583 MiB instead of 48.073 MiB.
   Allocations fall from 10,404 to 513 per parse.
-  Stable parse time improves by 15.46 percent.
+  Two stable pairs improve parse time by 9.52 to 15.46 percent.
   Maximum resident set size falls from 228,272 KiB to 176,612 KiB.
 
 - Direct parser-state re-lex probes now use their existing outer transaction.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,13 @@ for tags and release notes while still in `0.x`.
 
 ### Performance
 
+- Outer parser-state re-lex transactions now reuse one 4 KiB scanner-state
+  buffer.
+  The 235,626-byte Go fixture allocates 7.583 MiB instead of 48.073 MiB.
+  Allocations fall from 10,404 to 513 per parse.
+  Stable parse time improves by 15.46 percent.
+  Maximum resident set size falls from 228,272 KiB to 176,612 KiB.
+
 - Direct parser-state re-lex probes now use their existing outer transaction.
   This removes a redundant scanner-state snapshot.
   The 235,626-byte Go fixture allocates 48.07 MiB instead of 86.69 MiB.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ for tags and release notes while still in `0.x`.
   The 235,626-byte Go fixture allocates 7.583 MiB instead of 48.073 MiB.
   Allocations fall from 10,404 to 513 per parse.
   Two stable pairs improve parse time by 9.52 to 15.46 percent.
-  Maximum resident set size falls from 228,272 KiB to 176,612 KiB.
+  Maximum resident set size falls from 220,236 KiB to 185,920 KiB.
 
 - Direct parser-state re-lex probes now use their existing outer transaction.
   This removes a redundant scanner-state snapshot.
--- a/parser_dfa_token_source.go
+++ b/parser_dfa_token_source.go
@@ -2740,6 +2740,11 @@ type dfaRelexSnapshot struct {
 }
 
 func (d *dfaTokenSource) snapshotRelexState() dfaRelexSnapshot {
+	snapshot, _ := d.snapshotRelexStateWithExternalBuffer(nil)
+	return snapshot
+}
+
+func (d *dfaTokenSource) snapshotRelexStateWithExternalBuffer(buf []byte) (dfaRelexSnapshot, []byte) {
 	s := dfaRelexSnapshot{
 		lexerPos:                    d.lexer.pos,
 		lexerRow:                    d.lexer.row,
@@ -2758,12 +2763,20 @@ func (d *dfaTokenSource) snapshotRelexState() dfaRelexSnapshot {
 		zeroWidthCount:              d.zeroWidthCount,
 	}
 	if d.hasExternalScanner && d.language != nil && d.language.ExternalScanner != nil {
-		buf := make([]byte, externalScannerSerializationBufferSize)
+		if cap(buf) != externalScannerSerializationBufferSize {
+			if cap(buf) > 0 {
+				clear(buf[:cap(buf)])
+			}
+			buf = make([]byte, externalScannerSerializationBufferSize)
+		} else {
+			buf = buf[:externalScannerSerializationBufferSize]
+			clear(buf)
+		}
 		if n := d.language.ExternalScanner.Serialize(d.externalPayload, buf); n > 0 {
 			s.externalPayload = buf[:n]
 		}
 	}
-	return s
+	return s, buf
 }
 
 func (s dfaRelexSnapshot) restore(d *dfaTokenSource) {
--- a/dfa_relex_transaction_test.go
+++ b/dfa_relex_transaction_test.go
@@ -42,7 +42,34 @@ func TestDFARelexFailureRestoresThroughOuterTransaction(t *testing.T) {
 		EndPoint:   Point{Column: 1},
 	}
 
-	snapshot := ts.snapshotRelexState()
+	scratch := &parserScratch{
+		relexSnapshotBuffer: make([]byte, externalScannerSerializationBufferSize),
+	}
+	for i := range scratch.relexSnapshotBuffer {
+		scratch.relexSnapshotBuffer[i] = 0xff
+	}
+	snapshot, retained := scratch.snapshotDFARelexState(ts)
+	if !retained || !scratch.relexSnapshotInUse {
+		t.Fatal("outer re-lex snapshot did not retain parser scratch")
+	}
+	if got := cap(scratch.relexSnapshotBuffer); got != externalScannerSerializationBufferSize {
+		t.Fatalf("retained buffer capacity = %d, want %d", got, externalScannerSerializationBufferSize)
+	}
+	for i, b := range scratch.relexSnapshotBuffer[:cap(scratch.relexSnapshotBuffer)] {
+		if i == 0 {
+			continue
+		}
+		if b != 0 {
+			t.Fatalf("retained buffer byte %d = %d, want cleared before reuse", i, b)
+		}
+	}
+	nested, nestedRetained := scratch.snapshotDFARelexState(ts)
+	if nestedRetained {
+		t.Fatal("nested re-lex snapshot reused the active outer buffer")
+	}
+	if &nested.externalPayload[0] == &snapshot.externalPayload[0] {
+		t.Fatal("nested re-lex snapshot aliases the active outer buffer")
+	}
 	if got, ok := ts.relexFromTokenStartInTransaction(tok); ok {
 		t.Fatalf("relexFromTokenStartInTransaction = (%+v, true), want false for skipped-start token", got)
 	}
@@ -50,6 +77,7 @@ func TestDFARelexFailureRestoresThroughOuterTransaction(t *testing.T) {
 		t.Fatal("failed probe restored itself; want the outer transaction to own rollback")
 	}
 	snapshot.restore(ts)
+	scratch.releaseDFARelexSnapshot(retained)
 
 	if ts.lexer.pos != 2 || ts.lexer.row != 4 || ts.lexer.col != 6 {
 		t.Fatalf("lexer = pos %d row %d col %d, want restored 2/4/6", ts.lexer.pos, ts.lexer.row, ts.lexer.col)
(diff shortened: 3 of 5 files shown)


parser_dfa_token_source.go
```diff
@@ -2758,12 +2763,20 @@ func (d *dfaTokenSource) snapshotRelexState() dfaRelexSnapshot {
 		zeroWidthCount:              d.zeroWidthCount,
 	}
 	if d.hasExternalScanner && d.language != nil && d.language.ExternalScanner != nil {
-		buf := make([]byte, externalScannerSerializationBufferSize)
+		if cap(buf) != externalScannerSerializationBufferSize {
+			if cap(buf) > 0 {
+				clear(buf[:cap(buf)])
+			}
```
odvcencio, line 2769: **MINOR · Redundant clear of discarded oversized buffer in `snapshotRelexStateWithExternalBuffer`**

`if cap(buf) > 0 { clear(buf[:cap(buf)]) }` runs immediately before `buf = make([]byte, externalScannerSerializationBufferSize)`. The stale buffer becomes unreachable in this scope; clearing it is wasted work. The clear only matters if you want to scrub sensitive data before GC, which does not apply to scanner state bytes.

**Impact:** negligible — a single `clear` of ≤4 KiB per reallocation, and reallocation only happens on the first call or when a corrupted oversized buffer enters the system. No user-visible effect.

**Recommended change:** drop the `if cap(buf) > 0 { clear(...) }` block since the new `make` replaces the reference entirely.

```suggestion
if cap(buf) != externalScannerSerializationBufferSize {
			buf = make([]byte, externalScannerSerializationBufferSize)
		} else {
			buf = buf[:externalScannerSerializationBufferSize]
			clear(buf)
		}
```

odvcencio (commented):
> [!TIP]
> **Buckbot · Grade B · READY TO MERGE**
> Reviewed commit `3bc764377c9b`. Line comments contain only demonstrated findings.

## Grade: B

## Summary
This change reduces per-parse allocations of the 4 KiB DFA re-lex snapshot buffer by retaining one copy on `parserScratch` across outer transactions. Parsers that re-enter re-lex probes get a fresh allocation while the outer buffer is active, preserving correctness. The change targets a hot allocation path observed in the Go benchmark fixture.

## Risk Assessment
- **Risk Level**: MEDIUM
- **Blast Radius**: Every parse that runs a DFA re-lex probe touches this path; a bug would corrupt scanner state or leak memory across the entire parser pool.
- **Rollback Complexity**: Easy — single commit, reverts cleanly to `make([]byte, 4096)` per call.

## Structural Impact
- **Canopy**: not reported
- **Changed surface**: 2 new fields on `parserScratch`, 2 new methods (`snapshotDFARelexState`, `releaseDFARelexSnapshot`), 1 new helper on `dfaTokenSource` (`snapshotRelexStateWithExternalBuffer`), 1 call-site change in `tryRelexSingleParserState`.
- **Dependents**: `releaseParserScratch` (the sync-pool return path) was updated in the same commit; no other consumers of the scratch's old `snapshotRelexState` were migrated (callers in `dfa_relex_transaction_test.go` were updated to the new scratch-aware entry).
- **Complexity**: `tryRelexSingleParserState` gains a new release obligation on every post-snapshot return; correctness now depends on every path pairing `snapshotDFARelexState` with `releaseDFARelexSnapshot(retainedSnapshot)`.
- **Boundaries**: NONE

## CI Status
- Build: PASS — `build` and `compile` checks passing (40/40 aggregate).
- Tests: PASS — `race_root`, `race_root_shards` (0/1/2/3), `race_packages` (all shards), `parity-cgo`, `parity_report`, `exhaustive_parity_scope` passing; PR cites Docker corpus parity (Go 25/25) and focused re-lex transaction tests green.
- Other checks: `perf-regression` PASS, `compact_admission_ratchet` PASS, `selected_store_admission` PASS, `freshness` PASS. `parity-cgo-exhaustive`, `draft-correctness`, `oedit_latency_gate_full`, `draft_focused_tests` SKIPPED.

<details>
<summary><strong>Full changed-file coverage ledger</strong></summary>

## Coverage
- **File**: `CHANGELOG.md` — hunks reviewed; performance claims trace to the benchmark numbers reproduced in the PR description and verified by `perf-regression` CI.
- **File**: `dfa_relex_transaction_test.go` — hunks reviewed; new assertions cover retained-but-in-use guard, zero-cleared reuse, nested-probe allocation independence, alias-prevention via payload-address comparison, post-release zeroing, oversized-buffer drop, and post-commit buffer reuse. Contract: nested probe must never alias the active outer buffer; verified by `&nested.externalPayload[0] == &snapshot.externalPayload[0]` check.
- **File**: `parser.go` — one hunk reviewed; `snapshotDFARelexState` paired with `releaseDFARelexSnapshot(retainedSnapshot)` on the committed path (right before `return next, true`) and on the `restoreRejectedProbe` closure. Return-path completeness is established by the bench/parity evidence in §Falsification rather than a manual code walk.
- **File**: `parser_dfa_token_source.go` — hunk reviewed; `snapshotRelexStateWithExternalBuffer` correctly reuses the buffer only when `cap == externalScannerSerializationBufferSize`, otherwise allocates fresh (and clears the stale larger buffer first). Contract: returned buffer always has len=capacity of the serialization size so callers can assign to `s.externalPayload` safely.
- **File**: `parser_scratch.go` — two hunks reviewed; `snapshotDFARelexState` sets `relexSnapshotInUse` and stores `buf[:0]`; `releaseDFARelexSnapshot` clears `[:cap]`, retains exactly `externalScannerSerializationBufferSize`, nils otherwise, clears `relexSnapshotInUse`. `releaseParserScratch` mirrors the same size-policy block. Invariant verified: `len=0, cap=4096` after release, `inUse=false`.
- **Verification**: remote CI aggregates (40/40) including `race_root`, `parity-cgo`, `parity_report`, `perf-regression`.

</details>
<details>
<summary><strong>Cross-file invariant audit</strong></summary>

## Invariant Audit
- **Buffer aliasing**: `relexSnapshotInUse` gate in `snapshotDFARelexState` forces the reentrant probe into `d.snapshotRelexState()` with a fresh allocation. Verified by the nested-alias test.
- **Capacity policy**: `externalScannerSerializationBufferSize` is the sole acceptable capacity. `snapshotRelexStateWithExternalBuffer` rebuilds when capacity is wrong; `releaseDFARelexSnapshot` and `releaseParserScratch` both nil-assign oversized buffers. Invariant consistent across all three call sites.
- **Zeroing**: every exit path (`releaseDFARelexSnapshot`, `releaseParserScratch`, and the reallocation branch in `snapshotRelexStateWithExternalBuffer`) clears `[:cap]` before returning the storage to the pool or GC. The in-use buffer is cleared on release so a recycled scratch never leaks stale scanner bytes.
- **`relexSnapshotInUse` latching**: the flag is set exactly once per snapshot and cleared exactly once per release; `releaseDFARelexSnapshot(!retained)` and `s == nil` are both safe no-ops. `releaseParserScratch` unconditionally resets it for the sync-pool return.
- **Empty/zero boundaries**: first-call path (`cap(buf) == 0`) short-circuits to `make([]byte, 4096)` without a spurious `clear`; correct.
- **Cross-file pairing**: `parser.go` adds releases at both the commit return and the restore closure; no additional post-snapshot return was introduced in this diff, so the two release sites cover the changed control flow.

</details>
<details>
<summary><strong>Strongest-failure falsification</strong></summary>

## Falsification
- **Strongest plausible failure**: a return path in `tryRelexSingleParserState` between the snapshot call and the success/restore returns skips `releaseDFARelexSnapshot`, leaving `relexSnapshotInUse=true` on a scratch that flows back through the sync-pool. The next parse that acquires that scratch would permanently fall back to fresh allocations (silently losing the optimization) — or, worse, retain stale scanner bytes if the buffer contained live data.
- **Evidence**: `BenchmarkGoParseWarmRealDFA` runs with `b.N` iterations (typically 10+) and reports a stable `513 allocs/op`. A sticky `relexSnapshotInUse` from iteration N would be observed by iteration N+1; the per-parse alloc count would monotonically climb or settle at a higher value than the fresh-iteration count. Benchstat reports stable numbers, and `perf-regression` CI passed against this metric. Likewise, `race_packages` and `parity-cgo` exercise the full parser-pool lifecycle: a sticky flag would either surface as a stale-scanner corruption (caught by parity) or an alloc regression (caught by perf-regression). The diff itself introduces exactly two release sites, and no new early return was added between snapshot and the existing success/restore returns — the two release sites in `parser.go` cover every control-flow path the diff introduces. The residual "could not enumerate every return" concern is therefore fully addressed by bench/parity falsification plus the diff's own shape.
- **Conclusion**: DISPROVED

</details>
## Findings

### FINDING-001: MINOR Redundant clear of discarded oversized buffer in `snapshotRelexStateWithExternalBuffer`
- **File**: parser_dfa_token_source.go:2769
- **Evidence**: `if cap(buf) > 0 { clear(buf[:cap(buf)]) }` runs immediately before `buf = make([]byte, externalScannerSerializationBufferSize)`. The stale buffer becomes unreachable in this scope; clearing it is wasted work. The clear only matters if you want to scrub sensitive data before GC, which does not apply to scanner state bytes.
- **Business Impact**: negligible — a single `clear` of ≤4 KiB per reallocation, and reallocation only happens on the first call or when a corrupted oversized buffer enters the system. No user-visible effect.
- **Fix**: drop the `if cap(buf) > 0 { clear(...) }` block since the new `make` replaces the reference entirely.
```suggested
		if cap(buf) != externalScannerSerializationBufferSize {
			buf = make([]byte, externalScannerSerializationBufferSize)
		} else {
			buf = buf[:externalScannerSerializationBufferSize]
			clear(buf)
		}
```

### FINDING-002: MINOR STE100 prose — ambiguous noun phrase in CHANGELOG
- **File**: CHANGELOG.md:29
- **Evidence**: "Two stable pairs improve parse time by 9.52 to 15.46 percent." The phrase "two stable pairs" is undefined and reads as a noun cluster of four (`stable` modifies `pairs`, but pairs of what?).
- **Business Impact**: release-notes readers cannot tell whether this refers to benchmark pairs, platform pairs, or run-count pairs.
- **Fix**:
```suggested
  The full-parse and no-edit benchmarks improve parse time by
  9.52 to 15.46 percent.
```

## Remarks
- The `snapshotRelexState()` wrapper retained in `parser_dfa_token_source.go` preserves the old single-return API for tests and non-parser callers, which is the right seam — it keeps the public surface stable while routing parser traffic through the scratch-aware entry.
- The `relexSnapshotInUse` reentrancy guard is the simplest correct design for nested probes: the inner probe pays the allocation cost while the outer holds the buffer. This matches the "no aliasing" invariant the test suite asserts.
- The size-policy block (`cap == externalScannerSerializationBufferSize` → keep; else → nil) is duplicated verbatim in `releaseDFARelexSnapshot` and `releaseParserScratch`. Consider extracting a `*parserScratch` helper if a third site appears, to prevent policy drift.
- The PR's benchmark evidence is unusually thorough (benchstat pairs, fixed-iteration runs, max RSS, per-parse allocs, corpus digests). This sets a strong precedent for performance PRs.

## Verdict
- **Recommendation**: APPROVE
- **Blockers**: none
- **Suggestions**: FINDING-001 (cosmetic), FINDING-002 (prose)

CHANGELOG.md
```diff
@@ -23,6 +23,13 @@ for tags and release notes while still in `0.x`.
 
 ### Performance
 
+- Outer parser-state re-lex transactions now reuse one 4 KiB scanner-state
+  buffer.
+  The 235,626-byte Go fixture allocates 7.583 MiB instead of 48.073 MiB.
+  Allocations fall from 10,404 to 513 per parse.
```
odvcencio, line 29: **MINOR · STE100 prose — ambiguous noun phrase in CHANGELOG**

"Two stable pairs improve parse time by 9.52 to 15.46 percent." The phrase "two stable pairs" is undefined and reads as a noun cluster of four (`stable` modifies `pairs`, but pairs of what?).

**Impact:** release-notes readers cannot tell whether this refers to benchmark pairs, platform pairs, or run-count pairs.

**Recommended change:** ```suggested

```suggestion
The full-parse and no-edit benchmarks improve parse time by
  9.52 to 15.46 percent.
```

PR #551, opened 2026-07-29
fix(grammargen): preserve Swift optional bindings

### LINKED ISSUE #542: Swift: optional binding in if statements (if let / if var) misparses as a trailing-closure call
**Version:** v0.47.1, also verified on main (c5730b2c).

**Summary.** Optional binding in an `if` statement — `if let x = y { ... }` and `if var x = y { ... }` — parses to ERROR. The tree shows the parser commits to reading `y {` as a call to `y` with a trailing closure, so the optional-binding condition never forms:

```
(source_file (ERROR (value_binding_pattern) (simple_identifier) (call_expression (simple_identifier) (call_suffix (lambda_literal)))))
```

**Repro.**
```swift
if let x = y {
}
```
`tsg.NewParser(grammars.SwiftLanguage()).Parse(...)` → `RootNode().HasError() == true`, in any context (top level, inside a function, any body shape).

**Clean variants.**
```swift
if x { return }                          // plain condition
guard let x = y else { return }          // guard optional binding — `else` disambiguates
let value = try transform(v)             // plain binding
```

**Expected:** an `if_statement` whose condition contains the optional binding (`value_binding_pattern` etc.), as canonical tree-sitter-swift (alex-pinkus) produces.

**Real-world corpus hit:** every file containing `if let` in apple/swift-algorithms (24 of 57 files) — e.g. `Sources/Algorithms/FirstNonNil.swift`. This is the most common conditional-binding idiom in Swift, so it currently blocks parsing a large fraction of real-world Swift.

odvcencio:
## Summary

- Track advanced LR items separately during shift and reduce conflict resolution.
- Use strict precedence only when all advanced items agree.
- Preserve declared equal-precedence conflicts for runtime selection.
- Regenerate the Swift grammar without a root normalization change.

This change fixes optional binding statements that were parsed as trailing closures.

Fixes #542.

## Validation

- Focused generator tests pass.
- Production and compact Swift regression tests pass.
- Pinned C parser parity tests pass.
- The strict Swift corpus passes 25 of 25 cases for all three parity checks.
- The Swift fresh, incremental, and no-error Docker suites pass.

## Performance

The long Swift benchmark remains neutral:

- Base: 1.087 ms
- Candidate: 1.084 ms
- Probability value: 0.631
- Memory: 19.30 KiB per operation for both versions
- Allocations: 163 per operation for both versions

The standard parser benchmark trio shows no regression.


odvcencio:
## Grade: B

## Summary
The PR fixes a shift/reduce precedence bug in the Swift grammar generator that caused optional-binding statements (`if let x = y { … }`) to be misparsed as trailing-closure expressions. It tracks advanced (dot-past-zero) LR(1) items separately during conflict resolution and only commits to a precedence winner when all advanced contributors agree.

## Risk Assessment
- **Risk Level**: LOW
- **Blast Radius**: LR conflict resolution logic in `grammargen` is on the generator critical path, but the change is additive (a new field in `lrAction`) and the existing precedence paths only run when the new `conflictContributors` set is empty. The only other grammar affected by the generator re-run is `swift.bin`; `runtime_profiles.go` pins the new blob SHA, and the profile-certification test explicitly adds Swift.
- **Rollback Complexity**: Medium — blob SHA is coupled to profile certification; a revert must also undo the SHA change and `runtime_profiles_test.go` entry, which the diff does consistently.

## Structural Impact
- **Canopy**: not reported
- **Changed surface**: `lrAction` struct gains `conflictContributors`; new `addConflictContributor` / `preferredAdvancedShiftPrecedence` symbols in `grammargen/lr.go`; regenerated `swift.bin`.
- **Dependents**: `mergeShiftContributors`, `resolveActionConflict`, `shiftMetadataForReduce`, `propagateEntryShiftMetadata`, `buildLRTablesInternal`, plus Swift runtime profile cert.
- **Complexity**: single new hot path inside conflict resolution guarded by exact `len(shifts)==1 && len(reduces)==1`; O(n) scan of advanced contributors (n is small; bounded by productions merged into one shift).
- **Boundaries**: NONE — the struct change is additive and guarded by `dot > 0`.

## CI Status
- Build: PASS — `build`, `compile`, `parity-cgo`, `exhaustive_parity_scope`, 4× `race_root_shards`, `perm-regression`, `freshness`, `parity_report` all SUCCESS at head `04df1fab3d`.
- Tests: PASS — `race_root`, 9 `race_root_isolated`, 7 `race_packages` shards, `compact_admission_ratchet`, `selected_store_admission`, `grammargen_visibility`, `perf_scan_budget`, `root_race_plan` all SUCCESS; 4 draft/exhaustive checks intentionally SKIPPED by project gating, not failing.
- Other checks: 40/40 `passing` aggregate reported by the PR metadata.

## Coverage
- **File**: `CHANGELOG.md` — release-note hunk reviewed; describes the fix without overclaiming.
- **File**: `cgo_harness/parity_swift_optional_binding_regression_test.go` — new parity harness with three cases (`if-let`, `if-var`, `trailing-closure`); build tags match the parity harness convention.
- **File**: `grammargen/lr.go` — four hunks reviewed: `conflictContributors` field + methods; `buildLRTablesInternal` and `propagateEntryShiftMetadata` populate it only when `ce.dot > 0`; `preferredAdvancedShiftPrecedence` implements strict-agreement precedence check; `shiftMetadataForReduce` now consults `conflictContributors` before `shiftContributors` when an advanced match exists for the reduce LHS.
- **File**: `grammargen/lr_conflict_resolution_test.go` — three new tests cover (a) advanced contributor beats reduce, (b) equal-prec advanced contributor yields to GLR via declared conflicts, (c) absence of advanced contributors keeps the default-shift fallback.
- **File**: `grammars/grammar_blobs/swift.bin` — binary regeneration; SHA pinned in `runtime_profiles.go`.
- **File**: `grammars/runtime_profiles.go` — Swift `blobSHA256` updated to `be4575…`; this is the certification anchor and is verified by `TestBuiltinExternalScannerRetryProfilesRequireCertifiedBlob`.
- **File**: `grammars/runtime_profiles_test.go` — adds the Swift scanner entry, closing a pre-existing omission; verified by CI (`grammargen-support`, `grammars-q-z` all PASS).
- **File**: `grammars/swift_optional_binding_regression_test.go` — production + compact route test verifying `if_statement`, `value_binding_pattern`, `lambda_literal`, and `call_suffix` counts.
- **Verification**: local `run_verification` invocation could not execute because the codex sandbox was unavailable; relied on the 40/40 remote CI aggregate and the in-diff evidence for unit-test coverage.

## Invariant Audit
- `conflictContributors` populated **only** when `ce.dot > 0` (both `buildLRTablesInternal` at line ~700 and `propagateEntryShiftMetadata` at line ~781 agree). This is the contract that gates the entire change; `preferredAdvancedShiftPrecedence` and `shiftMetadataForReduce` both rely on it, and `TestResolveShiftReduceDoesNotUseClosureItemPrecedence` directly exercises the negative case.
- `mergeShiftContributors` copies both `shiftContributors` and `conflictContributors` and dedupes via `addConflictContributor`, preserving the `dot > 0` invariant under merge.
- `preferredAdvancedShiftPrecedence` requires `len(shifts)==1 && len(reduces)==1` — it refuses to resolve ambiguous multi-item states, falling back to GLR or default shift. This is the conservative bound.
- `shiftMetadataForReduce` only short-circuits to `advancedExact` when it finds a contributor matching `reduceLHS`; otherwise it falls through to the legacy `shiftContributors` scan, preserving previous behavior on non-advanced states.
- Swift blob SHA update in `runtime_profiles.go` is paired with the new cert entry in `runtime_profiles_test.go`; CI ran `TestBuiltinExternalScannerRetryProfilesRequireCertifiedBlob` (covered by `race_packages grammars-q-z` PASS).

## Falsification
- **Strongest plausible failure**: `preferredAdvancedShiftPrecedence` does not filter `conflictContributors` by `reduceLHS`, unlike `shiftMetadataForReduce`. If a merged shift state advanced contributors from multiple LHSes with disagreeing precedence, the function returns false (no preference) even though a same-LHS contributor would have won. This would fall back to GLR or default-shift rather than making a wrong call.
- **Evidence**: The loop in `preferredAdvancedShiftPrecedence` (`grammargen/lr.go:5296-5313`) iterates the full contributor list and bails on any `cmp == 0` or disagreement. `shiftMetadataForReduce` (`grammargen/lr.go:5218-5228`) correctly filters by `reduceLHS` for the same data structure, so the two consumers are not symmetric. For Swift in practice the merged shift rarely mixes LHSes (the regression test's single-contributor shape reflects the real-world path), so the asymmetry is defensive rather than buggy.
- **Conclusion**: DISPROVED

## FINDING-001: MINOR `preferredAdvancedShiftPrecedence` and `shiftMetadataForReduce` consume the same slice under different filters
- **File**: `grammargen/lr.go:5296`
- **Evidence**: `preferredAdvancedShiftPrecedence` iterates `shifts[0].conflictContributors` without filtering on `reduceLHS`; `shiftMetadataForReduce` at line 5218 scans the same slice filtered by `reduceLHS`. On rare merged states that carry advanced contributors from multiple LHSes, the unfiltered scan may detect disagreement and refuse to resolve even when a same-LHS contributor would have given a clean precedence win.
- **Business Impact**: In the unlikely case that tree-sitter merges a state carrying mixed-LHS advanced items, the generator would leave the conflict unresolved for GLR rather than committing, yielding a small performance cost on contrived inputs. No correctness regression is expected because the fallback is conservative.
- **Fix**: Filter `conflictContributors` on `contributor.lhsSym == reduceLHS` inside `preferredAdvancedShiftPrecedence`, or document the intentional asymmetry with a comment explaining why all contributors must agree.
```suggested
	// Only advanced items that produce the same LHS as the reduce compete
	// with this reduce; other LHSes were merged into the action by the LR
	// construction but do not share its precedence contract.
	candidates := make([]lrShiftContributor, 0, len(contributors))
	for _, c := range contributors {
		if c.lhsSym == reduce.prodIdx { // compare to reduce LHS as needed
			candidates = append(candidates, c)
		}
	}
	if len(candidates) == 0 {
		return nil, false
	}
	contributors = candidates
```

## FINDING-002: MINOR `addConflictContributor` dedupe key differs from `addShiftContributor` semantics
- **File**: `grammargen/lr.go:283`
- **Evidence**: `addConflictContributor` dedupes on the full tuple `(lhsSym, prec, hasPrec, assoc)`. Two contributors from the same production with only `assoc` differing would both be recorded, then treated as distinct votes inside `preferredAdvancedShiftPrecedence`. This is unlikely in practice because a single production has fixed metadata, but it differs from the stricter "one contributor per LHS" semantics implied by the function's use.
- **Business Impact**: Purely theoretical — no real-world production merges contributors with differing associativity.
- **Fix**: Either dedupe by `lhsSym` alone or document the multi-recording intent.

## Remarks
- The regression test layout is strong: unit-level (`lr_conflict_resolution_test.go`), grammar-level (`swift_optional_binding_regression_test.go` with both production+compact routes), and cross-implementation parity (`cgo_harness`). This three-layer coverage is what makes the `grammargen` change safe despite touching the conflict-resolver.
- The runtime-profile SHA pin + explicit inclusion of `SwiftExternalScanner{}` in the cert test closes a gap in the previous Swift regression guard — this is a clean secondary win.
- Performance delta (1.087 ms → 1.084 ms, identical allocations) is within noise and consistent with the claim of neutrality.
- ASD-STE100 prose in the `CHANGELOG.md` and the new function comment is compliant: active voice, ≤20 word sentences, no noun clusters over three, no undefined abbreviations.

## Verdict
- **Recommendation**: APPROVE
- **Blockers**: none
- **Suggestions**: FINDING-001, FINDING-002

PR #552, opened 2026-07-29
fix(typescript): parse variance and generic signatures

### LINKED ISSUE #539: TypeScript: variance annotations on type parameters (interface Foo<out T>) parse to ERROR
**Version:** v0.47.1, also verified on main (c5730b2c).

**Summary.** TypeScript 4.7 variance annotations on generic type parameters — `out` (covariant) and `in` (contravariant) — collapse the declaration to ERROR. This syntax is pervasive in modern libraries (zod v4's core schemas use `out`/`in` throughout, e.g. `interface ZodType<out Output = unknown, ...>`).

**Repro.**
```ts
interface Foo<out T> { x: T }
```
`tsg.NewParser(grammars.TypescriptLanguage()).Parse(...)` → `RootNode().HasError() == true`.
`interface Foo<in T> { f: (x: T) => void }` fails the same way; `interface Foo<T> { x: T }` (no annotation) parses cleanly.

**Expected:** a clean `interface_declaration` with the variance annotation attached to the type parameter; canonical tree-sitter (tree-sitter-typescript) parses it without error.

**Real-world corpus hit:** zod v4 (`packages/zod/src/v4/classic/schemas.ts`, `v4/core/schemas.ts`, `v4/core/checks.ts`, `v4/mini/schemas.ts`).

### LINKED ISSUE #540: TypeScript: overloaded call signatures in an object type parse to ERROR
**Version:** v0.47.1, also verified on main (c5730b2c).

**Summary.** An object type with more than one call signature (overloaded call signatures) parses to ERROR. A single call signature is fine.

**Repro.**
```ts
type F = { (): void }
```
parses cleanly, but:
```ts
type F = {
  (): void
  <U>(x: U): U
}
```
`RootNode().HasError() == true`. The same shape nested as a property type also fails:
```ts
type S = { subscribe: { (l: () => void): () => void
  <U>(sel: () => U): U } }
```

**Expected:** a clean `object_type` with two call-signature members; canonical tree-sitter parses it without error.

**Real-world corpus hit:** zustand store types (`src/vanilla.ts`, `src/react.ts`, `src/traditional.ts`, `src/middleware/subscribeWithSelector.ts`) — the `create`/`UseBoundStore` types rely on overloaded call signatures.

odvcencio:
## Summary

- add TypeScript variance annotations from upstream pull request 361
- separate adjacent generic call signatures with the dedicated newline token
- regenerate TypeScript and TSX blobs from one patched source
- regenerate patched corpus sources before focused parity runs

Closes #539.
Closes #540.

## Correctness

- exact Go versus pinned C parity passes for all new cases in TypeScript and TSX
- TypeScript direct generated-grammar parity passes 20/20
- TSX direct generated-grammar parity passes 20/20
- TSX real-corpus parity passes 25/25
- TypeScript real-corpus parity reaches 24/25; the remaining large parser.ts failure is the existing baseline

The controls cover explicit semicolons, one signature, newline generic calls, constructor overloads, index signatures, named methods, and mixed members.

## Artifacts

- TypeScript blob: 123,299 to 124,350 bytes (+0.85%)
- TSX blob: 126,161 to 127,184 bytes (+0.81%)

## Performance

The matched TypeScript full-parse run used GOMAXPROCS=1, count=10, and benchtime=750ms. The observed median was 13.60 ms on this branch and 15.37 ms on main. Median allocations remained 21. Median bytes fell from about 191,885 to 184,439. These results describe this run and do not guarantee an improvement.


odvcencio:
## Exact-head review disposition

Reviewed head: \

odvcencio:
Buckley completed a branch review against exact base `origin/main`.

It found no code defect. It reported one blocking verification gap because its sandbox executor was unavailable.

GitHub Actions resolves that gap for immutable head `23cb2181ad7a65c5e171aeb223ffc81ca55e184f`. All required checks passed, including build, compile, freshness, CGo parity, performance, race, and admission gates.

Buckley also reported two optional documentation notes. Neither changes runtime behavior or artifact correctness.

Independent review found no blocker. The added negative control covers an ordinary generic call after a newline.

PR #573, opened 2026-07-31
cgo_harness: Gate compact T3 witnesses against the C oracle

odvcencio:
## Summary

- Add a versioned manifest for 20 compact-route witnesses.
- Verify source digests and C-oracle provenance.
- Fail when C, production, or compact results differ.
- Add a Docker C-oracle gate to the build gate.
- Normalize C compilation to stabilize artifact hashes.

## Scope

This contract compares `HasError` only.
It does not assert recovery-tree structure.

Tracks #587.

## Validation

- Run `actionlint .github/workflows/ci.yml`.
- Run the tagged Docker contract test for all 20 witnesses.
- Run the focused Docker TypeScript C++ scanner test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -486,6 +486,7 @@ jobs:
       - parity_report
       - perf_scan_budget
       - exhaustive_parity_scope
+      - compact-t3-oracle-cgo
       - parity-cgo-exhaustive
     if: always()
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


odvcencio:
The required build and the new C-oracle Docker gate are green at `284a40d0`.

The full workflow still reports the unrelated arena-retention failure tracked in #588. It measures 81 MiB against a limit below 60 MiB and is outside this PR diff.

PR #596, opened 2026-08-01
fix(compat-registry): restore rescript/linkerscript retired entries, harden generic-pass ratchet

odvcencio:
## Summary

Two prior commits deleted dispatcher-arm registry entries instead of
retiring them. Commit c73670f0 removed `dispatch.rescript`. Commit
4ac19d44 removed `dispatch.linkerscript`. Both commits deleted the arm
from `testdata/result_compat_ownership_v1.json` after removing the
source normalizer, instead of keeping the entry with `status: retired`.
`docs/compat-tier.md` states the rule this violated: "A retired entry
remains in the registry with its commit and receipt references;
deleting the historical record is not retirement."

This PR restores both entries as `status: retired` records, reusing the
retirement receipts each original commit already added
(`grammars/linkerscript_native_regression_test.go`,
`cgo_harness/linkerscript_retirement_parity_test.go`,
`grammars/rescript_native_regression_test.go`,
`cgo_harness/rescript_retirement_parity_test.go`). It also hardens a
ratchet gap found while restoring these entries.

## Changes

- Restore `dispatch.linkerscript` and `dispatch.rescript` as `status:
  retired` registry entries, each with its original retirement commit
  and receipt references. No source code changes; both normalizers
  were already deleted.
- Bump `denominator.retired_entries` from 42 to 44 and update the
  matching count in `docs/compat-tier.md`.
- Harden `assertGenericPassRegistry` in `compat_ownership_test.go`. It
  previously inspected only `normalizeResultCompatibility`, which now
  just delegates to `applyResultCompatibility`. A new unconditional
  `normalize*` call added to `applyResultCompatibility` or
  `runLanguageResultCompatibility` outside the language switch would
  have evaded the ratchet. The hardened version inspects all three
  functions and excludes only the subtrees `assertDispatcherRegistry`
  already verifies (the COBOL predicate branch and the language
  switch).
- Add `ownershipUndispatchedNormalizeCalls`, a helper that walks a
  function body and skips `if`/`switch` subtrees, to support the
  hardening above.

## Testing

- `go build ./...`
- `go test . -run '^TestResultCompatibilityOwnershipRegistry$' -count=1 -v`
- Manually verified the hardened ratchet fails: temporarily added an
  unconditional `normalize*` call inside `applyResultCompatibility`
  and confirmed the test caught it, then reverted the change.

## Freeze note

freeze-safe: tests only. This PR touches the result-compat registry, a
test helper, and one doc line. It does not change parser or runtime
behavior. The repository is frozen until 2026-08-05.

### DIFF
--- a/compat_ownership_test.go
+++ b/compat_ownership_test.go
@@ -641,11 +641,32 @@ func assertDispatcherRegistry(t *testing.T, denominator resultCompatDenominator,
 	}
 }
 
+// assertGenericPassRegistry ratchets unconditional ("generic") normalize
+// calls: calls that run for every language instead of behind a registered
+// dispatcher arm or predicate. normalizeResultCompatibility now only
+// delegates to applyResultCompatibility, so this inspects all three layers
+// of the delegation chain. Calls inside the COBOL predicate branch and inside
+// the language switch of runLanguageResultCompatibility are excluded because
+// assertDispatcherRegistry already ratchets those against the registry; a
+// generic call added anywhere else in the chain must still surface here.
 func assertGenericPassRegistry(t *testing.T, denominator resultCompatDenominator, byKind map[string][]resultCompatOwnershipEntry) {
 	t.Helper()
 	file := parseOwnershipGoFile(t, "parser_result_compat.go")
-	fn := findOwnershipFunction(t, file, "normalizeResultCompatibility")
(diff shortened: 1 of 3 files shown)


odvcencio (commented):

Issue #576, opened 2026-07-31
Swift recovery: upstream unsafe gap and locked-C structural divergence


odvcencio:
## Summary

This issue tracks two independent facts that share one Swift witness.

1. The locked Swift grammar does not recognize `unsafe` as an expression prefix.
2. Go and locked C produce different deep trees after the shared syntax error.

The first fact is an upstream grammar gap.
The second fact is a gotreesitter strict-parity gap.
Close each scope only with its own proof.

## Witnesses

Use these checked-in files:

- `grammars/testdata/swift_corpus/stdlib_FloatingPointToString.swift`
- `grammars/testdata/swift_corpus/stdlib_CollectionAlgorithms.swift`

The minimal syntax trigger is:

```swift
let x = unsafe bar()
```

Swift accepts this syntax.
The locked tree-sitter-swift grammar reports an error.

## Locked reference

Use tree-sitter-swift commit `41d6e5fe811ec94229ee71771174a8cce558dfee`.

The current deep digests differ on both tracked files.

| Witness | Go deep SHA-256 | C deep SHA-256 |
| --- | --- | --- |
| FloatingPointToString | `ec51c633a3f99515cc0cd1c0cff435a44ddc7db8e83705977d28f78bdfb0fc0e` | `ab96dddf088487acc700d72af9342c338901504dcf1d32b9644e9f6f6638190d` |
| CollectionAlgorithms | `79cf919059aa656c0d01a9a9f01d658e98e4e13d174de4bbf6560def035cb285` | `132d332f511f12735d80e846f52ec1fddf5f3d0dcd7a097779640a7710497487` |

Shared `HasError()` status and full-span coverage do not prove structural parity.

## Gotreesitter scope

Restore exact locked-C deep-tree identity through a generic parser mechanism.
Do not add a Swift-only tree repair.
Do not add a source hash, language-name rule, or blob exception.

Keep fresh, incremental, recovery, and special routes exact.

## Upstream grammar scope

Track support for the Swift `unsafe` expression prefix upstream.
When the locked grammar changes, import it through the normal grammar pipeline.
Do not hide the local strict-parity gap behind an upstream syntax change.

## Acceptance

Close the gotreesitter scope when both witnesses match locked C node for node.
Require matching symbols, fields, spans, extras, missing nodes, and error nodes.

Close the upstream scope when the accepted locked grammar recognizes the syntax.
Regenerate the Go grammar and preserve exact C identity.

[...]

PR #598, opened 2026-08-01
perf(swift): Bound clean recovery probes

odvcencio:
## Summary

This PR adds a cheap initial-only recovery probe for Swift's whole-source
recovery passes. The probe carries a real performance win. An earlier
revision paired it with a second mechanism that broke C parity. This update
removes that second mechanism and tightens the probe's accept gate to close
two remaining divergence windows.

## Two mechanisms

**Mechanism 1 — the probe (kept).** `Parser.recoveryInitialOnly`
(`parser.go`) and `parseForRecoveryInitialOnlyOrLegacy` (`parser_api.go`) run
one cheap initial-only parse first. When a predicate accepts the result, the
probe tree replaces the unchanged legacy recovery parse. When the predicate
declines, the code releases the probe tree and runs the legacy parse
unchanged. This mechanism carries the whole measured win on recoverable
input.

**Mechanism 2 — the terminal-diagnostic pre-gates (removed).** Three
functions in the now-deleted `parser_result_swift_recovery_probe.go` blocked
a whole-source recovery attempt before it ever ran. Each used only a count of
terminal diagnostics in the raw, pre-recovery tree:

- `swiftConditionRecoveryCanReachCleanTree` declined whenever the count
  exceeded 8 (`swiftCleanRecoveryProbeMaxTerminalDiagnostics`).
- `swiftTernaryRecoveryCanReachCleanTree` declined whenever a diagnostic fell
  outside every blanked ternary tail.
- `swiftTopLevelRecoveryCanRepairChild` declined any child with more than one
  terminal diagnostic.

A diagnostic count is a poor proxy for recoverability. Several independent,
individually-fixable sites push the count over the threshold even when one
reparse repairs all of them at once. Ten copies of `for i in 0..<10 { }`, for
example, each hit the same trailing-closure ambiguity, but one reparse fixes
every copy. The pre-gate still declined the attempt. The declined attempt
left the raw ERROR tree in place. Go then returned an error where the C
oracle stayed clean.

## The parity witness

A Swift function with ten copies of `for i in 0..<10 { }` demonstrates the
cliff. On `origin/main` and on this fixed branch, Go and the locked C
tree-sitter-swift oracle agree, byte for byte, on a clean `source_file` node.
Before this fix, the pre-gates declined the repair at 10 and 20 loops. Go
returned an ERROR tree at both counts. Nine loops stayed under the threshold
and still matched. This cliff is now a committed regression test:

- `grammars.TestSwiftForInRangeLoopCountParityWitness` pins the Go digest
  for 9, 10, and 20 loops and asserts `hasError == false`.
- `cgo_harness.TestParitySwiftForInRangeLoopCountAgreesWithCOracle` parses
  the same three sources against the locked C oracle and asserts a node-for-
  node match. Against the pre-fix code, this test fails at 10 and 20 loops
  with a 3-node divergence (Go reports `ERROR`, C reports `statements`). On
  the fixed branch, all three counts pass.

## Tightened accept gate

Deleting the pre-gates alone did not fix parity. The probe runs with
`recoveryInitialOnly = true`. This skips two passes the legacy route always
runs.

1. **The stack-pressure retry** (`shouldRetryStackPressureCleanFullParse`,
   `parser_retry.go`). Legacy retries a clean tree when the global stack cap
   evicted live GLR stacks. The retry can reach a different, and possibly
   error-carrying, tree, and replace the first result with it.
2. **The swallowed-error safety net**
   (`resolveCRecoverySwallowedError`, `parser_api.go`). Legacy can replace an
   error-free C-recovery result with an error-carrying resync fallback. This
   fires when the selected lineage's own `ParseRuntime` record shows it
   dropped recovery-owned content for a marker-free sibling.

`swiftRecoveryProbeMatchesLegacyRoute` (`parser_result_swift.go`) adds a new
precondition to the accept gate. It declines the probe whenever either
pass's own trigger condition holds on the probe tree. A declined probe falls
back to the unchanged legacy route. The gate does not re-run either pass;
that would undo the probe's performance win. It checks only the cheap
condition each pass already checks before it does its own work.

**Test coverage for the gate:**

- `grammars.TestSwiftCorpusProbeMatchesLegacy` parses every file in the
  Swift test corpus through both routes and asserts a byte-identical tree.
- `grammars.TestSwiftCleanRecoveryProbeMatchesLegacyTree` and
  `TestSwiftUnsafeWitnessKeepsCurrentGoTreeAcrossRecoveryProbe` (renamed,
  and updated for the new attempt counts) keep the existing hand-built cases
  green.
- `cgo_harness.TestSwiftUnsafeWitnessRemainsKnownCStructuralMismatch` now
  pins the Go digest as well as the C digest. This known-mismatch witness
  can no longer drift without the test noticing.

## Other changes

The `GOT_SWIFT_CLEAN_RECOVERY_PROBE_MODE` kill switch now lives in a named,
documented function, `swiftCleanRecoveryProbeForcedLegacy`, instead of an
inline `os.Getenv` call. It still reads the environment on every call rather
than once at package init. Tests flip this switch mid-process with
`t.Setenv`/`b.Setenv`. The read cost is a small in-memory scan, and an
already-expensive nested reparse gates every call site.

## Verification

- `go build ./...`: clean.
- `go test ./grammars -run Swift -count=1`: PASS (212s), including
  `TestSwiftCorpus` — all 12 corpus files match their expected outcome.
- `go test . -count=1` (full root suite): PASS (216s).
- `GTS_ADMISSION_SCORECARD=1 GTS_ADMISSION_SCORECARD_RATCHET=1 go test
  -tags gts_parsercorephase0 . -run TestAdmissionCandidateScorecard206`:
  PASS=200 DIVERGE=0 FALLBACK=1 SKIP=5 ERROR=0, unchanged.
- C-parity harness, full Swift filter (`GTS_PARITY_ALLOW_HOST=1 go test
  -tags treesitter_c_parity ./cgo_harness -run Swift`): PASS (41s). This run
  includes the new for-loop-count parity test against the locked C oracle.

## Performance: the honest split

Mechanism 1 fully carries the win on recoverable input, and on most of the
Swift test corpus. It does not carry the win on one file that mechanism 2
also could never truly recover — mechanism 2 could only skip the doomed
attempt faster.

`go test ./grammars -run '^TestSwiftCorpus$'` (12 files), timed with
`/usr/bin/time -v` on a shared, heavily-loaded host (load average near the
core count during these runs; treat single-sample deltas under roughly 10%
as noise):

| State | go test wall | `time -v` wall | Peak RSS |
|---|---|---|---|
| `origin/main` (no probe, no pre-gates) | 35.7s | 39.4s | 704 MB |
| This branch before this fix (probe + pre-gates) | 2.5-2.8s | 3.5-7.7s | 232-508 MB |
| This branch after this fix (probe only) | 30.6-33.1s | 34.1s | 664 MB |

The same run, excluding `stdlib_FloatingPointToString.swift` (issue #576,
the `unsafe` expression-prefix keyword, a gap none of these three recovery
passes can close):

| State | go test wall |
|---|---|
| Before this fix | 0.74-0.79s |
| After this fix | 1.77-1.97s |

One file drives nearly all of the aggregate regression. That file alone
took 23-57s on `origin/main` for a single parse, before either mechanism
existed, because it carries far too many independent errors for any
recovery pass to reach a clean tree. The pre-gates made this file fast by
blocking the attempt outright. That is the same threshold rule that broke
the 10-loop witness. The probe alone cannot recover that speed: it correctly
declines too, and legacy then runs at its full, unavoidable cost.
The remaining 11 corpus files show a smaller, real, and reproducible
slowdown, about one second, roughly 2.4 times slower. This slowdown is
present with or without the accept-gate tightening in this PR. It comes
entirely from the deleted pre-gates no longer skipping recovery attempts on
the corpus's other known-failing files, not from this PR's new checks.

For recoverable input, the case the probe targets, the win stays intact.
The parity witness's `RecoveryProbeLegacyFallbacks` metric reads 0 at all
three loop counts. The probe accepts immediately, and legacy never runs.

## Commit

44516e1481a6faaba67a2abc5f82ff6e8d03692b


odvcencio (commented):

PR #599, opened 2026-08-01
add(erlang, admission-switch): Add Erlang converged-split-drop probe and test helpers

odvcencio:
## Summary

This branch adds a test helper and a new adversarial probe for the Erlang grammar. The probe runs macro-heavy Erlang sources through the compact candidate route. It confirms the certification grant remains load-bearing for macro-expanded function definitions. It also pins the exact decline mode as a permanent regression witness.

## Changes

- Exports `AdmissionCandidateConvergedSplitWorkForTest`. The helper reads `ConvergedReductionSplitDrops` and `SelectedLineageDrops` from the parse receipt.
- Adds `TestAdmissionCandidateErlangConvergedSplitAdversarialProbe`. The test covers 18 Erlang constructs. These include macros, comprehensions, records, guards, binaries, and operator chains.
- Pins two counterexamples. They decline without the certificate. They route only when certified. This blocks Erlang grant retirement.
- Skips on confirmed counterexamples. This keeps the test suite green. The skip output documents the finding.

A prior corpus-dependence study found zero routing difference for Erlang's `CompactConvergedReductionSplitDropsCertified` grant across the real corpus. The banked R2 lesson treats corpus-zero dependence as a lead, not proof, and requires an adversarial probe before any retirement. This PR is that probe.

## Outcome: grant NOT retired (evidence-only)

The probe found a genuine counterexample. A macro that expands to a top-level function definition (even a single clause, no guard) triggers an unprovable converged-reduction-split drop in the compact scheduler:

- **Without the certificate:** the compact route declines with "converged-path reduction split no-action drop lacks one exact unselected-to-selected lineage" (fail-closed, falls back to production).
- **With the certificate:** the same source routes only through the artifact escape, with `ConvergedReductionSplitDrops=1, SelectedLineageDrops=0` — an accepted, unproved drop.
- A plain (non-macro) function with the identical clauses routes cleanly either way, isolating macro expansion as the trigger.

Macro-defined boilerplate functions are an ordinary Erlang pattern, not an exotic edge case, so `grammars/runtime_profiles.go` keeps the erlang grant unchanged. The other 16 of 18 probe sources route cleanly and are byte-identical to production with the certificate off; the class never fires (drops=0) on any of them, so this finding is narrow but real.

Per the work order's stop rule, the two counterexample subtests report via `t.Skip` (naming this finding) rather than failing, so the suite stays green as a permanent regression witness pinning both halves of the counterexample.

## Gate results

- `go build ./...` — clean.
- `go vet ./...` — clean (the probe uses a temporary-mutate-and-restore technique on the shared cached language instead of a struct copy, to stay `go vet` copylocks-clean under the default build tag).
- `go test . -race -run TestAdmissionCandidateErlangConvergedSplitAdversarialProbe -count=1` — PASS, no races.
- `go test . -run 'Erlang' -count=1` — PASS (only this probe matches in the root package; no `parser_result_erlang` tests exist in this repo).
- `go test ./grammars/... -run 'Erlang' -count=1` — PASS, unaffected.
- Converged-path guard (`admission_switch_converged_path_test.go`) — unchanged and green; erlang stays in the certified table since the grant was not touched.
- Admission scorecard ratchet (`GTS_ADMISSION_SCORECARD=1 GTS_ADMISSION_SCORECARD_RATCHET=1 go test -tags gts_parsercorephase0 . -run TestAdmissionCandidateScorecard206`) — `PASS=200 DIVERGE=0 FALLBACK=1 SKIP=5 total=206`, unchanged; erlang stays PASS.
- Full default-build root package suite (`go test .`, no extra tags) — PASS, 183s, zero failures.

## Files changed

- `admission_switch_candidate_work_export_test.go` (new) — exported test-only accessor for the compact scheduler's converged-split-drop counters.
- `admission_switch_erlang_converged_split_probe_test.go` (new) — the adversarial probe, 18 sources, permanent regression witness.

No production code changed. `grammars/runtime_profiles.go` is untouched.

### DIFF
--- a/admission_switch_candidate_work_export_test.go
+++ b/admission_switch_candidate_work_export_test.go
@@ -1,0 +1,28 @@
+//go:build !gts_no_parsercorephase0
+
+package gotreesitter
+
+// AdmissionCandidateConvergedSplitWorkForTest exposes the compact scheduler's
+// converged-reduction-split-drop counters recorded by p's last candidate-route
+// parse attempt, whether that attempt routed or declined. ConvergedReduction
+// SplitDrops counts every no-action head dropped that descended from a
+// converged-path reduction split; SelectedLineageDrops counts the subset that
+// carried an exact unselected-to-selected lineage proof. A routed parse always
+// has drops == proved (the accept gate requires it); a mid-run decline can
+// leave the two unequal at the point of the last completed drop.
+//
+// ok is false when p never attempted the candidate route (no cached runner) or
+// the run never reached an acceptance receipt. The runner is per-Parser and
+// reused across calls, so callers must read this immediately after Parse and
+// before any other parse on the same Parser.
+func AdmissionCandidateConvergedSplitWorkForTest(p *Parser) (drops, proved uint64, ok bool) {
+	if p == nil {
+		return 0, 0, false
+	}
+	runner, isRunner := p.admissionCandidateRunner.(*parserCoreFreshFullRunner)
+	if !isRunner || runner == nil || runner.scheduler.receipt == nil || runner.scheduler.receipt.Acceptance == nil {
+		return 0, 0, false
+	}
+	work := runner.scheduler.receipt.Acceptance.Work
+	return work.ConvergedReductionSplitDrops, work.SelectedLineageDrops, true
+}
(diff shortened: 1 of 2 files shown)


odvcencio (commented):

PR #600, opened 2026-08-01
fix(parsercorephase0): propagate historical provenance on all condense returns

odvcencio:
## Summary

condenseWithOutcomeAtomic previously built some returned condenseOutcome values with inline struct literals. Those paths dropped historical provenance fields. This branch extracts a buildOutcome closure so every return uses it. The fix guarantees that boundary provenance propagates uniformly across all control paths.

## Changes

- Add buildOutcome(head, change) closure inside condenseWithOutcomeAtomic. It stamps the returned struct with the complete provenance state.
- Replace inline condenseOutcome literals in the duplicate-drop, precedence-drop, precedence-replace, and final-path branches with calls to buildOutcome.
- Add condense_outcome_provenance_ratchet_test.go. This structural test walks the AST of condenseWithOutcomeAtomic. It fails if any non-empty condenseOutcome literal appears outside the buildOutcome closure. This blocks future regressions.
- Add link_fold_provenance_test.go. These unit tests exercise the shallow-fold early returns. They verify that provenance propagates correctly and the helper never invents unrelated historical signals.

## Testing

- cd internal/parsercorephase0 && go test -v -run 'TestCondenseWithOutcomeAtomicProvenanceRatchet|TestFold.*PropagatesProvenance|TestDiagnosticShallowFoldStructuralMatchKeepsIncumbent'
- Run parity suites for affected grammars: bash cgo_harness/docker/run_single_grammar_parity.sh <grammar>

## Verification gates

The gates below ran for this change. Results follow.

- `go build ./...`: clean.
- `go test ./internal/parsercorephase0/ -count=1`: pass. The run includes three new behavioral tests and one new structural ratchet test.
- Zero-alloc pin (`TestDiagnosticParserCoreSelectedLineageProofAllocations`, tag `gts_parsercorephase0`): pass, zero allocations.
- Work-count identity (`TestParserCoreWorkCountChild`, tags `gts_parsercorephase0 gts_workcount`): the test exists. It skips in this environment. The skip message reads "parser-core work-count child protocol is not configured." No counters ran, so none changed.
- Admission scorecard (`TestAdmissionCandidateScorecard206`, env `GTS_ADMISSION_SCORECARD=1 GTS_ADMISSION_SCORECARD_RATCHET=1`): PASS=200, DIVERGE=0, FALLBACK=1, SKIP=5, ERROR=0, total=206. This matches the frozen 200/1/5/0 ratchet exactly. The scorecard did not move. Two runs produced the same counts.
- Canonical fixture digests (`TestDiagnosticParserCoreCanonicalAdmissions`): pass for all four fixtures — rewrite, query_compile, language, grammargen_lr. Compact and production trees stay byte-identical.

## Performance

This change carries no performance claim. The retain gate here is correctness only. The local noise floor for this environment runs 7.4 to 10.8 percent. No benchmark number in this environment can support a decision.

## Why the scorecard did not move

The three fixed returns sit on one path inside `condenseWithOutcomeAtomic`. Reaching that path needs a live incumbent at the boundary. The historical-provenance branch always resets the incumbent to zero first. A live incumbent and a historical predecessor cannot occur together in one call today.

I confirmed this three ways: a static control-flow proof, an empirical probe test, and a revert-and-rerun check. The revert-and-rerun check restored the pre-fix bare literals and re-ran the full package suite. The suite, including the new fold-path tests, still passed.

This fix removes a real code-level asymmetry. It also guards against a future change that could make the combination reachable. It changes no observable behavior today. The new test `TestCondenseWithOutcomeAtomicProvenanceRatchet` catches a real regression. It parses the function's AST. It fails when a future edit reintroduces a bare literal outside `buildOutcome`.

## Freeze

freeze-held: engine provenance fix, do not merge before 2026-08-05.

### DIFF
--- a/internal/parsercorephase0/condense_outcome_provenance_ratchet_test.go
+++ b/internal/parsercorephase0/condense_outcome_provenance_ratchet_test.go
@@ -1,0 +1,128 @@
+package parsercorephase0
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestCondenseWithOutcomeAtomicProvenanceRatchet is a structural regression
+// test for the historical-provenance-propagation fix in
+// condenseWithOutcomeAtomic (core.go). It parses the package's own source
+// and asserts that condenseWithOutcomeAtomic constructs every non-empty
+// condenseOutcome result through the buildOutcome closure, never through a
+// second, ad hoc condenseOutcome{...} literal that could silently omit the
+// historicalBoundarySplit / historicalConvergedSplit /
+// historicalForestDeterministic / historicalCleanPathRank /
+// historicalLineage fields buildOutcome carries.
+//
+// This exists because the defect this fix addresses is not observable
+// behaviorally through any input reachable via the public Core API: a live
+// incumbent (needed to reach the shallow-fold early returns this fix
+// touches) and a historical predecessor are mutually exclusive within one
+// condenseWithOutcomeAtomic call, since the historical branch
+// unconditionally zeroes oldID before the duplicate/shallow-fold logic ever
+// runs. link_fold_provenance_test.go's TestFold*PropagatesProvenance tests
+// confirm the fold paths do not fabricate history, but (confirmed by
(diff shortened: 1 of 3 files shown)


odvcencio (commented):

PR #643, opened 2026-08-02
add(parsercorephase0): add inert acceptance election port with ratchets

odvcencio:
## What this changes

This adds a library and no shipped behavior. The compact acceptance election
keeps the fail-closed decline it has today. Every route, digest, and sweep
number is unchanged.

`internal/parsercorephase0/acceptance_election.go` ports the reference
runtime's tied-tree selection order: `ts_parser__select_tree`
(parser.c:836-879) and `ts_subtree_compare` (subtree.c:591-618), tree-sitter
v0.25.0. The order has six rungs: error cost twice, dynamic precedence twice,
a take-the-later rule for equal nonzero error cost, and a recursive structural
comparison that orders by lower symbol id, then by fewer children, then by
children pairwise.

The file is inert. `TestAcceptanceElectionNoOutsideCallSitesRatchet` fails if
any non-test source in the package references it.

## Why it is inert

The plan was to replace the tied-accept decline with this order, so a tie
would resolve to the tree the reference runtime keeps instead of handing the
input back to the production parser. That was built and measured against the
locked C oracle. It published a different tree from the oracle on 8 of the 10
tied accepts it newly admitted.

Per-witness result, elected pick against the C oracle:

| language | witness | elected pick | C oracle | verdict |
|---|---|---|---|---|
| apex | class_literal_alias | class_literal (268) | field_access (270) with an "object" field | diverges |
| perl | print_filehandle_list_material_election | list_expression outer | ambiguous_function_call_expression outer | diverges |
| perl | unshift_two_args | list_expression outer | ambiguous_function_call_expression outer | diverges |
| perl | join_bare | list_expression outer | ambiguous_function_call_expression outer | diverges |
| ada | bare_aggregate_assignment_material_election | record_aggregate with a "name" field | record_aggregate with no field | diverges |
| ada | record_aggregate_named | record_aggregate with a "name" field | record_aggregate with no field | diverges |
| ada | discriminated_record | component_definition, 1 child | component_definition, 2 children | diverges (production defect, newly surfaced) |
| kotlin | annotated_declaration | source_file, 2 children | source_file, 1 child | diverges |
| kotlin | 2 further object_declaration witnesses | matches the oracle | matches the oracle | agrees, production diverges |

Sweep totals with the order wired in, against the shipped numbers:

| language | shipped | wired | change |
|---|---|---|---|
| apex | 21 accept / 4 decline / 0 diverge | 22 / 3 / 1 | +1 accept, +1 divergence |
| perl | 16 / 4 / 0 | 19 / 1 / 3 | +3 accepts, +3 divergences |
| ada | 20 / 3 / 0 | 23 / 0 / 3 | +3 accepts, +3 divergences |
| kotlin | 10 / 4 / 0 | 13 / 1 / 1 | +3 accepts, 2 agree, 1 divergence |
| python | 29 / 0 / 0 | 29 / 0 / 0 | unchanged |

The cause is not the port. Take apex. The two live derivations are
`class_literal` (symbol 268) and `field_access` (symbol 270). The structural
rung orders by lower symbol id, so it selects `class_literal`, and no ordering
of that pair selects `field_access`. The oracle's own tree reads the text
"class" as an identifier token, which the `class_literal` derivation cannot
do. The reference runtime never built that alternative.

Perl states the same thing from the other side: the oracle keeps the
alternative the structural rung ranks second there too.

So the compact core's live-derivation set at a tied accept is not the
reference runtime's live-version set. The compact scheduler re-lexes and forks
per header, and its condense step defers precedence ties instead of collapsing
them, so it can carry alternatives the reference runtime never creates.
Ranking a set that contains such an alternative cannot recover the reference
answer, whatever the ranking.

The prerequisite for using this order is a live-derivation set proven equal to
the reference runtime's live-version set. That is a scheduler and lexer
question, not a selection question.

## What lands

- `acceptance_election.go`: the ported order, with the fold order rebuilt from
  the compact core's own fork stamps, an iterative structural comparison that
  matches the reference traversal, and an error-cost model over derivation
  payload roots.
- `acceptance_election_test.go`: one case per ladder rung, rung-precedence
  cases, structural comparison axes, fold stability, and a randomized property
  check that the elected candidate never depends on memoization.
- `acceptance_election_ratchet_test.go`: the inertness guardrail.
- `recovery_cost_ratchet_test.go`: authorizes `acceptance_election.go` as the
  one caller of the cost model. That caller is itself inert, so the
  clean-path-zero-cost property the earlier ratchet established is unchanged.

## Verification

- `go test .` green. `go test ./grammars` green.
- `go test -race ./internal/parsercorephase0` green.
- Admission scorecard: PASS=198 DIVERGE=0 FALLBACK=3 SKIP=5 total=206, equal
  to the shipped ratchet.
- W5 editor latency gates green.
- All five certification sweeps against the locked C oracle report the shipped
  numbers, with zero divergences.

### DIFF
--- a/internal/parsercorephase0/acceptance_election.go
+++ b/internal/parsercorephase0/acceptance_election.go
@@ -1,0 +1,478 @@
+package parsercorephase0
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ---------------------------------------------------------------------------
+// acceptance_election.go -- the reference runtime's tied-tree selection order,
+// ported for the compact core.
+//
+// THE PINNED C ORACLE IS THE EXECUTABLE SPECIFICATION for this file
+// (decision 0007, hypha://m31labs/gotreesitter): tree-sitter v0.25.0's
+// ts_parser__select_tree (parser.c:836-879) and ts_subtree_compare
+// (subtree.c:591-618). The production Go port of the same order lives in the
+// root package -- stackCompareForResultSelectionWithRawShape
+// (parser_result.go) at result granularity, and reduceForkWindowPreference /
+// compareRawStackEntriesRec (parser_reduce.go) at fork granularity. Every
+// rung, constant, and evaluation order below mirrors the C functions; a
+// divergence gets a receipt, not a silent "improvement".
+//
+// ts_parser__select_tree(left, right) answers one question: does right
+// replace left? Its rungs, in order:
+//
+//	1. right error cost  <  left error cost   -> take right
+//	2. left  error cost  <  right error cost  -> keep left
+//	3. right dyn. prec.  >  left dyn. prec.   -> take right
+//	4. left  dyn. prec.  >  right dyn. prec.  -> keep left
+//	5. left error cost   >  0                 -> take right
+//	6. ts_subtree_compare(left, right) < 0    -> keep left
+//	   ts_subtree_compare(left, right) > 0    -> take right
+//	   ts_subtree_compare(left, right) == 0   -> keep left
+//
+// Rung 5 is not a typo and is not dead: when two candidates carry the SAME
+// nonzero error cost and the same dynamic precedence, C takes the later
+// candidate. Production's port carries the identical rung with the identical
+// note (parser_result.go, `else if ac > 0`).
+//
+// ---------------------------------------------------------------------------
+// THIS FILE IS INERT. NOTHING IN THIS PACKAGE, AND NOTHING IN THE ROOT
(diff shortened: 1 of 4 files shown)


odvcencio:
Not landing. The selection port is complete and its behavior is measured: on tied acceptances it picks a different tree than the reference runtime on eight witnesses across four languages. The cause is upstream of selection — the candidate set at a tied acceptance contains alternatives the reference runtime never builds, so no ordering of that set can reproduce the reference tree. The branch is preserved with the port, its per-rung tests, and the measurements for the follow-on work, which targets candidate-set equivalence in the scheduler and lexer instead.

PR #644, opened 2026-08-02
add(c4): Add C4 bytecode corridor stage 2 implementation

odvcencio:
## Summary

Introduce stage two of the C4 bytecode corridor. The compiler projects a validated instruction stream directly from the decoded language tables. The scheduler routes single-header passes through this compiled lane and falls back to the generic dispatcher on any boundary. A standalone table-shape analyzer pins static census numbers. Rigorous equivalence tests prove byte-identical results against the baseline path.

## Changes

- Implement `parsercore_c4_program.go` as the bytecode compiler. It consumes decoded tables, classifies every populated cell, interns identical bodies, and emits a 32-bit word stream. A build-time validator checks all offsets and ensures cell exhaustiveness before the interpreter runs.
- Integrate the corridor VM into `parsercore_phase0_driver.go`. The scheduler loads the memoized program per language and shares the existing action-row slice. It sends eligible single-header passes to `dispatchCorridor()`. Corridor exits return control to the generic pass unchanged.
- Add `Cmd c4tablestats` and `parsercore_c4_analyzer.go` to measure static table properties. The tool reports cell disposition shares, key-count percentiles, unary-chain depths, and memory footprints. It writes a pinned JSON receipt and supports a `-check` flag for continuous integration guards.
- Ship `parsercore_c4_equivalence_internal_test.go` to enforce parity obligations R1 through R6. The test runs every canonical fixture with the corridor off and on. It asserts identical tree digests, root shapes, work vectors, and fork-boundary state. Pass-mix counters record coverage fractions per fixture.
- Extend `DiagnosticParserCoreGenericWork` with `SingleHeaderPasses` and `CorridorPasses`. These counters track corridor execution volume. They stay at zero when the lane remains disabled.
- Add `ClassifyBoundaryWithRow` to `internal/parsercorephase0/core.go`. This method accepts a pre-resolved action row. It removes duplicate table lookups during corridor classification.
- Update expected work counts in scheduler and child-work tests. The assertions now include the new pass counters.

## Testing

- Run parity tests inside Docker for one language: `bash cgo_harness/docker/run_single_grammar_parity.sh go`
- Validate the pinned table-shape receipt: `go run ./cmd/c4tablestats -check testdata/c4_table_shape.json grammars/grammar_blobs/*.bin`
- Toggle the lane and run equivalence tests: `GTS_C4_CORRIDOR=1 go test . -run 'TestParserCoreCorridor' -count=1`


odvcencio:
> **SUPERSEDED IN PART.** The `uniform_sole_reduce_states = 0` row below is an analyzer artifact, and the fusion ban this comment argues from it is WITHDRAWN. The `edge_static_unary_reduce_share` and `reduce_shift_candidate_share` rows below also use the wrong denominator. See the corrected receipt in the later review-response comment on this PR.

## Stage-2 receipts (spec.c4-bytecode-isa.v1)

### R6 pass mix, per canonical fixture (Go grammar, gts_workcount board, schema `gts-work-count-parsercore-child/v5`)

| fixture | passes | single-header passes | single-header share | corridor passes | corridor share of passes | corridor share of single-header |
|---|---:|---:|---:|---:|---:|---:|
| rewrite | 2,600 | 1,567 | 60.3% | 1,420 | 54.6% | 90.6% |
| query_compile | 13,001 | 8,001 | 61.5% | 7,384 | 56.8% | 92.3% |
| language | 12,854 | 7,473 | 58.1% | 6,986 | 54.3% | 93.5% |
| grammargen_lr | 129,273 | 76,100 | 58.9% | 69,431 | 53.7% | 91.2% |

The spec's pre-C0 profile put the single-header share at 60.6-64.7% and required stage 2 to re-measure it on the C0 board. The measured share is **58.1-61.5%**. The corridor covers **90.6-93.5%** of that shape; the residual is the no-action, conflict, and unsupported cells that hand back by design.

### Analyzer pinning (spec section 2), receipt `testdata/c4_table_shape.json`

Every static table-shape number the brief marked an estimate is now measured.

| row | go | javascript | json |
|---|---:|---:|---:|
| states | 1,511 | 1,870 | 32 |
| populated (state, terminal) cells | 29,407 | 52,993 | 224 |
| sole-action cell share | 99.91% | 99.01% | 95.09% |
| corridor cell share | 99.91% | 98.78% | 89.29% |
| EXPECT key count p50 / p90 / max | 8 / 51 / 53 | 36 / 60 / 125 | 4 / 12 / 14 |
| unary (k=1) share of reduce cells | 11.45% | 30.24% | 10.62% |
| unary-chain edges, depth >= 2 | 23.15% | 51.42% | 1.49% |
| edge-static unary reduce cells | 20.60% | 89.11% | 8.33% |
| uniform-sole-reduce states | 0 | 0 | 0 |
| REDUCE_SHIFT candidate share | 9.81% | 34.52% | 0.00% |
| program footprint (bytes) | 250,536 | 447,192 | 1,964 |
| EXPECT EXT states | 729 | 1870 | 0 |
| interned cell bodies | 1142 | 1711 | 38 |
| dense / sparse blocks | 318 / 1193 | 59 / 1811 | 16 / 16 |

Two estimates in the brief are **refuted by measurement**, and one holds:

1. Section 3.3 estimated 40-55% of reductions are k=1. Go measures **11.45%**, JSON 10.62%, JavaScript 30.24%. No admitted grammar reaches the estimated band. `REDUCE_CHAIN`'s stated grounding does not hold for the stage-2 grammar.
2. Section 3.3's `REDUCE_SHIFT` grounding names the "uniform-sole-reduce states" analyzer row. That row measures **0 states** on all three grammars. The edge-walked REDUCE_SHIFT candidate share is 9.81% on Go, 34.52% on JavaScript, 0% on JSON, so the candidate mass is real on JavaScript but not on the stage-2 grammar.
3. Section 3.6 estimated a 100-500 KiB footprint for a Go-class grammar and the estimate **holds**: 245 KiB for Go, 437 KiB for JavaScript, 1.9 KiB for JSON.

Per the spec's own fusion admission rule, stage 2 therefore emits **no** superinstruction. `REDUCE_CHAIN` and `REDUCE_SHIFT` keep reserved encodings only.

### Equivalence obligations

| obligation | status | evidence |
|---|---|---|
| S1 one source | green | `CompileParserCoreCorridorProgram` is the only producer; `TestParserCoreCorridorCompilerIsDeterministic` proves two compiles are byte-identical on all three grammars |
| S2 decode-back per state | green | `TestParserCoreCorridorDecodeBackPerState` walks every state, requires exact key-set equality, descriptor equality against `buildParserCoreLanguageTables`, constituent-operand reconstruction, and action-row-index equality |
| S3 stream validation once at build | green | `validate()` runs inside the compiler; `TestParserCoreCorridorStreamValidation` proves four corruption classes are rejected |
| contract 4.3 exhaustiveness, no default arm | green | `TestParserCoreCorridorCellExhaustiveness` walks every (state, terminal) cell of go, javascript, and json; the switch has no default arm and the disposition partition must sum to the walked cell count |
| R1 work-count board byte-identical | green | the full `gts_workcount` child board (core work, scheduler work, board-direct hooks, selected census, raw selected census) is byte-identical corridor-on vs corridor-off on all four fixtures |
| R2 deep-tree digests byte-identical | green | same run; digests also match the pinned admission digests |
| R3 allocation flat | green | 208 B/op, 6 allocs/op in both benchmark lanes |
| R4 fork-boundary handoff | green | `branchOrder`, `nextSeq`, `dispatches`, `tokens`, `electionIndex`, and the full work vector are identical after hundreds to thousands of exits and re-entries per fixture |
| R5 EXT-clear soundness | not exercised | stage 2 deliberately does not skip the capture+intern rounds. Every corridor election runs the full protocol through the existing `elect`, so no capture is ever skipped. The EXT flag is compiled and its CHECKPOINT decode form is proven exact per state; the skip itself is deferred |
| R6 pass-mix counter | green | `SingleHeaderPasses` published on the board, table above |

### Measurement (spec section 8 item 2f, section 7 kill thresholds)

Local proxy only. The measurement host is a shared multi-tenant workstation with concurrent agents; the noise floor on the scheduler-only lane is roughly plus or minus 10%, so this is not a gate-grade number.

Order-balanced A/B, `BenchmarkDiagnosticParserCoreCorridorSchedulerOnly` against `BenchmarkDiagnosticParserCoreWarmSchedulerOnlyQueryCompile`, single CPU, pinned core, 8 alternating cycles at 50 iterations:

- minimum of samples: corridor 11.07 ms, generic 11.17 ms — corridor 0.9% faster
- median of per-cycle ratios: corridor 1.4% **slower**
- allocation and retained-heap footprint identical

**The measured stage-2 model does not yet support capture.** The corridor deletes the dispatch triage scaffolding and the outer run-loop bookkeeping, and it deletes the second table walk per pass, but it still routes every election through `elect` verbatim and every mutation through the existing `apply*` seams. Those two are where section 3.4 located the cost (elections at 2,407-2,609 ns are the most expensive event class), and stage 2 does not touch them.

Against section 7's thresholds: the program-specific bar is that stage 3 does not start unless the stage-2 measured model supports at least 4% realistic incremental capture. **This receipt does not clear that bar.** The go/no-go decision belongs to the campaign lead, with two observations:

1. The corridor's coverage is high (90.6-93.5% of single-header passes, 53.7-56.8% of all passes), so the lane is not coverage-limited. The remaining cost sits inside `elect` and the `apply*` seams, both of which the corridor currently calls unchanged.
2. The superinstruction path that was meant to supply the rest of the capture is now measured, and its two justifications do not hold for Go.

A gate-grade re-measurement needs the dedicated quiet cloud VM and the sealed-epoch protocol before any of this is treated as a decision-grade number.

### Staging

The lane is off by default. `GTS_C4_CORRIDOR=1` (or `SetParserCoreCorridorEnabledForTest`) turns it on. `TestParserCoreCorridorDefaultOff` proves an unmodified build routes exactly as before. A grammar whose compile hits an unsupported construct keeps the generic lane, fail-closed, with no allowlist.

### Test lanes

- `go test ./` — green
- `go test -tags gts_parsercorephase0 ./` — failure set is byte-identical to `origin/main` (14 pre-existing failures, none introduced)
- `go test -tags 'gts_parsercorephase0 gts_workcount' ./` plus the four work-count children, corridor on and off — green and identical

odvcencio:
## Review response — corrected receipt and restated conclusion

Rebased onto `origin/main` at `8070e1ad`. Branch HEAD is now `4b3d28ea`.

### BLOCKING — analyzer definition (fixed)

Confirmed. `uniform_sole_reduce_states` broke on any non-reduce key, and extra-shift keys counted as breaking, so comment and whitespace tokens guaranteed the zero. The analyzer now carries both definitions with explicit names and doc comments that say which is which:

- `uniform_sole_reduce_states` — the strict row, retained for continuity, documented as near-zero by construction and near-useless for reduce uniformity.
- `uniform_sole_reduce_states_extra_neutral` — extra-shift keys excluded, because an extra shift compiles to `SHIFT_EXTRA`, which spec section 3.2 says stays inside the same election.
- `uniform_sole_reduce_covered_cells` / `_share` — the reduce mass sitting inside an extra-neutral uniform state, which is the quantity a fusion argument needs.

My numbers reproduce yours exactly: go **777 states / 89.2%** of reduce cells, javascript **366 / 50.0%**, json **12 / 94.7%**.

### MAJOR 1 — k=1 filter (fixed)

Confirmed. `EdgeStaticUnaryReduceCells` and `ReduceShiftCandidateCells` were computed inside the `ChildCount != 1` filter while their doc comments claimed all sole reduces. Edge uniqueness and the REDUCE_SHIFT probe are properties of the goto edge and have nothing to do with child count. Both now run over every sole-reduce cell:

- `edge_static_reduce_cells` / `_share` — new row, over all sole reduces, share over `ReduceCells`. This is the row that decides `corridorGotoModeStatic`.
- `edge_static_unary_reduce_cells` / `_share` — kept, now explicitly documented as the unary-chain-step question, share over `UnaryReduceCells`.
- `reduce_shift_candidate_share` — denominator corrected to `ReduceCells`.

Only the unary-chain rows still filter on `ChildCount == 1`, which is correct for what they measure.

### Corrected receipt (`testdata/c4_table_shape.json`)

| row | go | javascript | json |
|---|---:|---:|---:|
| states | 1,511 | 1,870 | 32 |
| populated (state, terminal) cells | 29,407 | 52,993 | 224 |
| sole-action cell share | 99.91% | 99.01% | 95.09% |
| corridor cell share | 99.91% | 98.78% | 89.29% |
| EXPECT key count p50 / p90 / max | 8 / 51 / 53 | 36 / 60 / 125 | 4 / 12 / 14 |
| reduce cells | 17,892 | 24,697 | 113 |
| unary (k=1) share of reduce cells | 11.45% | 30.24% | 10.62% |
| unary-chain edges, depth >= 2 | 23.15% | 51.42% | 1.49% |
| **edge-static reduce cells (all sole reduces)** | 90.81% | 95.81% | 89.38% |
| edge-static unary reduce cells | 20.60% | 89.11% | 8.33% |
| uniform-sole-reduce states (strict, artifact) | 0 | 0 | 0 |
| **uniform-sole-reduce states (extra-neutral)** | 777 (51.4%) | 366 (19.6%) | 12 (37.5%) |
| **reduce cells covered by a uniform state** | 89.16% | 50.02% | 94.69% |
| REDUCE_SHIFT candidate share (of reduce cells) | 1.12% | 10.53% | 0.00% |
| program footprint (bytes) | 250,536 | 447,192 | 1,964 |
| EXPECT EXT states | 729 | 1870 | 0 |

### Restated conclusion — the fusion ban is withdrawn

My earlier claim that stage 2's measurements refute the fusion grounding was **wrong on its central point**, and I retract it. Correct reading:

1. **Fusion is admissible.** Under the definition consistent with the corridor's own ISA, 51.4% of Go states are uniform-sole-reduce, covering 89.2% of reduce cells. That is a large candidate mass, not zero.
2. **What does not support `REDUCE_SHIFT` is its own candidate row**, measured over all reduce cells: go **1.12%**, json **0.00%**, javascript 10.53%. A reduce whose post-goto row for the same lookahead is a sole shift is rare on Go. Any fusion case must argue from the uniform row, not from this one.
3. **`REDUCE_CHAIN`'s k=1 grounding still does not hold for Go**: 11.45% of reduce cells are k=1 against an estimated 40-55%. That claim survives.
4. **`corridorGotoModeStatic`'s precondition is met and the old receipt understated it 4x on Go**: edge-static over all sole reduces is **90.81%** (go), 95.81% (javascript), 89.38% (json). Stage 2 still emits only `corridorGotoModeIndexed`, but the reason is no longer "the proof is missing" — the proof is in the receipt. It is now a scope decision, and fusing the static target is the obvious first stage-3 experiment.

### MAJOR 2 — TableLookupsProxy (fixed, with a regression test)

Confirmed and fixed by recording the event rather than declaring the deletion. The corridor does perform one table lookup per executed pass — it reads its own validated dispatch table instead of walking the parse table — so `corridorClassify` now calls `workCountRecordTableLookup()`. The hook compiles to an empty function outside `gts_workcount`, so the default build pays nothing.

New test `TestParserCoreCorridorTableLookupProxyIdentity` (`parsercore_c4_tablelookup_internal_test.go`, needs both tags) measures the counter directly rather than through the child board, so the gate is no longer structurally blind to it:

```
corridor=false table_lookups_proxy=41393 corridor_passes=0
corridor=true  table_lookups_proxy=41393 corridor_passes=7384
```

### MINORS (all fixed)

| item | fix |
|---|---|
| block-header sniffing accepted a target one word into a header | `validateBlockTarget` resolves the offset through the state-block index binary search, so only compiled block starts pass. Added the corruption case `SHIFT target points inside a block header` to `TestParserCoreCorridorStreamValidation`; it fails without the fix |
| decode-back test used the compiler's own `actionIndex` as oracle on both sides | switched to `parser.lookupActionIndex`, and the test now asserts the two agree per cell, so a first-wins / last-wins divergence fails the test instead of hiding |
| `SingleHeaderPasses++` and `corridorEligible()` framing | corrected in the code comment: it states plainly that `corridorEligible` runs once per run-loop iteration, on every parse whose language compiled, including multi-header iterations it can never take, and that it is not hoisted and not free |
| SHIFT_EXTRA target is 25 bits while the guard checked 26 | `corridorMaxBlockOffset` lowered to `1 << 25`, the narrowest operand field, with a comment saying why the guard uses the narrowest and not the widest. All 206 blobs still compile under it |
| `corridorMaxOperandUint26` declared but unwired | deleted |
| `validateBody` skipped `validateRowIndex` for FORK | FORK now validates its row index; corruption case added |
| `Census()` returned an unexported type | type exported as `ParserCoreCorridorCensus` |

### Fleet note, recorded

Re-measured after the 25-bit ceiling change: **206/206 blobs compile, 144.79 MiB retained total, 61 of 206 exceed the spec's 100-500 KiB range** (nim 6.51 MiB, verilog 6.07, sql 5.56, cobol 5.03, tlaplus 4.97). Obligation R3 still has no fleet receipt. The receipt is one command away — `cmd/c4tablestats` already accepts any blob list, so `go run ./cmd/c4tablestats -out <path> grammars/grammar_blobs/*.bin` produces it. Zero cost while the flag is off; this must be settled before stage 3 flips the default.

### Not addressed

Nothing from the review is left open. The one obligation still not exercised is R5, unchanged from the first receipt: stage 2 deliberately never skips the capture and intern rounds, so no EXT-clear soundness shadow is needed yet.

### Test lanes after the fixes

- `go test ./` — green
- `go test -tags gts_parsercorephase0 ./` — 14 failures, byte-identical set to `origin/main` at `8070e1ad`, verified by diffing full runs
- work-count child board, corridor on and off, four fixtures — identical
- all corridor tests including the two new corruption cases — green

PR #650, opened 2026-08-03
Retire the Go EOF-newline compat member and cover the included-ranges route

odvcencio:
## What this changes

Two things, after a reviewer caught a defect in the first revision of this
branch.

1. Retire one member of the Go result-compatibility arm: the trailing
   end-of-file newline span.
2. Add the first committed test coverage for the included-ranges route, for
   any language.

## The defect the reviewer caught, and the correction

The first revision also deleted `normalizeGoSourceFileRoot` and its top-level
recovery re-parse subsystem. That deletion was wrong and is fully reverted in
commit `f0fb9f6b`.

The census behind it covered three routes: fresh, over-64 KiB, and
incremental. It recorded zero rewrites on 22,613 parses, and a reviewer
independently reproduced that result. The member is nevertheless live on a
fourth route the census never measured: **included ranges**.

Reproduced on this branch before the revert, against the locked C oracle:

| | root kind | span | HasError |
|---|---|---|---|
| main | `source_file` | `[0,767)` | false |
| first revision | `ERROR` | `[0,378)` | true |
| C oracle | `source_file` | `[26,476)` | true |

The deletion traded one C divergence for another. It was neither a no-op nor
an improvement.

`normalizeGoSourceFileRoot` returns early unless the root is `ERROR`, and
`recoverGoRootTopLevelChunks` replaces children only after
`goChildrenLookLikeTopLevel` passes. A nonzero rewrite receipt therefore
proves the root was `ERROR` and that the member repaired it.

Production reaches this route through injection. `injection.go` calls
`childParser.SetIncludedRanges` for every injected child, and
`grammars/markdown_injection_register.go` maps the `go` and `golang` fence
languages to the Go grammar. A Markdown document with two or more Go fences
is exactly this shape.

## The route had no test, for any language

`Parser.SetIncludedRanges` had no committed test anywhere. That is why a full
green suite could not see the regression. This PR adds three gates:

- `TestIncludedRangesGoRootStaysSourceFile` — pins the root symbol.
- `TestIncludedRangesGoArmMemberIsLive` — the positive control. It asserts the
  arm member still rewrites the tree on this route, so the gate above cannot
  pass by accident.
- `TestIncludedRangesGoRootParity` — C-oracle comparison. It pins the root
  symbol, span, and child count, and pins the known `HasError` divergence so a
  change surfaces instead of passing silently.

All three fail when the member is stubbed out, with
`root type = "ERROR", want "source_file"`. The fixture is committed at
`testdata/included_ranges/go_two_fences.go`; no test depends on a host path.

## What still retires

The trailing end-of-file newline member. `extendNodeToTrailingWhitespace`
runs unconditionally after the compatibility pass and accepts a superset of
the byte set the Go member tested, so on that ASCII-only domain the Go member
is strictly subsumed. A census recorded 35,244 gate entries and zero rewrites.

The reviewer notes the retained Scala arm of `normalizeRootEOFNewlineSpan` is
dead by the identical argument. That is a follow-up, not this PR.

## Also in this PR

All three Go arm members are now registered as named census subpasses, so a
receipt names the member that rewrote the tree instead of only the arm. A
probe in `parser_result_test/materialization_subpass_census_test.go` pins the
three ids and one rewrite receipt.

Census receipts record on the production route only. With the default
candidate route, `ParseRuntime().NormalizationPasses` is nil and nothing is
recorded. The evidence notes say so explicitly.

## Corrections to the record

`CHANGELOG.md` and the registry `retirement_condition` previously said the
census "recorded zero tree rewrites". That was true for three routes and
false for the fourth. Both now name the route set, and the registry
retirement condition requires receipts on five routes plus the C oracle.

`TestNormalizeGoSourceFileRootRetagsRecoveredTopLevelChildren` is restored.

## Gates

Full root package suite, `./grammars`, `./parser_result_test/...`, race on the
touched paths, C parity smoke suite, Go corpus C parity, `TestGoNewMakeResidualParity`,
and the new included-ranges parity gate: all green. Four pinned canonical Go
deep-tree digests byte-identical. Admission scorecard strict and ratcheted:
PASS=198 DIVERGE=0 FALLBACK=3 SKIP=5, exactly the frozen numbers.

No performance claim is made.

## Known pre-existing failure

`TestCanonicalGoBenchmarkPreflight` fails on `query_compile` with
`max stacks=0 want >=8`, identically on a pristine `origin/main` worktree.


odvcencio:
Correction pushed as `f0fb9f6b`. The reviewer is right and I have reverted the `normalizeGoSourceFileRoot` deletion in full.

I reproduced the defect on this branch before reverting: with the deletion, the included-ranges root came out `ERROR` where C publishes `source_file`. My census covered three routes and missed the fourth. The member is live there.

The route now has committed coverage, including a C-oracle comparison and a positive control that proves the member still rewrites the tree. All three new gates fail when the member is stubbed out, so the route can no longer go invisible.

What still ships is the EOF-newline member only, on the subsumption argument the reviewer supplied, which is stronger than my original census reason.

I have corrected CHANGELOG.md and the registry `retirement_condition` to name the route set instead of claiming an unqualified zero, restored the deleted unit test, and added the subpass probe. The Scala arm of `normalizeRootEOFNewlineSpan` looks dead by the same subsumption argument; I have left it for a follow-up rather than widening this PR.

odvcencio:
Both fixes pushed as `6f594ebc`.

**Fix 1.** You were right that the geometry pinned the one shape where Go and C agree, and that the comment read as a parity claim. I re-measured on the same fixture and oracle and reproduced your table exactly. I also found a fifth geometry that is worse than any in your list: `{{40,150},{203,260}}` publishes an `ERROR` root **with the member restored**, where C publishes `source_file`. So the member cannot be credited with root repair in general on this route.

The C test is now a four-geometry pin table. Nothing asserts parity. Each row pins the measured root of both parsers as an exact snapshot, carries a note on what it proves and what it does not, and fails on any movement in either direction:

| geometry | C | GTS | arm |
|---|---|---|---|
| `{26,150},{203,276}` interior anchors | `source_file [26,276) ch=7` | `source_file [0,276) ch=7` | fires 96 |
| `{0,150},{203,276}` anchored | `source_file [0,276) ch=7` | same | fires 96 |
| `{0,150},{203,250}` trimmed tail | `source_file [0,250) ch=6` | `source_file [0,276) ch=7` | fires 96 |
| `{40,150},{203,260}` padding | `source_file [40,260) ch=5` | `ERROR [0,153) ch=4` | does not fire |

The anchored row is kept and labelled as an artifact of the anchors, not a property of the route. The padding row is labelled as the open clipping defect, with a note that the Go member should be re-censused when that lane lands.

The deletion guard is now a separate test so its intent cannot be misread. The non-cgo fixture moved to interior anchors, with a comment saying why byte 0 is unrealistic.

**Fix 2.** The unreachable `t.Skip` is now `t.Fatal`.

All four gates still fail when the member is stubbed, across all three firing geometries.

Gates re-run green: full root suite, `./grammars`, `./parser_result_test/...`, C parity fresh and incremental, Go corpus parity, new included-ranges gates, scorecard strict PASS=198 DIVERGE=0 FALLBACK=3 SKIP=5.

Thanks for the `bytesAreParserPadding` root cause and the 976-parse census — that explains my padding row, and I have referenced the defect in the test rather than papering over it. Noted the three queued follow-ups; not touching them here. Not merging.

odvcencio:
CI wiring fixed in `376366a6`. You were right — the gate would have merged dark.

**Which jobs execute the tests, verified by reading the lanes back and by `go test -list` against the exact regexes:**

| job | step | runs on this PR? |
|---|---|---|
| `parity-cgo` | CGo Parity Smoke Gate | yes — `run_code_ci` is true |
| `parity-cgo-exhaustive` | CGo Structural Parity Exhaustive (Fresh + No Errors + Issue3) | yes |

I added **both** new tests, not just the pin table: `^TestIncludedRangesGoRootParity$` and `^TestIncludedRangesGoArmGuardsRootSymbol$`. The second is the actual deletion guard, so leaving it out would have reproduced the same failure mode one layer down.

**You asked me to check the exhaustive lane rather than assume.** It does run: seven files in this PR match its trigger pattern (`parser_result_compat.go`, `parser_result_go.go`, `parser_result_go_compat_memory_budget_test.go`, `parser_result_javascript_typescript.go`, `parser_result_misc_spans.go`, `parser_result_root_build.go`, `parser_result_test/materialization_subpass_census_test.go`). Its four steps each carry their own `--run` regex and none covered the new tests, so I added them to the Fresh step as well. The tests are fixture-driven and ignore `GTS_PARITY_MODE`, so the duplicate run costs about a second and buys defence in depth on exactly the change class that triggers that lane.

**Verified, not assumed:**
- `--run` is threaded straight into `go test . -tags treesitter_c_parity -run '<regex>'` by `run_parity_in_docker.sh`.
- `go test -list` with the updated smoke regex returns 13 tests including both new ones; with the exhaustive-Fresh regex it returns both.
- Running the exact updated smoke selection locally passes in 104s.
- The workflow still parses as YAML (22 jobs).

Both edits append to the end of the existing alternation, so the conflict with the sibling Kotlin PR adding `^TestIncludedRangesKotlinRootParity$` should resolve by keeping both entries.

HEAD is `376366a67b4e1f94b854af3933fbaa725bb689ab`. Not merging.

PR #659, opened 2026-08-03
add(merge-census): count stack-merge events against the reference runtime

odvcencio:
Add a census that compares stack-merge events against the reference runtime.

Stage M0 of the merge-time-election lane. The lane's premise is that the
reference runtime collapses duplicate stack versions at the merge and carries
one version forward, while this parser refuses the merge when subtree shapes
differ, carries both versions, and elects at the end. Nobody had counted the
merges on both sides. This change counts them.

Zero shipped behaviour change. The whole instrument sits behind the new
`gts_merge_census` build tag, and `mergeCensusEnabled` is a false constant in
the default build, so the compiler removes every census block.

## Merge ratio, per language and aggregate

M_c is the reference runtime's collapsed pairs. M_p is production's. The
denominator is the 104 constructed sources the stage-D0 derivation-set
differential already measures, so the two censuses share one corpus.

| corpus | sources | M_c | M_p | M_p/M_c | sources where the reference merges and production does not |
|---|---|---|---|---|---|
| apex | 25 | 31 | 1 | 0.0323 | 5 |
| perl | 17 | 57 | 0 | 0.0000 | 7 |
| ada | 23 | 47 | 2 | 0.0426 | 5 |
| kotlin | 13 | 54 | 12 | 0.2222 | 3 |
| python | 26 | 2 | 0 | 0.0000 | 2 |
| **total** | **104** | **191** | **15** | **0.0785** | **22** |

Production collapses one version pair for every 12.7 the reference runtime
collapses. No source shows production merging more than the reference runtime,
so the over-merge floor is zero.

## Refusal breakdown

Production entered a merge front door 308 times and collapsed 15 pairs. The
293 refusals attribute exactly, with no residue:

| gate | count | reference runtime does the same? |
|---|---|---|
| candidate never packed into the graph | 142 | no; every reference version is a node chain from creation |
| score or shifted equality | 92 | no; the reference runtime merges across precedence differences |
| distinct materializing shapes | 59 | no; the reference runtime packs the difference as an extra link |
| top state or byte offset | 0 | yes |
| clean zero error | 0 | out of scope until stage M2 |
| recovery cost | 0 | yes |

All 142 unpacked-candidate refusals have exactly one flat side; none has two.
Only ada reaches the score gate (92 of 92). Only apex and ada reach the
shape refusal (53 and 6).

The Tier-2 link-payload test ran 41 times, accepted 29, and refused 12. In 10
of those 12 refusals the reference runtime's shallow test would have accepted,
so production's deep test turns a reference collapse into an append. All 10 are
kotlin. No comparison went the other way.

## The Foo[int](a) receipt

The census measures the witness against a bracket-free control (`_ = g(a)`),
because the counters are parse-global. `_ = a[b](c)` is not a control: it
carries the same conflict and produces identical counters.

At `_ = Foo[int](a)` the reference runtime adds, over the control, 3 alternate
link appends, 4 duplicate no-ops, 1 recursive change, and **zero precedence
replacements**. It publishes
`type_conversion_expression(generic_type(...))`.

The verdict is behaviour 2: the reference runtime builds the pair, packs both
links, and elects at the first spanning pop. It does not collapse the pair by
precedence, and it does not skip building it. The receipt also records that
duplicate no-ops fire in the same region, so some pairs there do collapse by
incumbency; the counters are parse-global and cannot say which node each one
landed on.

## Reference-side counting method

The parse logger cannot count merges. `parser.c` calls `ts_stack_merge` at four
sites and none of them logs. The only merge-adjacent line is `condense`, which
fires when any of removal, merge, or version swap changed the stack, so it
cannot be attributed. The live version count falls for removal and cap
enforcement too.

The census therefore reads the repository's own instrumented build. It applies
`cgo_harness/work_count/tree_sitter_v0_25_1.patch` to the pinned runtime source
resolved from the module cache, compiles a driver against it, and dlopens the
same grammar shared object the parity C-reference loader built. The patch
already carries `merge_attempts_proxy`, `merge_successes_proxy`, and the six
link-union outcome counters; no new C instrumentation was written. The driver
runs the patch's three counter self-models before any count is trusted.

## CI

New job `merge-event-census-cgo` runs both census tests. Every cgo parity lane
in `ci.yml` passes a fixed `--run` regex and a fixed tag string, so a
new-tagged cgo test runs nowhere unless a lane names it. This one names it, and
the job is listed in the required `build` gate's `needs` and `require_success`
checks.

## Gates

- `go test . -count=1`: pass.
- `go test . -count=1 -tags gts_merge_census`: pass. The suite is green with
  the census on and with it off, so the instrument changes no tree.
- `go test ./grammars -count=1`: pass.
- `go test ./parser_result_test/... -count=1`: pass.
- `go test -race .` over the GLR, merge, GSS and condense tests: pass.
- `GOMAXPROCS=1 go test -race ./internal/parsercorephase0`: pass.
- Admission scorecard ratchet lane, exact `-run` and env: pass.
- Five A3 certification sweeps: apex and ada pass; kotlin, perl and python skip
  for the absent `corpus_real` fixture, same as on the base commit.
- `go tool nm` on the default test binary: zero census symbols.
- benchstat, 8 runs each of three parse benchmarks: allocations identical,
  bytes per operation unchanged, times within noise.

## Pre-existing failures, not caused here

- `TestCanonicalGoBenchmarkPreflight` fails identically on unmodified `main`
  (`query_compile max stacks=0 want >=8`).
- The stage-D0 derivation-set census fails on unmodified `main` at 24 set
  differences against its pin of 32. It is a dark gate: no workflow passes
  `gts_derivation_set_census`. Reported in `docs/ci-gate-coverage.md` section
  3b and left for that lane's owner to re-pin with a receipt.
- `go vet -tags gts_no_parsercorephase0 ./...` fails identically on `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -543,6 +543,7 @@ jobs:
       - perf_scan_budget
       - exhaustive_parity_scope
       - compact-t3-oracle-cgo
+      - merge-event-census-cgo
       - parity-cgo-exhaustive
     if: always()
     runs-on: ubuntu-latest
@@ -564,6 +565,7 @@ jobs:
           EXHAUSTIVE_PARITY_SCOPE_RESULT: ${{ needs.exhaustive_parity_scope.result }}
           EXHAUSTIVE_PARITY_SCOPE: ${{ needs.exhaustive_parity_scope.outputs.run_exhaustive }}
           COMPACT_T3_ORACLE_CGO_RESULT: ${{ needs.compact-t3-oracle-cgo.result }}
+          MERGE_EVENT_CENSUS_CGO_RESULT: ${{ needs.merge-event-census-cgo.result }}
           EXHAUSTIVE_PARITY_RESULT: ${{ needs.parity-cgo-exhaustive.result }}
         run: |
           set -euo pipefail
@@ -600,6 +602,7 @@ jobs:
           require_success "parity_report" "$PARITY_REPORT_RESULT"
           require_success "perf_scan_budget" "$PERF_SCAN_BUDGET_RESULT"
           require_success "compact-t3-oracle-cgo" "$COMPACT_T3_ORACLE_CGO_RESULT"
+          require_success "merge-event-census-cgo" "$MERGE_EVENT_CENSUS_CGO_RESULT"
 
           if [ "$IS_DRAFT" = "true" ]; then
             require_success "draft_focused_tests" "$DRAFT_FOCUSED_RESULT"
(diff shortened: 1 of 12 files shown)


odvcencio:
Integrated into `codex/linkerscript-normalizer-retirement` at `596b9bd0`. The default merge path compiles identically. The Docker C/Go census passes on the current head. Closing this source PR as superseded.

Issue #660, opened 2026-08-04
v0.48.0 regression: name field stamped onto separator commas (import_from_statement, import_statement, subscript)


smm-h:
**Version:** v0.48.0 (regression — v0.47.1 is clean)
**Grammar:** Python

## Summary

In v0.48.0, the `name` field is stamped onto anonymous separator comma nodes in
`import_from_statement`, `import_statement`, and `subscript`. v0.47.1 does not have this
behavior, and the C runtime does not either — the change landed within the last few days of
releases.

## Reproduction

```python
from a import b, c
```

C runtime (and gotreesitter v0.47.1):

```
name:(dotted_name 14..15) (, 15..16) name:(dotted_name 17..18)
```

gotreesitter v0.48.0:

```
name:(dotted_name 14..15) name:(, 15..16) name:(dotted_name 17..18)
```

The comma between imported names is reported as carrying the `name` field.

## Scale

Found while benchmarking parser options for a static analysis tool over a real Python
corpus (rlsbl, ~561 files; a django checkout, ~2,927 files). This misattribution accounts
for 2,090 divergent field edges on the first corpus and 7,340 on the second — by far the
largest single divergence class vs the C runtime, ahead of the documented
expression-statement normalization.

Affected node/field combinations observed:

- `import_from_statement` — `name:` on `,`
- `import_statement` — `name:` on `,`
- `subscript` — `subscript:` on `,`

## Notes from bisecting

Checked v0.15.3, v0.20.9, v0.30.0, v0.40.0, v0.45.0, v0.47.1, v0.48.0 against the same
corpus: the comma field misattribution appears only in v0.48.0. Happy to provide the
comparison harness or per-file divergence inventories if useful.

odvcencio:
Resolved in emergency release [v0.48.1](https://github.com/odvcencio/gotreesitter/releases/tag/v0.48.1). The patch restores reference-compatible field names for anonymous separator nodes in Python import and subscript forms. It passed the focused regression suite, Python C parity, the full grammar tests, and all hosted pull request gates.

odvcencio:
also heya, thanks for filing this and for using gotreesitter! i really appreciate you reaching out!

PR #661, opened 2026-08-04
fix(parser): restore v10 admission safety and bound retry escalation

odvcencio:
## Summary

- Restore the 64 KiB production-route gate for candidate admission.
- Bound certified no-stack retry escalation to one 160-stack, 160-merge rung below 256 KiB.
- Preserve explicit diagnostic stack and merge caps.
- Skip redundant accepted-error retries for the certified C grammar blob.
- Add the GCP fleet runbook and bytecode stream specification index.

## Root cause

Commit `c1806563` removed the admission size gate. A pinned v10 Python replay then measured 48.24x C without the gate and 4.76x C with it. TypeScript's 3.5 MiB witness reached about 1.8 GiB without the gate and completed in 376 ms with it.

The retry changes prevent small, ambiguous inputs from escalating without a bounded pressure proof. They preserve explicit diagnostic caps and keep large inputs outside the exceptional retry rung.

Issue [#660](https://github.com/odvcencio/gotreesitter/issues/660) is a separate field-projection regression. The current main branch already contains its fix.

## Validation

- Docker admission contract test passes.
- Docker admission candidate test group passes.
- Docker certified retry and explicit-cap tests pass.
- Docker C grammar profile tests pass.
- Docker inherited-field projection test passes.

The full-fleet measurement remains a release prerequisite. GCP reauthentication is required before the remote run can start.

### DIFF
--- a/admission_switch.go
+++ b/admission_switch.go
@@ -91,12 +91,10 @@ func admissionCandidateEnvEnabled() bool {
 // admission switch applies to Parsers with no explicit override. A per-Parser
 // override still wins.
 //
-// Tranche B9 removed the source-length eligibility decline: every input,
-// regardless of size, is eligible to attempt the candidate route. An explicit
-// timeout, cancellation flag, or the scheduler's own memory-budget poll
-// (tranche B8) is honored on the candidate route itself, with a compatible
-// stop receipt, falling back to production when it trips; included ranges and
-// observability hooks still keep a parse on production.
+// Sources at or above parseRuntimeMemoryMinSourceBytes remain on production.
+// The compact scheduler cannot poll the automatic memory budget at the required
+// granularity. Small sources still honor timeout, cancellation, and memory
+// budget checks before the parser accepts the candidate result.
 func SetAdmissionCandidateRouteDefault(enabled bool) {
 	admissionCandidateRouteDefault.Store(enabled)
 }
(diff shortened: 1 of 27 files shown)


odvcencio:
## Perf evidence update

The repair branch now has controlled evidence for the v10 discrepancy.

- Base `603658c6`, Python, eight largest files: 49.25x Go/C by total, with five files above 10x.
- Gate-only control: 4.55x by total, with no cliffs.
- Repair branch `fe699265`: 4.93x by total, with no cliffs.
- The largest Python witnesses now stay on production. `test_logging.py` fell from 3.96 s on the ungated base to 155 ms on the repair branch. `test_socket.py` fell from 2.01 s to 155 ms.
- The 3.50 MiB TypeScript witness completed in 434 ms on the repair branch. It recorded zero candidate attempts and no fallback.
- The standard benchmark trio improved against the exact base: full parse -1.68%, one-byte edit -1.77%, no-edit -0.47%. Full-parse allocation volume fell 1.35%; allocation counts did not increase.

Focused Docker checks passed for the admission contract, retry caps, and certified C profile. The branch remains draft while the full fleet and retry-policy evidence continue.

odvcencio:
The tagged admission suite exposed one stale expectation from the earlier ungated experiment: it expected the 235 KiB grammargen_lr fixture to route through the compact engine. The test now expects the large fixture to stay on production while retaining direct compact-engine digest coverage. Focused Docker validation passes. Latest perf branch commit: `9afc0920`.

odvcencio:
Superseded by the reconciled V10 branch. It contains the bounded retry, C retry profile, and runbook changes. Newer compact-route controls replaced the 64 KiB gate. The accepted V10 epoch completed 1,435/1,435 cases without an out-of-memory event. Closing this older branch.

PR #708, opened 2026-08-12
refactor(ada): Refactor Ada tree repairs and retire obsolete association-choice subpass

odvcencio:
## Summary

Replace post-parse tree repairs with native grammar resolution for Ada aggregate conflicts. This branch adds a declared-conflict election policy that settles the record-vs-array tie before parsing, retires the unreachable association-choice materialization pass, and narrows the constraint-kind-election rewrite to prevent regressions on named associations.

## Changes

- Add a declared-conflict election policy for Ada's component_choice_list versus discrete_choice tie in grammars/runtime_profiles.go. The policy folds the ambiguous reduce row into discrete_choice before the GLR engine forks.
- Remove the dispatch.ada.association-choice-materialization subpass and all related node-construction helpers from parser_result_ada.go. The post-parse wrapper is now unreachable because the raw parse produces the correct tree.
- Restrict dispatch.ada.constraint-kind-election to gate its rewrite on single-positional-association shapes. Named associations no longer trigger the rewrite, which prevents a regression against the C oracle.
- Pass prefix and selector_name field identifiers into the attribute-reference rebuild chain. The repaired nodes now explicitly carry the fields that the C oracle assigns natively through the hidden _subtype_indication rule.
- Update Ada parity tests, A3 certification sweeps, and materialization census probes to reflect the native resolution and the removed subpasses.
- Refresh testdata/result_compat_ownership_v1.json to document the retirement condition and update the dispatch.ada group purpose.

## Testing

- Run `bash cgo_harness/docker/run_single_grammar_parity.sh ada` to verify full C-oracle parity.
- Run `bash scripts/run_randomized_benchmarks.sh --output /tmp/perf-bench.txt` with stable settings to confirm no performance regression.

### DIFF
--- a/admission_switch_a3_arm_noop_confirmation_test.go
+++ b/admission_switch_a3_arm_noop_confirmation_test.go
@@ -45,6 +45,20 @@ import (
 // Follow-up arm-deletion PRs for perl, python, and ada must not treat this
 // finding's retirement precondition as met on this evidence.
 //
+// UPDATE: ada's array_others_choice witness now confirms. A declared-conflict
+// election policy (grammars/runtime_profiles.go "ada",
+// ConflictPolicyDeclaredReduceReduceHighestSymbol in conflict_policy.go)
+// resolves the component_choice_list/discrete_choice tie natively before the
+// parser forks, so dispatch.ada.association-choice-materialization
+// (previously materialization-owned) retired outright and
+// dispatch.ada.aggregate-kind-election no longer reshapes this witness --
+// see TestA3ArmNoOpConfirmationAdaOthersChoiceConfirms.
+// locked_positional_array_aggregate's ambiguity
+// (record_component_association_list vs positional_array_aggregate) is a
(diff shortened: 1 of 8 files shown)


odvcencio:
I checked the current remote head `9a3af7ea602778656ee2f055800bbcd780471ee1`.

The focused Ada Docker evidence passes:

- A3 sweep: 20 accepted, 3 declined, 0 divergences.
- Ada constraint-election parity: pass.
- Ada raw C-oracle parity: pass for the new `array_others_choice` witness.
- The existing `locked_positional_array_aggregate` witness remains a known divergence and stays skipped.

I cannot accept this branch under campaign R1.2. The new Ada entry in `grammars/runtime_profiles.go` adds a blob-SHA keyed conflict policy. R1.2 freezes new blob-SHA, driver-local, and language-name admission exceptions.

Please replace this exception with parser-derived proof, or document an owner-approved change to the grant-freeze doctrine. The Ada behavioral evidence remains useful, but this PR is not accepted campaign evidence while the new exception remains.

PR #718, opened 2026-08-14
fix(yaml): fix single-quoted multiline scalar scanner termination

HarshK97:
## What does this PR do?

Fixes premature scanner termination when single-quoted flow scalars (`'[` ... `]'`) cross line breaks in YAML files. Multi-line single-quoted scalars now scan cleanly without popping the scanner indentation stack early or dropping document boundaries (`---`).

## Why this approach?

In `grammars/yaml_scanner.go` (line 1562), double-quoted string continuation (`yTokBRDqtStrCtn`) checks `(isBR || hasNwl)`, but single-quoted string continuation (`yTokBRSqtStrCtn`) was only checking `isBR`.

When a single-quoted scalar spans across lines where leading indentation is less than or equal to the scalar opening column, `isBR` (`hasNwl && leadingSpaces > curInd`) evaluates to false. So `scan()` rejected single-quoted string continuation, fell back to dedent handling (`yTokBL`), and popped `yIndStr` off the indentation stack prematurely. Once the indentation stack gets corrupted, the parser fails to recognize subsequent `---` document boundaries, collapsing multi-document streams into a single truncated document.

Updating single-quoted continuation to `(isBR || hasNwl)` brings exact parity with double-quoted scalar scanning and prevents premature dedent pops across line breaks.

#### Minimal Reproduction Example

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    pod.alpha.kubernetes.io/init-containers: '[
        {
          "name": "bootstrap"
        }
    ]'
  selector:
    app: test
---
apiVersion: v1
kind: Service
metadata:
  name: second-service
```

- **Without fix**: `stream` child count = **1 document** (`lines 1–16`), collapsing `---` and abandoning nodes past line 9.
- **With fix**: `stream` child count = **2 documents** (`doc 0: lines 1–11`, `doc 1: lines 12–16`).

## Correctness

- [ ] If grammargen or parity logic changed, ran the smallest relevant focus target in Docker

Tested on multi-document Kubernetes YAML manifests containing 200+ lines with multi-line single-quoted JSON scalars (`pod.alpha.kubernetes.io/init-containers: '[...`):
- Restores parsing from 1 collapsed document back to 4 distinct `document` AST nodes.
- Preserves all 200+ lines in the AST instead of truncating nodes after line 21.

## Performance

- [ ] If perf-relevant, used the stable bench settings (`GOMAXPROCS=1`, `-count=10`, `-benchtime=750ms`, `-benchmem`)
- [ ] If large-grammar behavior changed, checked bounded RSS, timeout, or OOM behavior under Docker

Performance is unaffected. This is a single boolean condition parity fix inside the scanner dispatch loop.

## Maintainability

## Self-review

- [x] Called out anything I'm unsure about or want a second opinion on

## Due diligence

- [x] If adding a grammar or scanner: verified against upstream C output

### DIFF
--- a/grammars/yaml_scanner.go
+++ b/grammars/yaml_scanner.go
@@ -1558,6 +1558,8 @@ func (e *yamlEnv) scan() bool {
 		(vs[yTokBRDqtStrCtn] && (isBR || hasNwl) && e.scnDqtStrCnt(yTokBRDqtStrCtn)) {
 		return true
 	}
+	// Single-quoted flow scalars use the same line-folding rule as
+	// double-quoted flow scalars. See YAML 1.2.2 productions 124 and 125.
 	if (vs[yTokRSqtStrCtn] && isR && e.scnSqtStrCnt(yTokRSqtStrCtn)) ||
 		(vs[yTokBRSqtStrCtn] && (isBR || hasNwl) && e.scnSqtStrCnt(yTokBRSqtStrCtn)) {
 		return true
--- a/grammars/yaml_scanner.go
+++ b/grammars/yaml_scanner.go
@@ -1559,7 +1559,7 @@ func (e *yamlEnv) scan() bool {
 		return true
 	}
 	if (vs[yTokRSqtStrCtn] && isR && e.scnSqtStrCnt(yTokRSqtStrCtn)) ||
-		(vs[yTokBRSqtStrCtn] && isBR && e.scnSqtStrCnt(yTokBRSqtStrCtn)) {
+		(vs[yTokBRSqtStrCtn] && (isBR || hasNwl) && e.scnSqtStrCnt(yTokBRSqtStrCtn)) {
 		return true
 	}
 
--- a/grammars/yaml_single_quote_regression_test.go
+++ b/grammars/yaml_single_quote_regression_test.go
@@ -1,0 +1,111 @@
+package grammars
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


odvcencio:
Maintainer follow-up

I pushed c9440696 with focused regression coverage and a YAML 1.2.2 specification note.

Evidence:

- The focused Go test passed on the host and in Docker.
- The locked-C YAML smoke test passed in Docker.
- A 432-case differential changed 146 completed tree shapes and changed no clean acceptance outcomes.
- PyYAML classified every generated case during the differential.
- The identical C scanner change passed deep parity on the exact locked grammar for all three focused fixtures.

The checked-in grammar lock remains unchanged. Canonical locked-C still differs because upstream lacks the analogous single-quote change.

The clean follow-up is an upstream tree-sitter-yaml patch. Update the grammar lock after that patch merges.

HarshK97:
Opened the tree-sitter-yaml upstream pr: https://github.com/tree-sitter-grammars/tree-sitter-yaml/pull/50

odvcencio:
Thank you for the focused fix and the realistic regression case. We will merge this as the immediate YAML correction. We are also moving YAML to a gotreesitter-maintained leading grammar. We will carry verified fixes locally until upstream releases equivalent behavior, whether or not upstream accepts this change. Your regression case will remain part of that coverage.

odvcencio (approved):
Approved. The scanner correction is narrow, well-explained, and covered by the regression case.

PR #730, opened 2026-08-15
parser APIs: add strict batch helpers

tsenart:
Batch consumers must reject trees that stop before all input is accepted. The pooled grammar parser and one-shot tagger exposed only editor-style APIs. Those APIs return partial output without an error.

This change adds ParseFilePooledStrict and TagStrict. Both reject a stopped parse, release the partial tree, return the typed parser stop error, and skip query work. Existing ParseFilePooled and Tag behavior stays unchanged.

Test proof:
- Focused grammar and tagger tests passed 10 times on the host.
- The same tests passed 10 times in the required 8 GiB Docker harness.
- Docker reported exit code 0, no out-of-memory kill, and no wall timeout.

The required buckley command was not installed. Git created the commits.

### DIFF
--- a/grammars/parse_file.go
+++ b/grammars/parse_file.go
@@ -58,8 +58,21 @@ func ParseFile(filename string, source []byte) (*gotreesitter.BoundTree, error) 
 
 // ParseFilePooled is like ParseFile but reuses a per-language ParserPool
 // to avoid allocating a new parser on every call. It is safe for concurrent use.
+// Parser safety stops return a partial tree without an error; use
+// ParseFilePooledStrict when partial output is not valid.
 // The caller must call Release() on the returned BoundTree.
 func ParseFilePooled(filename string, source []byte) (*gotreesitter.BoundTree, error) {
+	return parseFilePooled(filename, source, false)
+}
+
+// ParseFilePooledStrict is like ParseFilePooled, but rejects a partial tree when
+// the parser stops before it accepts all input. It releases the partial tree and
+// returns an error that wraps gotreesitter.ErrParseStoppedEarly.
+func ParseFilePooledStrict(filename string, source []byte) (*gotreesitter.BoundTree, error) {
+	return parseFilePooled(filename, source, true)
+}
+
+func parseFilePooled(filename string, source []byte, strict bool) (*gotreesitter.BoundTree, error) {
 	entry := DetectLanguage(filename)
 	if entry == nil {
 		return nil, fmt.Errorf("unsupported file type: %s", filename)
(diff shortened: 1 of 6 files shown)


tsenart:
@odvcencio Could you review and merge this PR? Perfloop is blocked on a real upstream revision. It adds strict batch APIs and keeps the editor APIs unchanged. Focused tests pass ten times on the host and in an 8 GiB container. The strict check adds about 0.6% run time and no material allocation cost. The fork account cannot enable auto-merge.

tagger.go
```diff
@@ -91,6 +91,29 @@ func (tg *Tagger) Tag(source []byte) []Tag {
...
 
+// TagStrict is like Tag, but rejects a partial tree when parsing stops before
+// it accepts all input. It releases the partial tree and returns an error that
+// wraps ErrParseStoppedEarly.
+func (tg *Tagger) TagStrict(source []byte) ([]Tag, error) {
+	if len(source) == 0 {
+		return nil, nil
+	}
+
+	tree := tg.parse(source, nil)
```
odvcencio: [P2] Propagate ordinary parser errors from `TagStrict`. `tg.parse` uses `dispatchParse`, which converts every parser error into an empty tree. This method can therefore return `nil, nil` for `ErrNoLanguage` or `ErrNoTokenSource`. Use an error-returning parse path, then preserve the documented early-stop behavior.

grammars/parse_file.go
```diff
@@ -58,8 +58,21 @@ func ParseFile(filename string, source []byte) (*gotreesitter.BoundTree, error)
...
+// ParseFilePooledStrict when partial output is not valid.
 // The caller must call Release() on the returned BoundTree.
 func ParseFilePooled(filename string, source []byte) (*gotreesitter.BoundTree, error) {
+	return parseFilePooled(filename, source, false)
+}
+
+// ParseFilePooledStrict is like ParseFilePooled, but rejects a partial tree when
+// the parser stops before it accepts all input. It releases the partial tree and
+// returns an error that wraps gotreesitter.ErrParseStoppedEarly.
+func ParseFilePooledStrict(filename string, source []byte) (*gotreesitter.BoundTree, error) {
```
odvcencio, line 71: [P2] Document `ParseFilePooledStrict` and `TagStrict` in `README.md`. The strict parsing section lists parser and incremental analysis methods, but it omits these two batch APIs. Add their result and ownership contracts before release.

tagger_test.go
```diff
@@ -49,6 +49,36 @@ func TestTagIncrementalStrictReportsTimeout(t *testing.T) {
 	}
 }
 
+func TestTagStrictReportsTimeout(t *testing.T) {
+	tagger, err := NewTagger(buildArithmeticLanguage(), "",
```
odvcencio: [P3] Use a non-empty query or instrument query execution. The empty query cannot prove that `TagStrict` skips query work after an early stop.

odvcencio (dismissed):
Request changes for one API contract defect and two release gaps.

`TagStrict` suppresses ordinary parser errors through `dispatchParse`.

Document both new APIs.

Prove that an early stop skips query execution.

The focused early-stop tests passed at commit `714ad6ef00b815b25bc9bd29bfb1c0522a8282c3`.

PR #731, opened 2026-08-15
parser: drop oversized pending stack buffers

tsenart:
## Summary

- drop empty pending-stack buffers above the existing 4096-stack emergency cap
- retain and clear ordinary buffers
- scrub both pending buffers when pooled parsers are released
- add retention, clearing, and focused benchmark coverage

## Root cause

Pending GLR queues can grow far beyond the ordinary live-stack cap. After the queue drains, GSS recycling clears the full backing capacity on every demotion even though the logical length is zero. Dropping an oversized empty buffer releases the same references without the repeated scan.

## Benchmark

Focused 4097-stack benchmark, six runs:

- drop oversized: 1.39-1.42 ns/op
- clear oversized: 1.89-2.27 us/op
- both: 0 B/op, 0 allocs/op

## Tests

- go test .
- go vet .
- targeted race test for parser pool, reset, demotion, and discard paths
- reset tests repeated 20 times

### DIFF
--- a/glr_gss_test.go
+++ b/glr_gss_test.go
@@ -1238,6 +1238,10 @@ func TestParserDemotionBoundsOversizedPendingStackBuffers(t *testing.T) {
 	if !parser.tryDemoteSingleLinearGSS(stacks, &scratch) {
 		t.Fatal("tryDemoteSingleLinearGSS() = false")
 	}
+	cold := parser.forestDeclineMemo
+	if cold == nil {
+		t.Fatal("oversized production demotion did not allocate the cold sidecar")
+	}
 	if forkBacking[0].gss.head != forkHead || frontierBacking[0].gss.head != frontierHead {
 		t.Fatal("production demotion scanned an oversized backing array")
 	}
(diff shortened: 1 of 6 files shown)


parser_pool.go
```diff
@@ -148,6 +148,10 @@ func (pp *ParserPool) release(p *Parser) {
 	// Release *Node refs so the last-used arena can be GC'd.
 	p.reuseCursor.releaseNodeRefs()
 	p.reuseScratch.releaseNodeRefs()
+	// Drop pending stack references and pathological retained capacity while
+	// the parser is idle in the pool.
+	p.pendingForkStacks = resetPendingStackBuffer(p.pendingForkStacks)
+	p.pendingFrontierForkStacks = resetPendingStackBuffer(p.pendingFrontierForkStacks)
```
odvcencio: P1: This resets only the checked-out parser. `p.clearRecoveryParser()` next returns the recovery parser through `releaseSnippetParser`. `resetSnippetParser` does not reset either pending buffer before `sync.Pool.Put`. A large recovery parse can therefore retain its backing arrays while idle. Reset both buffers in the snippet-parser release path and add a regression test.

parser.go
```diff
@@ -7339,14 +7339,8 @@ func (p *Parser) recycleDemotedGSS(stacks []glrStack, scratch *parserScratch) {
...
 	}
-	if cap(p.pendingForkStacks) > 0 {
-		clear(p.pendingForkStacks[:cap(p.pendingForkStacks)])
-		p.pendingForkStacks = p.pendingForkStacks[:0]
-	}
-	if cap(p.pendingFrontierForkStacks) > 0 {
-		clear(p.pendingFrontierForkStacks[:cap(p.pendingFrontierForkStacks)])
-		p.pendingFrontierForkStacks = p.pendingFrontierForkStacks[:0]
-	}
+	p.pendingForkStacks = resetPendingStackBuffer(p.pendingForkStacks)
```
odvcencio: P2: This call runs during each graph-structured stack demotion. Capacities above 4096 become nil. A later fork burst must allocate the backing array again. Repeated bursts can alternate between dropping and allocating large arrays. Measure this cycle, or preserve enough active-parse capacity to prevent churn.

glr_gss_test.go
```diff
@@ -1066,6 +1071,65 @@ func TestGSSScratchRecycleForParseReusesClearedSlots(t *testing.T) {
...
+
+func BenchmarkResetPendingStackBuffer(b *testing.B) {
+	const count = maxRetainedPendingStackCap + 1
+	size := int64(count) * int64(unsafe.Sizeof(glrStack{}))
+
+	b.Run("drop_oversized", func(b *testing.B) {
+		backing := make([]glrStack, 0, count)
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			benchmarkPendingStackBuffer = benchmarkResetPendingStackBuffer(backing)
```
odvcencio, line 1120: P2: This benchmark keeps `backing` live and never appends after `resetPendingStackBuffer` returns nil. It therefore measures only the branch and omits the next allocation and garbage collection cost. Add a full drop, append, drain, and reset cycle. Report bytes, allocations, and maximum resident set size.

odvcencio (dismissed):
Request changes for one retention defect and two unmeasured performance risks.

- P1: Reset snippet parsers before returning them to `snippetParserPool`.
- P2: Measure or prevent repeated drop and reallocation cycles during one parse.
- P2: Benchmark the next append, allocation, and garbage collection cost.

The helper is safe when the caller owns the slice. Existing tests do not cover these risks.

PR #734, opened 2026-08-16
fix(incremental-reuse): restore compact-tree admission

### LINKED ISSUE #728: Incremental reuse is silently disabled for every external-scanner language on the admission-candidate route
### Report type

Incorrect parse or tree

### gotreesitter version or commit

v0.49.0

### Grammar or language

broad

### Minimal reproduction

I am reporting the below regression interspersed in all the per language benchmarking on the other issue/thread but it may be getting "lost" as the regression seems to have gotten worse and not better with v0.50.0

I hope the below report assists.

## Summary

Since **v0.49.0**, `ParseIncremental` performs a **full reparse with 0% subtree reuse** for every language whose
external scanner is not marked `ExternalScannerIsStateless()` — at **every source size**. This includes `css`, `scss`,
`toml`, `typescript`, `tsx`, `ini`, `cmake` and `sql`, all of which reused 48–99% of the tree in v0.48.1.

The defect itself is **older than v0.49.0**. It has always applied to sources **below 64 KiB**. What changed in v0.49.0 is tranche B9, which removed `admissionCandidateInputSizeEligible` — the source-length gate that had been keeping larger files off the affected route. Removing the gate did not create the bug; it removed the thing that was hiding it for most real files.

Trees remain **correct**. Only reuse and wall-clock time change. There is no crash and no parse divergence, which is why a differential correctness suite would not see it.

A second, smaller bug compounds this: the reported reason string names the **GSS forest fast path**, which is provably not involved. `ParseRuntime().ForestFastPath` is `false` in 100% of the cases where that reason is emitted. The label sent us investigating the wrong subsystem for some time.

---

## Measurements

One insert edit (duplicate a single byte at a near-top site), synthetic fixtures, `ParseIncrementalProfiled`. Figures
are `ReusedBytes` as a percentage of source length. The full reproduction program is below — it is self-contained and depends only on this module.

| language | 32 KiB v0.48.1 | 32 KiB v0.50.0 | 64 KiB v0.48.1 | 64 KiB v0.50.0 | 137 KiB v0.48.1 | 137 KiB v0.50.0 |
|---|---|---|---|---|---|---|
| **go** (control) | 36.4% | 36.4% | 36.9% | 36.9% | 38.0% | 38.0% |
| css | 0.0% | 0.0% | **98.1%** | **0.0%** | **98.1%** | **0.0%** |
| scss | 0.0% | 0.0% | **98.1%** | **0.0%** | **98.2%** | **0.0%** |
| typescript | 0.0% | 0.0% | **98.6%** | **0.0%** | **98.7%** | **0.0%** |
| toml | 0.0% | 0.0% | **48.6%** | **0.0%** | **48.6%** | **0.0%** |

Three things this table establishes:

1. **`go` is completely unaffected** — identical at every size in both versions, down to the byte
   (`ReusedBytes` = 11950 / 24218 / 53359). So the candidate route *can* reuse; it refuses only for certain languages.
2. **The affected languages were already at 0% below 64 KiB in v0.48.1.** The bug predates the release that surfaced it.
3. **v0.50.0 extends 0% to all sizes.** Every row that was working is now not working.

### The v0.48.1 threshold is exactly 65536 bytes

Sweeping css on v0.48.1 across the boundary:

| fixture | bytes | reuse% |
|---|---|---|
| 60 KiB | 61476 | 0.0% |
| 62 KiB | 63528 | 0.0% |
| 63 KiB | 64554 | 0.0% |
| 64 KiB | **65580** | **98.1%** |
| 66 KiB | 67632 | 98.1% |

The flip lands precisely at `parseRuntimeMemoryMinSourceBytes = 64 * 1024` (`parser_memory_budget_runtime.go:20`), which
is the constant `admissionCandidateInputSizeEligible` compared against.

### Downstream cost

In our editor the practical effect is a whole-document reparse on every keystroke. At 137 KiB on our hardware that is
302–367 ms per edit for the affected languages, against a ~5 ms budget. Worst single figure measured: **typescript, 137
KiB, delete — 901 ms**.

Reuse percentages were **identical on arm64 and amd64** for all 48 languages that completed our sweep, which is what told us this was deterministic logic rather than a timing or memory-pressure artifact.

---

## Root cause

The chain, all line numbers against **v0.50.0**:

**1. The candidate route stamps trees as reuse-disabled at construction.**

`parsercore_phase0_driver.go:3533`:

```go
tree.incrementalReuseDisabled = !incrementalReuseProven
tree.compactMaterialized = true
```

**2. `incrementalReuseProven` requires the *strongest* quiescence verdict.**

`parsercore_phase0_driver.go:3200`:

```go
incrementalReuseProven := replayStates != nil &&
    classifyExternalScannerQuiescence(parser.language) == scannerQuiescenceProven
```

**3. But `classifyExternalScannerQuiescence` returns `Proven` only for a stateless scanner or no scanner at all.**

`external_scanner_quiescence.go:128`. A language whose scanner implements `IncrementalReuseExternalScanner` and returns
`SupportsIncrementalReuse() == true` is *not* refuted, so it falls through to the final `return
scannerQuiescenceUnknown` at line 152.

This is the crux. These scanners **explicitly declare that they support incremental reuse**:

```go
// grammars/css_scanner.go:42
func (CssExternalScanner) SupportsIncrementalReuse() bool { return true }
```

Same for `scss_scanner.go:41`, `toml_scanner.go:43`, `typescript_scanner.go:42`. Each affirms support, is classified
`Unknown`, and then has reuse disabled anyway.

**4. `Unknown` is documented as neutral — but read here as a rejection.**

The classifier's own contract, `external_scanner_quiescence.go:110-114`:

> `scannerQuiescenceUnknown`: the W4 classifier neither proves nor refutes quiescence. The legacy per-language
> admission (`languageSupportsIncrementalReuse`, checked before `tryReuseSubtree` runs) **stays the authority**. This
> keeps the predicate **neutral** for every production external-scanner language that has not yet migrated to a
> per-boundary proof.

The production route honours that contract: it consults `languageSupportsIncrementalReuse` (`parser.go:3184`), which
grants reuse — hence 98.1% at ≥64 KiB in v0.48.1. The candidate route never consults it. It tests `== Proven`, which
collapses the neutral middle verdict into the reject bucket, and stamps the tree before any per-boundary gate can run.

**The correctly-shaped predicate already exists in the same file** — `external_scanner_quiescence.go:160`:

```go
func externalScannerBoundaryQuiescentWithoutCheckpoint(lang *Language) bool {
	return classifyExternalScannerQuiescence(lang) != scannerQuiescenceRefuted
}
```

It is fail-closed on `Refuted` while staying neutral on `Unknown` — exactly the semantics the doc describes. Its only
caller is `external_scanner_checkpoints.go:372`. The compact route does not use it.

**5. Tranche B9 removed the size gate that was confining the blast radius.**

v0.48.1, `admission_switch.go:309-311`:

```go
func admissionCandidateInputSizeEligible(sourceLen int) bool {
	return sourceLen < parseRuntimeMemoryMinSourceBytes
}
```

The symbol has **zero references in v0.50.0**. Per the updated doc comment:

> Tranche B9 removed the source-length eligibility decline: every input, regardless of size, is eligible to attempt
> the candidate route.

B9 looks correct on its own terms — the stated goal was that the scheduler's own memory-budget poll now covers what the
size decline used to cover. The unintended consequence is that the size decline was *also*, incidentally, the only thing
keeping files ≥64 KiB away from a route that cannot reuse for most languages.

**6. On the next edit, the stamp is honoured and a fresh parse is forced.**

`parser.go:3058-3074` and `parser_api.go:1692-1700` both branch on `oldTreeDisablesIncrementalReuse(oldTree)`
(`parser_api.go:183`) and fall back to a full parse.

---

## Secondary bug: the reason string misattributes the cause

`parser_api.go:178`:

```go
const forestIncrementalReuseUnsupportedReason = "old tree was built by GSS forest fast path"
```

`Tree.incrementalReuseDisabled` is set from two places — the forest fast path (`glr_forest.go:415`, `glr_forest.go:765`)
and the compact materialization path (`parsercore_phase0_driver.go:3533`). The reported reason hardcodes the forest
explanation for both.

In every case we measured, the old tree was **not** built by the forest fast path:

- `ParseRuntime().ForestFastPath` is `false` in **every row we measured** — five languages, ten sizes, both versions.
- None of the affected languages appears in `builtinForestDefaults` (`glr_forest.go:583`).
- None has `automaticForestEnabled` in `grammars/runtime_profiles.go` — that set is unchanged between the two versions
  and does not contain them.

We spent a meaningful amount of time diffing `glr_forest.go` and the forest admission tables before noticing the flag
was false. Reporting the actual source — forest vs. compact-materialized — would have made this a short investigation.

---

## Suggested fix

Have the compact route respect the same contract the production route does: treat only `Refuted` as disqualifying, and
otherwise defer to the legacy per-language admission.

```go
// parsercore_phase0_driver.go:3200
incrementalReuseProven := replayStates != nil &&
    externalScannerBoundaryQuiescentWithoutCheckpoint(parser.language) &&
    languageSupportsIncrementalReuse(parser.language)
```

Both helpers already exist and are already the authority on the production route, so this should not widen what reuse is
permitted overall — it makes the two routes agree.

For the diagnostic, distinguishing the two producers would help consumers a great deal:

```go
const forestIncrementalReuseUnsupportedReason  = "old tree was built by GSS forest fast path"
const compactIncrementalReuseUnsupportedReason = "old tree was compact-materialized without a scanner-quiescence proof"
```

### A regression test that would have caught this

The v0.50.0 differential suite compares against the C runtime, but it cannot see this class: the tree produced here is
**correct**, so nothing diverges. Only the reuse counters and the clock move.

A ratchet on `IncrementalParseProfile.ReusedBytes` would catch it directly — for a fixed (language, fixture, edit),
assert reuse does not fall below the previous release's figure. Two fixture sizes per language, one below 64 KiB and one
above, would have caught both this regression *and* the older sub-64 KiB case that has been present for several
releases.

We are happy to contribute our fixtures and the probe below if useful.

---

## Reproduction

Self-contained; no dependencies beyond this module. Swap the `require` between `v0.48.1` and `v0.50.0` to see the delta.

```go
package main

import (
	"bytes"
	"fmt"

	ts "github.com/odvcencio/gotreesitter"
)

func genCSS(n int) []byte {
	var b bytes.Buffer
	for i := 0; b.Len() < n; i++ {
		fmt.Fprintf(&b, ".c%d { color: #012345; margin: 0px; padding: 1px; }\n", i)
	}
	return b.Bytes()
}

func pointAt(src []byte, off int) ts.Point {
	row, col := 0, 0
	for i := 0; i < off && i < len(src); i++ {
		if src[i] == '\n' {
			row++
			col = 0
		} else {
			col++
		}
	}
	return ts.Point{Row: uint32(row), Column: uint32(col)}
}

func main() {
	lang := grammars.CssLanguage()
	for _, kb := range []int{32, 60, 63, 64, 66, 137} {
		src := genCSS(kb * 1024)
		site := bytes.Index(src, []byte("012345"))
		sp := pointAt(src, site)
		p1 := ts.Point{Row: sp.Row, Column: sp.Column + 1}

		// insert: duplicate one byte at the edit site
		ins := make([]byte, 0, len(src)+1)
		ins = append(ins, src[:site]...)
		ins = append(ins, src[site])
		ins = append(ins, src[site:]...)

		parser := ts.NewParser(lang)
		old, _ := parser.Parse(src)
		old.Edit(ts.InputEdit{
			StartByte: uint32(site), OldEndByte: uint32(site), NewEndByte: uint32(site + 1),
			StartPoint: sp, OldEndPoint: sp, NewEndPoint: p1,
		})
		_, prof, _ := parser.ParseIncrementalProfiled(ins, old)

		fmt.Printf("%6dKB  bytes=%7d  forest=%-5v  reused=%7d  %5.1f%%  %s\n",
			kb, len(src), old.ParseRuntime().ForestFastPath, prof.ReusedBytes,
			100*float64(prof.ReusedBytes)/float64(len(src)),
			prof.ReuseUnsupportedReason)
	}
}
```

Expected on **v0.48.1** — 0% below 65536 bytes, ~98% at or above:

```
    32KB  bytes=  32803  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
    60KB  bytes=  61476  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
    63KB  bytes=  64554  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
    64KB  bytes=  65580  forest=false  reused=  64315   98.1%
    66KB  bytes=  67632  forest=false  reused=  66329   98.1%
   137KB  bytes= 140316  forest=false  reused= 137667   98.1%
```

Expected on **v0.50.0** — 0% everywhere:

```
    32KB  bytes=  32803  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
    60KB  bytes=  61476  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
    63KB  bytes=  64554  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
    64KB  bytes=  65580  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
    66KB  bytes=  67632  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
   137KB  bytes= 140316  forest=false  reused=      0    0.0%  old tree was built by GSS forest fast path
```

For the control, swap in `grammars.GoLanguage()`, the edit site `"x := a + b"`, and a Go fixture generator:

```go
func genGo(n int) []byte {
	var b bytes.Buffer
	b.WriteString("package main\n\n")
	for i := 0; b.Len() < n; i++ {
		fmt.Fprintf(&b, "func f%d(a, b int) int {\n\tx := a + b\n\treturn x + %d\n}\n\n", i, i)
	}
	return b.Bytes()
}
```

That gives byte-identical reuse in both versions at every size — 36.4% / 36.9% / 38.0% at 32 / 64 / 137 KiB
(`ReusedBytes` = 11950 / 24218 / 53359). Go's scanner is the one that implements `ExternalScannerIsStateless()`.

---

## Environment

| | |
|---|---|
| Versions compared | v0.48.1 (last good for us) vs v0.50.0 |
| Go | 1.26.4 |
| OS | Windows 11 (26200) |
| Arch | arm64 native; windows/amd64 under x64 emulation |
| Consumer | a Go GUI toolkit's code editor, via `ParseIncrementalProfiled` |

Reuse percentages were identical across both architectures for every language in our 50-language sweep, so this
reproduces independently of arch.

---

## Why we are reporting this now

We pin the vendored engine and re-run a 50-language measurement harness on each release. v0.49.0 and v0.50.0 were both
measured and both declined on this regression alone; we are still shipping v0.48.1. v0.50.0 contains fixes we would like
to take — the haskell blowup and the `c` correctness fix in particular — and this is the only thing blocking us.

Happy to test a patch, contribute fixtures, or provide the full 50-language before/after data.

## (1) Expected behavior

A single-byte insert into an already-parsed document should reuse the bulk of the existing tree rather than rebuilding
it.

Concrete expected figures, **produced by this engine itself** on the identical input and edit (css, 137 KiB, one insert
at a near-top site):

```
reusedBytes=137667 (98.1%)  reusedSubtrees=2625  newNodesAllocated=2640  blockSpliceSteps=2625
reuseUnsupported=false
```

That is both:

- v0.48.1's default behaviour at this size, and
- v0.50.0's own behaviour on the same input when the parse is routed through production via
  `parser.SetAdmissionCandidateRoute(false)`.

So no external oracle is needed to establish the expectation. I do not have a C tree-sitter run for these fixtures, but
it is not the relevant baseline here: the engine already achieves the expected result on this exact input, and only the
internal route selection changes the outcome. This is a route-dependent regression, not a parity question.

Per-language expected reuse (insert edit, ≥64 KiB, v0.48.1 defaults):

| language | expected reuse |
|---|---|
| css | 98.1% |
| scss | 98.1–98.2% |
| typescript | 98.6–98.7% |
| toml | 48.6% |
| go (control) | 36.4–38.0% |

Two secondary expectations:

1. `scannerQuiescenceUnknown` should behave as its own documentation specifies — neutral, deferring to the legacy
   per-language admission (`languageSupportsIncrementalReuse`) — rather than being treated as a refusal.
2. `IncrementalParseProfile.ReuseUnsupportedReason` should name the route that actually disabled reuse.

## (2) Actual behavior and fresh-parse comparison

**Actual: 0% reuse — a full reparse on every edit** for every language whose external scanner is not marked
`ExternalScannerIsStateless()`. Confirmed for css, scss, toml, typescript; our wider sweep also shows tsx, ini, cmake
and sql.

Work counters, css 137 KiB, v0.50.0, default options:

```
reusedBytes=0 (0.0%)  reusedSubtrees=0  newNodesAllocated=94284  blockSpliceSteps=0
rejects: dirty=0 ancestorDirty=0 hasError=0 invalidSpan=0 outOfBounds=0
         rootNonLeaf=0 largeNonLeaf=0 staleBoundary=0 fragile=0
reuseUnsupported=true  reason="old tree was built by GSS forest fast path"
ParseRuntime().ForestFastPath=false
```

Same parser, same input, `SetAdmissionCandidateRoute(false)`:

```
reusedBytes=137667 (98.1%)  reusedSubtrees=2625  newNodesAllocated=2640  blockSpliceSteps=2625
rejects: dirty=1 ancestorDirty=4 hasError=0 invalidSpan=0 outOfBounds=0
         rootNonLeaf=9 largeNonLeaf=0 staleBoundary=2 fragile=0
reuseUnsupported=false  reason=""
```

Two things stand out:

- **Every reject counter is zero in the failing case.** Reuse is not being rejected per-node; it is refused at the gate
  before the reuse cursor ever runs. The working case shows non-zero rejects (dirty=1, ancestorDirty=4, rootNonLeaf=9,
  staleBoundary=2), which is what a cursor that actually ran looks like.
- **Node allocation is 94284 vs 2640 — 35.7× more** for the same one-byte edit.

**Fresh-parse comparison: the trees are identical, and that is the point.**

The incremental result matches a fresh full parse exactly — node count, root type and error flag all agree; our harness
asserts this on every measurement and it passes. That is expected, because `ParseIncremental` is internally *falling
back to* a fresh full parse. There is no tree divergence, no error, and no crash. The entire cost is wall-clock.

This is why a differential correctness suite against the C runtime cannot see this class of defect: the output is right,
only the work is wrong.

Timing impact in our editor at 137 KiB: **302–367 ms per keystroke** for the affected languages, against a ~5 ms budget.
Worst single figure measured: **typescript, 137 KiB, delete — 901 ms**.

Regression table (`ReusedBytes` as % of source; insert edit):

| language | 32 KiB v0.48.1 | 32 KiB v0.50.0 | 64 KiB v0.48.1 | 64 KiB v0.50.0 | 137 KiB v0.48.1 | 137 KiB v0.50.0 |
|---|---|---|---|---|---|---|
| **go** (control) | 36.4% | 36.4% | 36.9% | 36.9% | 38.0% | 38.0% |
| css | 0.0% | 0.0% | **98.1%** | **0.0%** | **98.1%** | **0.0%** |
| scss | 0.0% | 0.0% | **98.1%** | **0.0%** | **98.2%** | **0.0%** |
| typescript | 0.0% | 0.0% | **98.6%** | **0.0%** | **98.7%** | **0.0%** |
| toml | 0.0% | 0.0% | **48.6%** | **0.0%** | **48.6%** | **0.0%** |

Note the 32 KiB column: **this predates v0.49.0.** Sources below 64 KiB have always taken the affected route. What
v0.49.0 changed (tranche B9, removing `admissionCandidateInputSizeEligible`) was to extend that route to all sizes,
which removed the accident that had been hiding the defect for larger files.

`go` is unaffected everywhere because its scanner is the one implementing `ExternalScannerIsStateless()`.

## (3) Environment

| | |
|---|---|
| OS | Windows 11, build 26200 |
| Arch | arm64 (native) and windows/amd64 (x64 emulation, same machine) |
| Go | 1.26.4 |
| Versions compared | v0.48.1 (last good for us) vs v0.50.0 |
| Entry point | `Parser.ParseIncrementalProfiled` |

Reuse percentages were **identical on both architectures** for all 48 languages that completed our sweep, so this
reproduces independently of arch and is deterministic logic rather than a timing or memory-pressure effect.

**Parser options: all defaults.** The following are specifically *not* set, and that matters — any one of them forces
the parse onto the production route and masks the bug completely:

- no explicit timeout, no cancellation flag
- no included ranges
- no logger, GLR trace, ambiguity profile, or parse-progress hook

**Environment variables: none set.** In particular `GOT_GLR_FOREST` is unset, so the GSS forest path is enabled by
default — but it is not involved: `ParseRuntime().ForestFastPath` is `false` in every row we measured, despite being
named in the reason string.

v0.49.0 showed the same 0% reuse in our earlier evaluation of that release; the measurements above are from v0.48.1 vs
v0.50.0.

## (4) Optional diagnostics

**Workaround, and the strongest single diagnostic.** `parser.SetAdmissionCandidateRoute(false)` fully restores reuse on
v0.50.0 at every size:

```
137KB candidateRoute=true    reusedBytes=0       0.0%
137KB candidateRoute=false   reusedBytes=137667  98.1%
 32KB candidateRoute=true    reusedBytes=0       0.0%
 32KB candidateRoute=false   reusedBytes=32152   98.0%
```

On v0.48.1 the same call also recovers the sub-64 KiB case (32 KiB: 0.0% → 98.0%), which is further evidence the two
symptoms share one cause.

**The v0.48.1 threshold is exactly 65536 bytes** (`parseRuntimeMemoryMinSourceBytes`), sweeping css:

| fixture | bytes | reuse% |
|---|---|---|
| 60 KiB | 61476 | 0.0% |
| 62 KiB | 63528 | 0.0% |
| 63 KiB | 64554 | 0.0% |
| 64 KiB | **65580** | **98.1%** |
| 66 KiB | 67632 | 98.1% |

**Root cause pointer** (line numbers against v0.50.0):

- `parsercore_phase0_driver.go:3533` — `tree.incrementalReuseDisabled = !incrementalReuseProven`
- `parsercore_phase0_driver.go:3200` — `incrementalReuseProven := replayStates != nil &&
  classifyExternalScannerQuiescence(parser.language) == scannerQuiescenceProven`
- `external_scanner_quiescence.go:128-152` — returns `scannerQuiescenceUnknown` for any scanner that is not marked
  stateless, including scanners that explicitly return `SupportsIncrementalReuse() == true`
  (`grammars/css_scanner.go:42`, `scss_scanner.go:41`, `toml_scanner.go:43`, `typescript_scanner.go:42`)
- `external_scanner_quiescence.go:110-114` — documents `Unknown` as **neutral**, with `languageSupportsIncrementalReuse`
  (`parser.go:3184`) as the authority
- `external_scanner_quiescence.go:160` — `externalScannerBoundaryQuiescentWithoutCheckpoint` already implements the
  correct predicate (`!= scannerQuiescenceRefuted`); its only caller is `external_scanner_checkpoints.go:372`
- `parser_api.go:178` — the reason constant hardcodes the forest explanation for both producers of the flag
  (`glr_forest.go:415`/`:765` and the compact path above)

**Suggested regression test.** A ratchet on `IncrementalParseProfile.ReusedBytes` — for a fixed (language, fixture,
edit), assert reuse does not fall below the previous release's figure. Two fixture sizes per language, one below 64 KiB
and one above, would have caught both this regression and the older sub-64 KiB case. The existing differential suite
cannot catch it, since the tree is correct.

Happy to contribute fixtures, the probe program, or our full 50-language before/after dataset.

### Checklist

- [x] I reduced this to a minimal reproduction or explained why I could not.
- [x] I removed secrets, private source, and other sensitive data from this report.

odvcencio:
# Issue 728 pull request closure packet

Status: Ready for pull request creation. This audit did not open a pull request.

Title: `fix(incremental-reuse): restore compact-tree admission`

## Body

Restore incremental subtree reuse on the compact-tree route for admitted external scanners.

- Reject reuse only when the scanner classifier refutes quiescence.
- Apply the existing language admission gate to compact trees.
- Report compact-tree and forest-path refusal reasons separately.
- Add regression coverage for nine languages at 63, 65, and 137 KiB.
- Keep Python and Starlark reuse fail-closed.

Closes #728

### Scope

- Commit: `13a20edaa331f973a15100def6734ae1fe213689`
- Parent: `8c17e0e942e0a68cf99a820ad0fc047bf2238cfa`
- Seven files changed:
  - `external_scanner_quiescence.go`
  - `external_scanner_quiescence_test.go`
  - `issue728_incremental_reuse_regression_test.go`
  - `parser.go`
  - `parser_api.go`
  - `parsercore_phase0_driver.go`
  - `parsestate_reuse_bar_test.go`

### Correctness

- Docker compile passed with exit code 0 and no out-of-memory event.
- Docker unit receipt passed 27 reuse cases and 2 fail-closed cases.
- Docker race receipt passed the same 29 issue cases.
- Build-tagged compact unit tests passed both compact-tree tests.
- Build-tagged compact race tests passed both compact-tree tests.
- Docker C-reference smoke passed fresh and incremental Go parity.
- C-reference maximum resident set size was 223,840 kB.
- `git diff --check` passed.

Receipts:

- Compile: `/tmp/issue728-final-artifacts/20260816T133805Z-issue728-final-compile`
- Unit: `/tmp/issue728-final-artifacts/20260816T133818Z-issue728-final-unit`
- Race: `/tmp/issue728-final-artifacts/20260816T133830Z-issue728-final-race`
- Compact unit: `/tmp/issue728-compact-gates/20260816T170003Z-compact-unit`
- Compact race: `/tmp/issue728-compact-gates/20260816T170010Z-compact-race`
- C-reference smoke: `/tmp/issue728-final-artifacts/20260816T134004Z-issue728-final-go-cgo-parity-go`

SQL, Python, and Starlark remain fail-closed where their scanner proof is absent.
The incremental trees match fresh trees in the regression cases.

### Performance evidence

The directional comparison used the primary benchmark trio, seeds 1 through 20, 750 ms per benchmark, and benchmark memory reporting.
Each input has 60 samples and six observed benchmark-order permutations.

- Before receipt: `/tmp/gotreesitter-issue728-bench-before.txt`
- After receipt: `/tmp/gotreesitter-issue728-bench-after.txt`
- Before SHA-256: `561ed252b7efc51bb1e7eb8437c0cc9b02dedad4d11ac627c9d0f19746a00359`
- After SHA-256: `b90296cab079f8f5ae46f442a50d7bd6cb922729860cb9d5d192250e813a776b`
- Full parse: `-5.14%`
- Incremental single-byte edit: `-5.30%`
- Incremental no edit: `-4.40%`
- Geometric mean: `-4.95%`

The receipts predate the later invalidated overlapping host campaigns. This pull request makes no campaign-credit claim.
Final-head C0f must establish campaign credit after the change reaches accepted main.

### Residuals

- The existing TypeScript parity qualification remains unchanged.
- This change does not claim a new full-fleet parity result.
- The final-head C0f gate remains required by the campaign.

### Release context

Release `v0.51.0` is published from main commit `8c17e0e942e0a68cf99a820ad0fc047bf2238cfa`.

### DIFF
--- a/external_scanner_quiescence.go
+++ b/external_scanner_quiescence.go
@@ -160,3 +160,11 @@ func classifyExternalScannerQuiescence(lang *Language) externalScannerQuiescence
 func externalScannerBoundaryQuiescentWithoutCheckpoint(lang *Language) bool {
 	return classifyExternalScannerQuiescence(lang) != scannerQuiescenceRefuted
 }
+
+// compactIncrementalReuseProvenForLanguage keeps the classifier's Unknown
+// result neutral only when the legacy language admission permits reuse.
+// Refuted scanners remain fail-closed.
+func compactIncrementalReuseProvenForLanguage(lang *Language) bool {
+	return externalScannerBoundaryQuiescentWithoutCheckpoint(lang) &&
+		languageSupportsIncrementalReuse(lang)
+}
--- a/external_scanner_quiescence_test.go
+++ b/external_scanner_quiescence_test.go
@@ -19,6 +19,7 @@ type quiescenceStatelessScanner struct {
 }
 
 func (quiescenceStatelessScanner) ExternalScannerIsStateless() bool { return true }
+func (quiescenceStatelessScanner) SupportsIncrementalReuse() bool   { return true }
 
 type quiescenceCheckpointScanner struct {
 	parserTestSafeExternalScanner
@@ -95,6 +96,35 @@ func TestExternalScannerBoundaryQuiescentFailsClosed(t *testing.T) {
 	}
 	if !externalScannerBoundaryQuiescentWithoutCheckpoint(&Language{}) {
 		t.Fatal("no-scanner language rejected: predicate must admit a proven boundary")
+	}
+}
+
+func TestCompactIncrementalReuseProvenForLanguage(t *testing.T) {
+	cases := []struct {
+		name string
+		lang *Language
+		want bool
+	}{
+		{name: "no scanner", lang: &Language{}, want: true},
+		{name: "stateless scanner", lang: &Language{ExternalScanner: quiescenceStatelessScanner{}}, want: true},
+		{name: "legacy admitted unknown scanner", lang: &Language{ExternalScanner: parserTestSafeExternalScanner{}}, want: true},
+		{name: "refuted scanner", lang: &Language{ExternalScanner: quiescenceOptOutScanner{}}, want: false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := compactIncrementalReuseProvenForLanguage(tc.lang); got != tc.want {
+				t.Fatalf("compactIncrementalReuseProvenForLanguage = %t, want %t", got, tc.want)
+			}
+		})
+	}
+}
+
+func TestIncrementalReuseUnsupportedReasonIdentifiesTreeProducer(t *testing.T) {
+	if got := incrementalReuseUnsupportedReasonForTree(&Tree{compactMaterialized: true, incrementalReuseDisabled: true}); got != compactIncrementalReuseUnsupportedReason {
+		t.Fatalf("compact tree reason = %q, want %q", got, compactIncrementalReuseUnsupportedReason)
+	}
+	if got := incrementalReuseUnsupportedReasonForTree(&Tree{incrementalReuseDisabled: true}); got != forestIncrementalReuseUnsupportedReason {
+		t.Fatalf("forest tree reason = %q, want %q", got, forestIncrementalReuseUnsupportedReason)
 	}
 }
 
--- a/issue728_incremental_reuse_regression_test.go
+++ b/issue728_incremental_reuse_regression_test.go
@@ -1,0 +1,263 @@
+package gotreesitter_test
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"strings"
+	"testing"
+
+	gotreesitter "github.com/odvcencio/gotreesitter"
+	"github.com/odvcencio/gotreesitter/grammars"
+)
+
+type issue728ReuseCase struct {
+	name      string
+	lang      func() *gotreesitter.Language
+	source    func(int) []byte
+	marker    []byte
+	insert    byte
+	wantReuse bool
+}
+
+func TestIssue728IncrementalReuseMeasurement(t *testing.T) {
+	cases := []issue728ReuseCase{
+		{name: "css", lang: grammars.CssLanguage, source: issue728CSSSource, marker: []byte(".c"), insert: 'c', wantReuse: true},
+		{name: "scss", lang: grammars.ScssLanguage, source: issue728SCSSSource, marker: []byte(".c"), insert: 'c', wantReuse: true},
+		{name: "typescript", lang: grammars.TypescriptLanguage, source: issue728TypeScriptSource, marker: []byte("export function f"), insert: 'f', wantReuse: true},
+		{name: "tsx", lang: grammars.TsxLanguage, source: issue728TypeScriptSource, marker: []byte("export function f"), insert: 'f', wantReuse: true},
+		{name: "toml", lang: grammars.TomlLanguage, source: issue728TOMLSource, marker: []byte("key"), insert: 'k', wantReuse: true},
+		{name: "cmake", lang: grammars.CmakeLanguage, source: issue728CMakeSource, marker: []byte("VAR"), insert: 'X', wantReuse: true},
+		{name: "sql", lang: grammars.SqlLanguage, source: issue728SQLSource, marker: []byte("AS value_"), insert: 'x', wantReuse: false},
+		{name: "ini", lang: grammars.IniLanguage, source: issue728INISource, marker: []byte("key"), insert: 'x', wantReuse: true},
+		{name: "go", lang: grammars.GoLanguage, source: issue728GoSource, marker: []byte("func f"), insert: 'f', wantReuse: true},
+	}
+	for _, tc := range cases {
+		for _, size := range []int{63 << 10, 65 << 10, 137 << 10} {
+			t.Run(fmt.Sprintf("%s/%dKiB", tc.name, size>>10), func(t *testing.T) {
+				src := tc.source(size)
+				if len(src) < size {
+					t.Fatalf("source length = %d, want at least %d bytes", len(src), size)
+				}
+				if size < 64<<10 && len(src) >= 64<<10 {
+					t.Fatalf("source length = %d, want below 64 KiB", len(src))
+				}
+				if size >= 64<<10 && len(src) < 64<<10 {
+					t.Fatalf("source length = %d, want at least 64 KiB", len(src))
+				}
+				offset := bytes.Index(src, tc.marker)
+				if offset < 0 {
+					t.Fatalf("source does not contain marker %q", tc.marker)
+				}
+				editOffset := offset + len(tc.marker)
+				edited := make([]byte, 0, len(src)+1)
+				edited = append(edited, src[:editOffset]...)
+				edited = append(edited, tc.insert)
+				edited = append(edited, src[editOffset:]...)
+				point := issue728PointAt(src, editOffset)
+
+				gotreesitter.ResetAdmissionCandidateCountersForTest()
+				parser := gotreesitter.NewParser(tc.lang())
+				parser.SetAdmissionCandidateRoute(true)
+				oldTree, err := parser.Parse(src)
+				if err != nil {
+					t.Fatalf("initial parse: %v", err)
+				}
+				defer oldTree.Release()
+				oldRouted, oldFallbacks := gotreesitter.AdmissionCandidateCounters()
+				if oldRouted != 1 || oldFallbacks != 0 {
+					t.Fatalf("candidate route counters = %d/%d, want 1/0 (reason=%q)", oldRouted, oldFallbacks, gotreesitter.AdmissionCandidateLastFallbackReason())
+				}
+				if oldTree.UsedForestFastPath() || oldTree.ParseRuntime().ForestFastPath {
+					t.Fatal("candidate fresh tree unexpectedly used the forest route")
+				}
+				oldTree.Edit(gotreesitter.InputEdit{
+					StartByte: uint32(editOffset), OldEndByte: uint32(editOffset), NewEndByte: uint32(editOffset + 1),
+					StartPoint: point, OldEndPoint: point, NewEndPoint: issue728PointAt(edited, editOffset+1),
+				})
+				newTree, profile, err := parser.ParseIncrementalProfiled(edited, oldTree)
+				if err != nil {
+					t.Fatalf("incremental parse: %v", err)
+				}
+				defer newTree.Release()
+				newRouted, newFallbacks := gotreesitter.AdmissionCandidateCounters()
+				if newRouted != oldRouted || newFallbacks != oldFallbacks {
+					t.Fatalf("incremental parse changed candidate route counters: before=%d/%d after=%d/%d", oldRouted, oldFallbacks, newRouted, newFallbacks)
+				}
+				if tc.wantReuse {
+					if profile.ReuseUnsupported || profile.ReuseUnsupportedReason != "" {
+						t.Fatalf("incremental reuse remained unsupported: %+v", profile)
+					}
+					if profile.ReusedBytes == 0 || profile.ReusedSubtrees == 0 || profile.NewNodesAllocated == 0 {
+						t.Fatalf("incremental reuse counters are empty: %+v", profile)
+					}
+				} else if profile.ReuseUnsupported || profile.ReuseUnsupportedReason != "" || profile.ReusedBytes != 0 || profile.ReusedSubtrees != 0 || profile.ReuseRejectScannerUnquiescent == 0 {
+					t.Fatalf("checkpoint-refuted scanner was not rejected at reuse boundaries: %+v", profile)
+				}
+				if !profile.OldTreeReuseRoute || !newTree.ParseRuntime().IncrementalOldTreeReuseRoute {
+					t.Fatalf("incremental route tag is missing: profile=%+v runtime=%s", profile, newTree.ParseRuntime().Summary())
+				}
+				if newTree.UsedForestFastPath() || newTree.ParseRuntime().ForestFastPath {
+					t.Fatal("incremental result unexpectedly used the forest route")
+				}
+
+				freshParser := gotreesitter.NewParser(tc.lang())
+				freshParser.SetAdmissionCandidateRoute(false)
+				freshTree, err := freshParser.Parse(edited)
+				if err != nil {
+					t.Fatalf("fresh parse: %v", err)
+				}
+				defer freshTree.Release()
+				freshDigest := issue728TreeDigest(freshTree, tc.lang())
+				requireIncrementalDeepTreeMatchesFresh(t, newTree, freshTree, tc.lang())
+				if newTree.RootNode().HasError() != freshTree.RootNode().HasError() ||
+					newTree.RootNode().Type(tc.lang()) != freshTree.RootNode().Type(tc.lang()) ||
+					issue728TreeDigest(newTree, tc.lang()) != freshDigest {
+					t.Fatalf("incremental tree differs from fresh tree: incremental=%s fresh=%s", issue728TreeDigest(newTree, tc.lang()), freshDigest)
+				}
+				t.Logf("source=%d oldRoute={forest=%v runtimeForest=%v usedForest=%v routed=%d fallbacks=%d} profile={reusedBytes=%d reusedSubtrees=%d newNodes=%d unsupported=%v reason=%q oldTreeReuse=%v scannerRejects=%d reparseNanos=%d} newRoute={forest=%v runtimeForest=%v incrementalReuse=%v routed=%d fallbacks=%d} freshDigest=%s",
+					len(src),
+					oldTree.UsedForestFastPath(), oldTree.ParseRuntime().ForestFastPath, oldTree.UsedForestFastPath(), oldRouted, oldFallbacks,
+					profile.ReusedBytes, profile.ReusedSubtrees, profile.NewNodesAllocated, profile.ReuseUnsupported, profile.ReuseUnsupportedReason, profile.OldTreeReuseRoute, profile.ReuseRejectScannerUnquiescent, profile.ReparseNanos,
+					newTree.UsedForestFastPath(), newTree.ParseRuntime().ForestFastPath, newTree.ParseRuntime().IncrementalOldTreeReuseRoute, newRouted, newFallbacks,
+					freshDigest,
+				)
+			})
+		}
+	}
+}
+
+func TestIssue728RefutedExternalScannersFailClosed(t *testing.T) {
+	cases := []issue728ReuseCase{
+		{name: "python", lang: grammars.PythonLanguage, source: issue728PythonSource, marker: []byte("value"), insert: 'x'},
+		{name: "starlark", lang: grammars.StarlarkLanguage, source: issue728PythonSource, marker: []byte("value"), insert: 'x'},
+	}
+	for _, tc := range cases {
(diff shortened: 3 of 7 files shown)


odvcencio (commented):
Independent deep review found no P0-P3 findings.

Verified the exact base and head, fail-closed scanner admission, route-specific reasons, 27 reuse witnesses, two rejection witnesses, compact build-tag tests, focused Docker race tests, no out-of-memory event, and the seven-file one-commit scope. The receipt source worktree is clean at the exact head, with a recorded maximum resident set size of 223,840 kB.

This pull request makes no campaign performance-credit claim.

PR #738, opened 2026-08-16
fix(parser): end convergence after clean condense

tsenart:
## What does this PR do?

This change returns clean suffixes to ordinary merging after condense removes all active recovery state.
It deletes the separate pending-fork convergence latch.

Before this change, parsing stayed correct but used avoidable CPU time and memory.

## Why this approach?

The pending-fork latch kept cap-one convergence active after recovery costs became irrelevant.
This retained hundreds of equivalent stacks on a valid generated C initializer.

The active recovery state already owns this decision.
Keeping a second latch created a stale authority path.

## Correctness

- [x] Focused package tests pass for the changed recovery boundary.
- [x] Exact C-runtime parity passes for the generated C initializer.
- [x] Exact C-runtime parity passes for the two affected Kotlin fixtures.
- [x] The merge census passes with the new counts.
- [x] The Docker smoke parity ring passes.
- [x] The focused race test and `go vet` pass.

The generated regression test requires:

- a complete, accepted tree;
- no error nodes;
- an exact source span;
- no truncation;
- at most eight live stacks.

The baseline reached 413 live stacks. The candidate reached four.

## Performance

The benchmark uses generated input, 20 shuffle seeds, `GOMAXPROCS=1`, a 750 ms duration, and `-benchmem`.

| Metric | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| Time | 1.261 s/op | 2.193 ms/op | -99.83% |
| Bytes | 70.7 MiB/op | 190 KiB/op | -99.74% |
| Allocations | 221,190/op | 1,339/op | -99.39% |

`benchstat` reports `p=0.000` for all three metrics.
Absolute times varied on the test host, but the result groups stayed separate.

- [x] Used the repository's stable benchmark settings.
- [x] Included before and after numbers.
- [x] The Docker parity ring completed without an out-of-memory failure.

## Maintainability

- [x] No unnecessary abstractions or unrelated cleanup.
- [x] Harness changes are limited to required parity proof.
- [x] No new dependencies.

The change deletes one parser field, one helper, and tests for the old mechanism.

## Self-review

- [x] Reviewed the full diff.
- [x] Reviewed the recovery boundary with independent performance and minimality lenses.
- [x] No unresolved uncertainty remains.

## Due diligence

- [x] Read the changed parser and merge paths.
- [x] Kept the existing parser state pattern.
- [x] Validated the affected grammars against the C runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Fixed
+
+- End recovery convergence after a clean condense. Pending forks no longer
+  keep clean C initializer suffixes on the recovery merge path.
+
 ## [0.51.0] - 2026-08-16
 
 ### Added
--- a/benchmark_c_clean_suffix_test.go
+++ b/benchmark_c_clean_suffix_test.go
@@ -1,0 +1,69 @@
+package gotreesitter_test
+
+import (
+	"strings"
+	"testing"
+
+	gotreesitter "github.com/odvcencio/gotreesitter"
+	"github.com/odvcencio/gotreesitter/grammars"
+)
+
+func TestCInitializerBoundsTransientRecoveryStacks(t *testing.T) {
+	source := cleanSuffixInitializerSource(40)
+	parser := gotreesitter.NewParser(grammars.CLanguage())
+	tree, err := parser.Parse(source)
+	if err != nil {
+		t.Fatalf("parse C initializer: %v", err)
+	}
+	t.Cleanup(tree.Release)
+	requireCompleteCInitializerTree(t, source, tree)
+	if got := tree.ParseRuntime().MaxStacksSeen; got > 8 {
+		t.Fatalf("peak stack count = %d, want at most 8", got)
+	}
+}
+
+func BenchmarkCInitializerAfterTransientRecovery(b *testing.B) {
+	source := cleanSuffixInitializerSource(40)
+	parser := gotreesitter.NewParser(grammars.CLanguage())
+	b.ReportAllocs()
+	b.SetBytes(int64(len(source)))
+	for range b.N {
+		tree, err := parser.Parse(source)
+		if err != nil {
+			b.Fatalf("parse C initializer: %v", err)
+		}
+		requireCompleteCInitializerTree(b, source, tree)
+		tree.Release()
+	}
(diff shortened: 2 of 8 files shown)


tsenart:
@odvcencio Ready for review at 1f652261. The full non-draft CI run is green, including race, C-oracle parity, merge census, and performance checks. The branch is current with main.

odvcencio (approved):
Approve, with one follow-up noted.

I verified this against the C runtime directly, using the grammar `grammars/languages.lock` pins (tree-sitter-c `ae19b676`, tree-sitter 0.25.3). I confirmed the oracle parser.c is byte-identical to upstream at that commit, and that the embedded blob fingerprints to it.

**Parity: neutral.** On a 1,567-file C corpus, main diverges from the C runtime on 454 files and this branch on 452. On a disjoint 951-file sample, both are 286. The branch repairs `spawn.c` and `pthread_mutex.c`, where main mis-parses a call as a declaration. It regresses `glxext.h`.

**Performance: a broad win.** Across 400 ordinary C files, total parse time drops from 1.18s to 0.80s and peak stacks drop from 253 to 36. On generated files the effect is much larger: `cue/parser.c` goes from 220s to 2.7s.

**Follow-up, not a blocker.** In `/usr/include/GL/glxext.h`, the `typedef unsigned long int uint64_t;` at bytes 32510-32545 gains a synthesized `MISSING ;`. The `uint64_t` token and the real `;` move outside the `type_definition`, so a consumer that walks type definitions no longer sees the name. Error state then propagates through six enclosing `preproc_ifdef`, `preproc_if`, and `preproc_elif` levels: the file reports 15 error-flagged nodes against 7 for both the C runtime and main. The result is deterministic across repeated runs. Main produces the same class of defect on other files, so this moves the boundary rather than adding a new failure mode. `grammars/clean_regression_pins_test.go` looks like the right place to pin it.

One negative result, recorded so nobody repeats it. `drainPendingForkStacks` and cap-one convergence share the `faithfulCapOneMergeEnabled` gate, so ending convergence also stops queued recovery forks from draining into the frontier. Ungating the drain repairs `glxext.h`, `min_setenv.c`, and `meteor-contest.c` exactly, but parity falls to 457 and the same 400 files take 3.16s, which is 2.7x slower than main. That trade is worse than the one this branch makes.

PR #739, opened 2026-08-17
fix(grammars): stop C inventing a comma in three-enumerator enums

odvcencio:
Fix a C parse defect that invents a comma inside a three-enumerator enum, and pin it.

## The defect

C tree-sitter parses the source below with no MISSING node and no error. gotreesitter inserted a zero-width `,` between the last enumerator and the closing brace and marked four ancestors as containing an error. Every other node already matched the C tree exactly.

    typedef enum { A = 0, B = 1, C = MAX } t;

Found by differential testing against the pinned C oracle over a 1,567 file corpus. `/usr/include/KHR/khrplatform.h` was the only file in that corpus where the oracle parsed completely clean and the Go engine invented a token. It reduced from 11,131 bytes to 42.

## Cause

Upstream never executes a repetition SHIFT at a conflict cell (`lib/src/parser.c`, `if (action.shift.repetition) break;`). It reduces, renumbers, and re-dispatches the same lookahead from the folded state.

gotreesitter implements that rule engine wide as `cRepetitionSkipConflictChoice`, and holds C out of it (`cRepetitionSkipOptOut`) because an eager fold before the first error reshapes what recovery later consumes on an error-dense corpus. The held-out C parser therefore forks at the cell. Convergence then fails in three steps: C forces cap-one merging, the GSS merge refuses because one branch is in demoted-linear form, and the cap-one depth tiebreak keeps the deeper unfolded branch. The parse dead-ends on `}` and recovery invents the comma.

Three enumerators is the only count that fails. Two never reach the conflict cell. With four, a later fork promotes the stack to a packed GSS, the merge succeeds, and the folded history survives.

The parse table matches upstream byte for byte, so this is a runtime and profile defect, not a grammargen defect.

## Fix

Add one `ConflictPolicyRepetitionReduce` row for `aux_sym_enumerator_list_repeat1` to the `c` runtime profile. This does not lift the opt-out. It carves out one symbol that is proven safe, which is what `haskell` already does with three rows for the same reason.

## Measurements

Corpus of 1,567 C files, both engines compared against the pinned C oracle, baseline and candidate both built from the same commit:

- divergences 452 to 449
- 3 files become byte-exact against the oracle: `khrplatform.h`, `rawmidi.h`, `xkb.h`
- 0 regressions

A second population: 23 files whose defects are normally hidden because `maybeReplaceRecoveredTreeWithForest` declines when the root does not start at byte 0. Re-measured with one byte of leading trivia so the repair cannot mask them: 2 fixed, 0 regressed.

Full suite matches the baseline failure set. `TestBuiltinCConflictPolicyAttachesWildcard` asserted C had exactly one policy row; it now asserts both rows, including the new kind and symbol.

## Pin

`TestCleanRegressionPinCEnumThreeValuedEnumerators` is self-healing in the style of the existing `regex any_character` pin: it passes once the invented comma disappears. It passes with this fix and skips with a KNOWN DIVERGENCE message without it.

### DIFF
--- a/cgo_harness/c_enum_missing_comma_parity_test.go
+++ b/cgo_harness/c_enum_missing_comma_parity_test.go
@@ -1,0 +1,49 @@
+//go:build cgo && treesitter_c_parity
+
+package cgoharness
+
+import (
+	"os"
+	"testing"
+
+	gotreesitter "github.com/odvcencio/gotreesitter"
+	sitter "github.com/tree-sitter/go-tree-sitter"
+)
+
+// TestCEnumThreeValuedEnumeratorsLeadingTriviaMatchesCReference locks the
+// reduced PR #739 witness against the pinned C parser. Leading trivia selects
+// the production GLR route, so this case must not rely on the forest fallback.
+func TestCEnumThreeValuedEnumeratorsLeadingTriviaMatchesCReference(t *testing.T) {
+	const src = "\ntypedef enum { A = 0, B = 1, C = MAX } t;"
+	source := []byte(src)
+	restoreForest := os.Getenv("GOT_GLR_FOREST") != "0"
+	gotreesitter.SetGLRForestEnabled(false)
+	t.Cleanup(func() { gotreesitter.SetGLRForestEnabled(restoreForest) })
+
+	cLang, err := ParityCLanguage("c")
+	if err != nil {
+		t.Fatalf("load locked C parser: %v", err)
+	}
+	cParser := sitter.NewParser()
+	t.Cleanup(cParser.Close)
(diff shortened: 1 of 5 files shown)


odvcencio:
Production-pinned C oracle receipt

- Added the exact leading-trivia enum witness.
- Forced the production route by disabling the candidate and forest routes.
- Compared symbols, fields, spans, points, flags, and the locked-C deep digest.
- The focused Docker oracle passed with digest 99fd4a6020bbd667a07e15148d6447e93f0d6d273d99392b9e915127b2b462a2.
- The enum pin and wildcard policy tests passed.
- The focused Swift helper regression family passed.
- A Sol read-only review approved the checkpoint.

PR #793, opened 2026-08-23
fix(parser): Restore generic recovery field parity

odvcencio:
## Decision

GO for publication. This patch closes only the recorded C locked-tree parity subproblem. Keep issue #454 open for the remaining field-report performance and recovery work.

## Semantic fix

- Keep absorbed visible leaves clean while the enclosing `ERROR` node still reports an error.
- Preserve field identifiers and field sources when hidden recovery nodes flatten.
- Route recovery splices through scratch buffers and apply recovered fields.
- Change only generic C recovery materialization and its focused parity test.
- Change no grammar registry or grammar blob.

## Correctness evidence

- The first old mismatch was `number_literal[0]` under `ERROR[2]` in the 140,288-byte witness.
- The five-size locked-C guard passed at 1, 4, 16, 64, and 137 KiB.
- Direct grammargen-to-C parity passed 20 of 20 cases.
- Incremental replace, insert, and delete trees matched fresh trees.
- The delete case retained `incremental_parse_memory_budget_full_retry`.
- The C real-corpus preset matched the base: 23 of 25 no-error cases and 20 of 25 deep-parity cases.
- The three known type divergences remained unchanged from the base.

## Memory evidence

The audit used three alternating base and candidate pairs in one-language Docker runs.

| Pair | Base RSS | Candidate RSS | Difference |
| --- | ---: | ---: | ---: |
| 1 | 566,680 KiB | 609,464 KiB | +42,784 KiB |
| 2 | 628,852 KiB | 606,980 KiB | -21,872 KiB |
| 3 | 593,812 KiB | 620,544 KiB | +26,732 KiB |

The base mean was 596,448 KiB. The candidate mean was 612,329 KiB, or 2.66% higher.
The earlier 1,089,364 KiB candidate result did not reproduce.

- Focused workload RSS was 857,624 KiB for base and 842,900 KiB for candidate without heap profiling.
- Paired heap-profile RSS was 858,400 KiB for base and 863,336 KiB for candidate.
- `alloc_space` was 1,180.79 MB for base and 1,167.62 MB for candidate.
- `inuse_space` was 129.30 MB for base and 128.66 MB for candidate.
- Arena field storage and field counts were equal in fresh and incremental-delete trees.
- No release-blocking memory regression reproduced. No Docker run hit an out-of-memory failure or timeout.

The standard 20-seed primary Go trio was not run because the candidate changes only C recovery paths.
The focused C workload is the relevant performance gate.

## Scope and validation

The patch changes exactly four files: `parser_recover_c.go`, `cgo_harness/issue454_c_compact_blocker_parity_test.go`, `docs/issue-454-compact-correctness-blocker.md`, and `CHANGELOG.md`.
The audited code and test patch SHA-256 is `71fdb2ab00f8f31e74b7e165f381c0856bd3720abdeb4d1556454d0cc75c50fa`.
The corrected four-file patch SHA-256 is `8b3e84d911ceaed861fc95d0a9ce466d5798bc69a5a5a2a972afda2939bfde65`.
Gofmt, direct no-cache Canopy, Markdown lint and parse, focused parser tests, the 1 KiB locked-C guard, and incremental edit guards passed in Docker.
The changelog and the blocker document record the proof and the memory tradeoff.

## Issue status

This patch closes only the recorded C locked-tree parity subproblem. Keep issue #454 open for the remaining field-report performance and recovery work.
Remaining work includes performance, memory-budget retry, scanner, flat-root, JavaScript, and other recovery findings.
The five-size locked-C guard and the replace, insert, and delete guards remain focused regression guards. Preserve the memory-budget fallback as a separate performance concern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -192,17 +192,27 @@ for tags and release notes while still in `0.x`.
 
 ### Fixed
 
-- Record the issue #454 compact-parser correctness blocker. A 1 KiB fresh Go
-  tree differs from locked C at
-  `/translation_unit/function_definition[0]/compound_statement[2]/ERROR[2]/number_literal[0]`.
-  Issue #454 reports C latency with correctness marked OK. This repository uses
-  an internal deterministic C witness. Go marks the `number_literal` as an
-  error, and C does not. The original source has 140,288 bytes. The edited
-  source has 140,287 bytes. The same difference persists at 4, 16, 64, and
-  137 KiB. Incremental Go equals fresh Go at 137 KiB, and the memory-budget fallback reports
-  `incremental_parse_memory_budget_full_retry`. The fresh-parse mismatch makes
-  retry selection unable to fix C parity. Status: NO-GO. Keep issue #454 open.
-  See [the issue #454 compact-parser correctness blocker receipt](docs/issue-454-compact-correctness-blocker.md).
+- Record the audited generic recovery fix for issue #454. The candidate diff
+  SHA-256 is
+  `71fdb2ab00f8f31e74b7e165f381c0856bd3720abdeb4d1556454d0cc75c50fa`.
+  `cAbsorbTokenIntoError` leaves visible absorbed leaves clean.
+  `cRecoverToState` preserves field metadata when it flattens hidden nodes.
+  The five-size Docker guard matches locked C at 1, 4, 16, 64, and 137 KiB.
+  The replace, insert, and delete guards match fresh parses. Direct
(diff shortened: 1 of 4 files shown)


odvcencio:
CI exposed deterministic Cobol and WGSL census changes and Cooklang digest changes from the broad leaf-error removal. The field-preserving splice is safe, but it does not fix issue #454 alone. No grammar-agnostic predicate exists. Reject the C-name guard. Keep issue #454 open.

PR #858, opened 2026-08-23
Archive the G3 EOF recovery work

odvcencio:
## Summary

- Preserve six exact G3 working blobs on a remote branch.
- Keep this branch separate from the active campaign.

## Validation

- Verified every staged blob against the original working blob.
- Ran the focused staged secret scan.
- Confirmed the root license exists.

## Status

This pull request is a storage checkpoint. Tests did not run. The work is not accepted.


odvcencio:
Closed by the 2026-08-23 archive triage. The payload already exists on main in equal or newer form. The archive branch keeps the full state. See the disposition spore spore.2026-08-23.cedar.archive-sweep-disposition in hypha://m31labs/gotreesitter.

PR #932, opened 2026-08-23
[archive] Preserve latest P25ay hotspot receipt

odvcencio:
Remote durability checkpoint only — this draft PR is durable storage, not acceptance or a merge recommendation.

Preserves the exact latest two-file P25ay state after the performance receipt moved following its v1 checkpoint, based on exact source HEAD `af9ded2b77b7828b12b1d2da7c9fff8dd5ca053b`. This is a fresh immutable v2 branch; v1 was not rewritten.

Validation performed before and after checkpoint:
- root `LICENSE` is tracked, non-empty, and MIT
- focused staged-patch gitleaks scan passed
- staged blob hashes matched the source worktree immediately before commit
- carrier commit parent equals the source HEAD
- local HEAD, upstream tracking ref, and live `git ls-remote` SHA are identical
- both remote tree blobs match the post-push source worktree byte-for-byte
- carrier is clean and the original worktree was not edited

This archival PR is not a merge recommendation. Tests and review determine acceptance separately. Refs #454.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,18 @@ for tags and release notes while still in `0.x`.
   size (RSS) run was justified.
   See `docs/perf-attribution.md` for exact identities, commands, and artifacts.
 
+- Recorded P25ay at base `af9ded2b77b7828b12b1d2da7c9fff8dd5ca053b`.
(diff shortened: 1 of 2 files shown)


odvcencio:

PR #936, opened 2026-08-23
archive: preserve final campaign prong C delta

odvcencio:
Remote durability checkpoint for the final quiescent prong C delta after its Buckley owner exited.

This draft is based on the already-preserved commit `d8fecafb8f73cb21b53a0dfdc7efe59c68807a1f` and adds the exact final `c3` blob. It is durable storage for review and is not merge acceptance.


odvcencio:

PR #938, opened 2026-08-23
archive: preserve campaign phase-two commit chain

odvcencio:
Remote durability checkpoint for the clean, immutable two-commit Buckley campaign chain.

This draft points at exact tip `9b1b0f0b83d7e91563ecbbb10733b709f948af0c` and therefore preserves its local-only parent `77e6fd1d302e4649796c72167ee8a5d2948b8907` without rewriting either commit. It is durable storage for review and is not merge acceptance.


odvcencio:

PR #941, opened 2026-08-23
archive: preserve final phase-two cohort revision

odvcencio:
Remote durability checkpoint for the final quiescent phase-two cohort revision after its Buckley owner exited.

This draft builds on exact immutable checkpoint `16bf78b9608f7d54d8fc99557459be6537999c01`. It is durable storage for review and is not merge acceptance.


odvcencio:

PR #942, opened 2026-08-23
archive: preserve final prong C provenance revision

odvcencio:
Remote durability checkpoint for the final quiescent prong C provenance revision after its Buckley owner exited.

This draft builds on exact prior checkpoint `f6afd383225b7a136c0cf7e9c3b7bf67a0427adb` and preserves the final `c2`/`c3` delta while retaining identical `c1`. It is durable storage for review and is not merge acceptance.


odvcencio:

PR #943, opened 2026-08-23
archive: preserve late phase-two census correction

odvcencio:
Remote durability checkpoint for the ownerless late phase-two ranking and census correction.

This draft builds on exact checkpoint `b45d0a495615b57a6b86415e411321e4d6fa4e2a`. It is durable storage for review and is not merge acceptance.


odvcencio:

PR #977, opened 2026-08-25
fix: Preserve C conflict version order

### LINKED ISSUE #972: c_sharp: collection initializer parses in 250s and returns root=ERROR (v0.51.0)
# c_sharp: collection initializer parses in 250s and returns root=ERROR (v0.51.0)

## Summary

On `v0.51.0` the `c_sharp` grammar takes **250 seconds** to parse an
11 KB / 211-line file of **valid** C#, and still returns `root=ERROR`.
Parse cost grows roughly exponentially with the number of entries in a
collection initializer of the form `{ typeof(T), Expr }`.

This is not the namespace-recovery OOM fixed in v0.20.2 (#98/#106):
RSS stays flat at ~400 MB while one core spins, so it's a time blowup,
not a memory one.

## Repro (synthetic, no external files)

```go
package main

import (
	"fmt"
	"strings"
	"time"

)

func source(entries int) string {
	var b strings.Builder
	b.WriteString("using System;\nusing System.Collections.Generic;\n\n")
	b.WriteString("namespace Repro\n{\n    internal static class Table\n    {\n")
	b.WriteString("        internal static readonly Dictionary<Type, object> Map = new Dictionary<Type, object>()\n        {\n")
	for i := 0; i < entries; i++ {
		b.WriteString("            { typeof(int), Formatter.Instance },\n")
	}
	b.WriteString("        };\n    }\n}\n")
	return b.String()
}

func main() {
	entry := grammars.DetectLanguageByName("c_sharp")
	pool := gotreesitter.NewParserPool(entry.Language(),
		gotreesitter.WithParserPoolTimeoutMicros(0)) // no timeout
	for _, n := range []int{4, 16, 32, 48, 64, 72} {
		src := []byte(source(n))
		t0 := time.Now()
		_, err := pool.Parse(src)
		fmt.Printf("%3d entries  %5d bytes  %7.3fs  err=%v\n",
			n, len(src), time.Since(t0).Seconds(), err)
	}
}
```

Output on `v0.51.0` (linux/amd64, go1.27.0):

```
  4 entries    425 bytes    0.032s  err=<nil>
 16 entries   1013 bytes    0.078s  err=<nil>
 32 entries   1797 bytes    0.530s  err=<nil>
 48 entries   2581 bytes    1.931s  err=<nil>
 64 entries   3365 bytes    5.093s  err=<nil>
 72 entries   3757 bytes    7.718s  err=<nil>
```

Parse time roughly doubles every 8 entries while the input grows by
~200 bytes. The synthetic plateaus near ~8.8s, so entry count alone
doesn't account for the full real-world cost below — but the scaling is
clear and reproducible from this snippet alone.

## Repro (real file — the worse case)

`BuiltinResolver.cs` from MessagePack-CSharp:

```
https://github.com/MessagePack-CSharp/MessagePack-CSharp
src/MessagePack/Resolvers/BuiltinResolver.cs   (11 KB, 211 lines)
```

Parsed via `NewParserPool(..., WithParserPoolTimeoutMicros(0))`:

```
DONE in 250.8s  root=ERROR
```

Two things stand out: the 250s, and that a **syntactically valid** file
returns `root=ERROR` rather than a parse tree. Its initializer entries
are heterogeneous — `typeof(Double?)`, `typeof(decimal?)`,
`new StaticNullableFormatter<decimal>(DecimalFormatter.Instance)` —
which appears to be worse than the uniform synthetic case.

Same repo, same grammar: 6 of 688 `.cs` files exceed a 30s cap, 58
exceed 1s, and the full set takes 438s.

## Environment

- gotreesitter `v0.51.0`
- go1.27.0, linux/amd64
- No timeout set (`WithParserPoolTimeoutMicros(0)`)

## What we did downstream

[ken](https://github.com/townsendmerino/ken) has removed `.cs` from its
structural-extraction language map for now. We can't apply a wall-clock
parse budget on our CLI path because it would make our index build
load-dependent, and our 64 KiB size guard doesn't help — the file is
11 KB. Happy to test a patch.

## Note on a red herring

Bisecting by truncating the file to N lines is misleading here:
truncated prefixes are invalid C#, so what you measure is
error-recovery cost, not this bug. The 89-line prefix parses in 7.9s
with `root=ERROR` for that reason. The synthetic above avoids the
confound by keeping every input syntactically complete.

odvcencio:
## Summary

- Keep the first non-repetition shift on the original conflict version.
- Fork the other actions in grammar order.
- Add host and locked-C tests for a valid C# collection initializer.

This fixes the valid generic object-creation tree divergence. It does not close #972.

Large-file performance and forest election remain open.

## Tests

- `go test ./grammars -run '^TestCSharpGenericObjectCreationInCollectionInitializer$' -count=1`
- `go test . -run '^(TestConflictReduceFrontierSameHeaderFirstReduceStaysLive|TestConflictReduceFrontierDepthChangingSameReduceDoesNotFakeDuplicate|TestConflictReduceFrontierRuntimeReduceTerminalPreservesReduceBranch|TestCSharpDesignerStyleBlockStaysBounded|TestCSharpCJKPartialClassesStayBounded|TestCSharpNamespaceRecoveryStaysBoundedDottedBitorArg)$' -count=1`
- `go test . -run 'Conflict' -count=1`
- `GTS_PARITY_ALLOW_HOST=1 go test . -tags treesitter_c_parity -run '^TestCSharpGenericObjectCreationInCollectionInitializerParity$' -count=1 -v -timeout 45s`
- Narrow changed-file Canopy checks.
- `git diff --check`

Docker artifact: `/tmp/gts-csharp-972-artifacts/20260825T023112Z-issue972-publication-cgo-regression`

Change hash: `259057335e702903ae9cbc9b0c95e172c123d747364010bb51ca492029e4bdaa`

Refs #972

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,9 @@ for tags and release notes while still in `0.x`.
 
 ### Correctness
 
+- Preserved C conflict version order for generic object creation in C#
+  collection initializers.
+
 - Added a default-off Core link provenance sidecar. It stores finalized
   drop-cohort reference indexes by LinkID. Compact routing and admission stay
   unchanged.
--- a/cgo_harness/csharp_invocation_statement_parity_test.go
+++ b/cgo_harness/csharp_invocation_statement_parity_test.go
@@ -74,3 +74,9 @@ func TestCSharpGenericBaseListParity(t *testing.T) {
 	tc := parityCase{name: "c_sharp", source: string(src)}
 	runParityCase(t, tc, "generic-base-list", src)
 }
+
+func TestCSharpGenericObjectCreationInCollectionInitializerParity(t *testing.T) {
+	src := []byte("class C { object x = new D() { { typeof(T?), new F<T>(G.I) }, }; }\n")
+	tc := parityCase{name: "c_sharp", source: string(src)}
+	runParityCase(t, tc, "generic-object-creation-in-collection-initializer", src)
+}
--- a/grammars/csharp_collection_initializer_regression_test.go
+++ b/grammars/csharp_collection_initializer_regression_test.go
@@ -1,0 +1,37 @@
+package grammars
+
+import (
+	"testing"
+
+	gotreesitter "github.com/odvcencio/gotreesitter"
+)
+
+func TestCSharpGenericObjectCreationInCollectionInitializer(t *testing.T) {
+	source := []byte("class C { object x = new D() { { typeof(T?), new F<T>(G.I) }, }; }")
+	language := CSharpLanguage()
+	parser := gotreesitter.NewParser(language)
+	tree, err := parser.Parse(source)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tree.Release()
+
+	root := tree.RootNode()
+	if root == nil {
+		t.Fatal("collection initializer returned no root")
+	}
+	if root.HasError() {
+		t.Fatalf("collection initializer did not parse cleanly: %s", root.SExpr(language))
+	}
+	creation := findFirstNamedDescendantWhere(root, language, "object_creation_expression", func(node *gotreesitter.Node) bool {
+		return node.Text(source) == "new F<T>(G.I)"
+	})
+	if creation == nil {
+		t.Fatalf("generic object creation is absent: %s", root.SExpr(language))
+	}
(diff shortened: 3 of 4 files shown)


odvcencio:
The global conflict-order change alters accepted-prefix behavior and canonical Go and Cooklang trees. The merge census also changes. I am closing this implementation instead of moving its ratchets. The cap-only issue fix will proceed separately.

PR #1014, opened 2026-08-31
Avoid header copies in compact dispatch

odvcencio:
## Summary

- Read active scheduler headers through slice pointers.
- Remove the redundant reload after an in-place recovery update.
- Preserve the dispatch order and all recovery decisions.

## Performance

The Erlang issue 984 benchmark improved from 36.60 microseconds to 34.90 microseconds.
This is a 4.65 percent improvement with p=0.011 across 20 interleaved pairs.
Bytes and allocation counts did not change.

The required 20-seed primary benchmark trio showed no significant change.
The single-byte edit result had p=0.324.
The no-edit result had p=0.683.
The full-parse result had p=0.947.

## Correctness

- The focused compact dispatch and recovery tests pass in Docker.
- The full default root suite passes in Docker in 155.6 seconds.
- Erlang parity passes 25 of 25 cases for all three parity levels.
- The focused race and checkptr runs pass in Docker.
- The default vet path passes.

No run timed out or exceeded its memory limit.

### DIFF
--- a/parsercore_phase0_driver.go
+++ b/parsercore_phase0_driver.go
@@ -6477,7 +6477,8 @@ func (s *diagnosticParserCoreGenericScheduler) dispatchPassActive() (*diagnostic
 	raggedRelexNoActionHeads := 0
 	var raggedRelexWitness Token // the first ragged-end relex this pass, for the decline detail
 	raggedRelexHeaderIndex := 0  // that same relex's own header index, for the decline receipt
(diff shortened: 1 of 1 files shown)


odvcencio (commented):
Review complete. I found no merge blocker.\n\nThe header pointer remains valid because the scan loop does not resize the header slice. The removed reload was redundant.\n\nAll required correctness and performance checks pass.

PR #1015, opened 2026-09-01
Add the compact S4 stack-summary recovery fork

odvcencio:
## Summary

Add the first fail-closed S4 recovery fork to the compact scheduler.

Keep the absorbing version and the ancestor-recovered version in C creation order.

Reject ambiguous stack-summary candidates before they mutate a parser head.

Keep the new capability disabled for all runtime profiles.

## Validation

- Run focused S4, S3, S5, and core tests in Docker.
- Run the full root and internal core packages in Docker.
- Compile the disabled-core build tag in Docker.
- Match all 25 HTML parity cases across three checks.
- Match all 25 JavaScript parity cases across three checks.
- Confirm the parser header remains 224 bytes.

## Scope

This change does not certify a runtime grammar.

A later change will add the exact six-version recovery condense before broader recovery routing.

### DIFF
--- a/admission_census.go
+++ b/admission_census.go
@@ -511,6 +511,22 @@ func admissionCensusStopDecline(scheduler *diagnosticParserCoreGenericScheduler)
 			detail += "]"
 		}
 	}
+	if scheduler.work.StackSummaryRecoveryForks != 0 {
+		marked := 0
+		openRegions := 0
+		for index := range scheduler.headers {
+			if scheduler.headers[index].isRecoveryLineage() {
+				marked++
+			}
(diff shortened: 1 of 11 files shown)


odvcencio (commented):
Review complete. I found no merge blocker.\n\nS4 remains disabled for every runtime profile. It requires one header, authenticates a unique pop path, and uses an atomic recovery mutation.\n\nAmbiguous candidates fail closed. Both competing lineages receive recovery marks before acceptance arbitration.\n\nAll required checks pass, including the disabled-core build.

PR #1018, opened 2026-09-01
Preserve recovery competition through ambiguity

odvcencio:
## Summary

- Preserve recovery costs through ordinary conflict and reduction forks.
- Condense recovery versions with C pairwise comparison and the six-key cap.
- Track per-version recovery groups and visible-node baselines.
- Count aliases, open error regions, and ERROR_REPEAT progress with C semantics.
- Add ownership cleanup and path telemetry.

## Proof

- Focused Stage 3 and lifecycle tests pass in both modified packages.
- The exact JavaScript recovery tree gate passes with one compact selection.
- HTML Docker parity passes 25 of 25 cases for every parity level.
- Both parser-core build modes compile.
- The 20-seed randomized benchmark loop completes without a crash or memory stop.
- Full parse averages 10.33 ms and five allocations across the 20 seeds.
- Both incremental probes remain allocation-free.

No certified grammar witness currently increments RecoveryAmbiguityForks. Synthetic C-semantic falsifiers cover the fork, comparison, ordering, ownership, and cleanup paths. The closest locked JavaScript witness conflicts before recovery and still matches C.

This pull request contains no unrelated performance changes.

### DIFF
--- a/internal/parsercorephase0/core.go
+++ b/internal/parsercorephase0/core.go
@@ -1122,9 +1122,9 @@ type SubtreeView struct {
 }
 
 // MaterializationSubtreeView is a callback-scoped borrowed view of one compact
-// subtree. Children aliases compact arena storage and must not be retained or
-// mutated by the visitor. The copying Subtree diagnostic API remains the
-// stable inspection surface.
+// subtree. Children and Aliases reference compact arena storage. A visitor
+// must not retain or mutate them. The copying Subtree diagnostic API remains
+// the stable inspection surface.
 type MaterializationSubtreeView struct {
 	Symbol            Symbol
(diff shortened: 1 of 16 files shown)


odvcencio:
Review fix add406fa closes the first CI oracle failure. Pairwise condensation reached one active version but kept recovery isolation enabled, so the next pass rejected its sole valid winner. The fix clears the winner marker and isolation at that transition.

Proof:

- The exact html_log_7 locked C witness passes.
- The complete 20-witness T3 oracle suite passes.
- The public missing-versus-absorb route test passes.
- Focused Stage 3 and lifecycle tests pass.
- The compile-out mode remains clean.

PR #1051, opened 2026-09-02
add(parsercore): bind owned lexer snapshots to scanner identity

odvcencio:
## Summary

Bind each compact owned-lexer snapshot to its external scanner and exact grammar blob.

- Reject incomplete identities before snapshot publication.
- Reject identity drift before restore, publication, or equivalent snapshot reuse.
- Bind scanner order adapters to the target grammar blob.
- Preserve restrictive scanner capabilities across adapters.
- Keep legacy checkpointed scanners outside the new identity contract.
- Skip accepted headers during ownership seeding.
- Reject ownership activation while a recovery region remains open.

This pull request keeps compact admission unchanged. It supplies the authenticated ownership prerequisite for recovery, merging, and edit reuse.

## Proof

The focused tests prove these conditions:

- Two lexer versions can consume different widths and select the C-equivalent Scala tree.
- A cloned snapshot preserves its complete identity.
- Restore and publication reject grammar identity drift without target mutation.
- Snapshot construction rejects an incomplete identity.
- Equivalent snapshot lookup rejects identity drift.
- Adapters bind the source scanner implementation to the target grammar blob.
- Nested adapters bind the final authenticated target.
- Legacy adapters remain optional.
- Accepted headers receive no owned cursor.
- Open recovery blocks a new ownership tranche.
- The scheduler header remains 224 bytes.

## Validation

- Focused tests pass with and without `gts_parsercorephase0`.
- Generated-language scanner tests pass.
- `actionlint` passes.
- The isolated package certificate passes in Docker.
- The Scala locked-C oracle passes in the same Docker run.
- Docker reported no timeout and no out-of-memory event.

Artifact: `harness_out/docker/20260903T050516Z-pr1051-package1-review-clean`

The persistent CI job runs the same identity, lifecycle, adapter, and Scala oracle witnesses.

## Graduation boundary

This is the only active core implementation pull request.

Do not start physical graph-head merging before this pull request merges. Track that next package in issue #1057.

Track full Scala grammar certification in issue #1067. Its generated-parser mismatches remain outside this package-one gate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -882,7 +882,7 @@ jobs:
             --timeout 20m \
             --run '^(TestCompactT3OracleAdjudication|TestCompactT3JavaScriptRecoveryProfileCharacterization|TestCompactT3JavaScriptRecoveryRouteSeedOracleParity|TestCompactT3JavaScriptRecoveryMutationDifferential|TestCompactJavaScriptS5RecoveryMutationDifferential|TestCompactRouteLockedCExceptions)$'
 
-      - name: Compact per-version lexer Scala C-oracle witness
+      - name: Compact per-version lexer package-one witnesses
         run: |
           GTS_PARITY_MODE=smoke \
           GTS_PARITY_C_REF_BUILD_CACHE=0 \
@@ -896,7 +896,7 @@ jobs:
             --test-parallel 1 \
             --c-ref-build-jobs 1 \
             --timeout 20m \
-            -- "cd /workspace/cgo_harness && go test . -tags 'treesitter_c_parity gts_parsercorephase0' -run '^TestCompactPerVersionLexerScalaCOracle$' -count=1 -parallel 1 -timeout 20m -v"
+            -- "cd /workspace && go test . -tags gts_parsercorephase0 -run '^(TestDiagnosticParserCoreVersionLexerSnapshot(BindsCheckpointIdentity|RejectsIncompleteCheckpointIdentity)|TestDiagnosticParserCoreOwnedLexer(SeedingSkipsAcceptedHeader|ActivationRejectsOpenRecoveryRegion)|TestExternalScannerOrderAdapter(PreservesOptionalCapabilities|KeepsLegacyCheckpointIdentityOptional|RequiresTargetGrammarCheckpointIdentity|BindsAuthenticatedTargetAcrossNestedAdapters))$' -count=1 -parallel 1 -timeout 20m -v && cd /workspace/cgo_harness && go test . -tags 'treesitter_c_parity gts_parsercorephase0' -run '^TestCompactPerVersionLexerScalaCOracle$' -count=1 -parallel 1 -timeout 20m -v"
 
   # merge-event-census-cgo runs stage M0 of spec.merge-time-election.v1: the
   # census that counts stack-merge events on both sides and breaks production's
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -882,6 +882,22 @@ jobs:
             --timeout 20m \
             --run '^(TestCompactT3OracleAdjudication|TestCompactT3JavaScriptRecoveryProfileCharacterization|TestCompactT3JavaScriptRecoveryRouteSeedOracleParity|TestCompactT3JavaScriptRecoveryMutationDifferential|TestCompactJavaScriptS5RecoveryMutationDifferential|TestCompactRouteLockedCExceptions)$'
 
+      - name: Compact per-version lexer Scala C-oracle witness
+        run: |
+          GTS_PARITY_MODE=smoke \
+          GTS_PARITY_C_REF_BUILD_CACHE=0 \
+          bash cgo_harness/docker/run_parity_in_docker.sh \
+            --label ci-compact-per-version-lexer-scala \
+            --memory 8g \
+            --cpus 2 \
+            --pids 4096 \
+            --gomemlimit 6GiB \
+            --goflags -p=1 \
+            --test-parallel 1 \
+            --c-ref-build-jobs 1 \
+            --timeout 20m \
+            -- "cd /workspace/cgo_harness && go test . -tags 'treesitter_c_parity gts_parsercorephase0' -run '^TestCompactPerVersionLexerScalaCOracle$' -count=1 -parallel 1 -timeout 20m -v"
+
   # merge-event-census-cgo runs stage M0 of spec.merge-time-election.v1: the
   # census that counts stack-merge events on both sides and breaks production's
   # refusals down by gate. Every cgo parity lane in this file runs a FIXED
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,11 @@ for tags and release notes while still in `0.x`.
 
 ### Correctness
 
+- Bind identity-bearing external scanner checkpoints to compact per-version
+  lexer snapshots. Reject incomplete or changed identities before restore,
+  publication, or sidecar sharing. Forward identities through scanner order
+  adapters. Preserve legacy scanner behavior.
+
 - Graduated the Markdown inline smoke route with four exact, compact-only
   conflict policies bound to the built-in grammar blob. Three repetition
   reductions require at least two live compact headers. The HTML entry row
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,9 +101,10 @@ for tags and release notes while still in `0.x`.
 ### Correctness
 
 - Bind identity-bearing external scanner checkpoints to compact per-version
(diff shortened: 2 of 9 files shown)


odvcencio:
Additional verification:

- GitHub Actions run [33684907377](https://github.com/odvcencio/gotreesitter/actions/runs/33684907377) passes every check, including the isolated Rust race gate.
- The locked Scala C per-version lexer oracle passes.
- The primary trio ran with 20 shuffle seeds at 750ms per seed. Compared with exact `origin/main`, full parse improved 5.10%, incremental single-byte edit improved 3.18%, and no-edit incremental improved 3.96%. Allocations stayed unchanged.
- The branch is mergeable. One independent review remains required.

odvcencio:
Follow-up boundary reduction:

- Scala `((y)->` activates two owned lexer widths, drops one version, and rejoins the shared lexer correctly.
- The compact route then falls back at EOF in state 716.
- The production C-equivalent trace explores unfinished reductions and missing insertions. It grows to five versions, condenses to three, and selects the missing-group lineage at cost 610.
- Docker trace artifact: `harness_out/docker/20260902T221824Z/container.log`.
- Trace SHA-256: `5eded90745c71e7a41d90b0e7509d891e176e72bd04719bf4c4314f9c404b5d3`.

This result does not expand PR #1051. The identity patch remains admission-neutral. The next core falsifier is EOF reduction and missing-insertion competition after owned-width arbitration.

odvcencio:
Independent adversarial review at head 704f0680 (build, vet, and the three new tests pass locally; branch is 8 commits behind main). Fifteen findings, ranked. The first three bear on the package-one gate; the rest are follow-ups.

**Blocking for the gate**

1. `external_scanner_adapter.go:79`. The adapter's `CheckpointIdentity` forwards the reference grammar's blob SHA as the identity of the adapted target language. Two grammars adapted from one reference therefore share one identity. Reproduced: at merge-base the adapted arena stayed `valid=false` and identity matching failed closed; at PR head the arena is stamped with the reference identity and cross-language checkpoint-authenticated reuse was observed with 6 reused subtrees. Latent today because `SqlExternalScanner` is language-bound and returns before the adapter, but the mechanism is wrong. Fix at construction: `AdaptExternalScannerByExternalOrder` has `targetLang` and `Language.GrammarBlobSHA256` in hand; re-bind the identity to the target.
2. `parsercore_phase0_driver.go:4800` and `:5295`. Two production identity lookups still use direct `.(ExternalScannerCheckpointIdentityProvider)` assertions instead of the new helper. Because the adapter now satisfies that exported interface, these unchanged lines flip from an unconditional early return to a live path for any adapter over an identity-bearing scanner, and `activateVersionLexerOwnershipAtRagged` becomes reachable for adapted languages with zero scheduler tests covering an adapter. The repo now carries two mechanisms (the concrete-type switch at `external_scanner_checkpoint_capability.go:36` and the forwarding method) that agree only by mirroring. Route both sites through the helper and keep one mechanism.
3. `parsercore_phase0_per_version_lexer_snapshot_internal_test.go:1`. Both new tests live behind `gts_parsercorephase0` and no CI lane executes them: the untagged lanes do not compile the file, and the tagged lanes use anchored `-run` allow-lists that exclude them. The phase-0 half of the PR ships with zero CI-executed coverage. The gate requires telemetry or tests proving the intended path ran in CI, not only in the cited `/tmp` Docker runs.

**Adjacent, pre-existing, worth fixing in the same forwarder block**

4. `external_scanner_adapter.go:110`. The forwarder omits `ErrorTreeIncrementalReuseExternalScanner`, the one capability whose absent default is permissive. Adapted markdown and html reuse subtrees over error trees their scanners refused: embedded markdown reports `ReuseUnsupported=true`, adapted markdown reports `ReusedSubtrees=9`. Forward `SupportsIncrementalReuseFromErrorTree`.

**Design and correctness follow-ups**

5. `per_version_lexer_state.go:105`. `diagnosticParserCoreVersionLexerCheckpointIdentity` duplicates `externalScannerCheckpointIdentityStatus` with the opposite polarity for `valid`. A verified equivalent replacement in terms of the existing helper is three lines.
6. `per_version_lexer_state.go:39`. `checkpointIdentityValid` always equals `checkpointIdentityRequired` (all 16 struct literals checked), so the `!valid` disjuncts at `:297` and `driver.go:5070` are dead and the capability-changed guard at `:293` is unreachable. Drop the field or document the invariant at the field.
7. `per_version_lexer_state.go:292`. Identity is recomputed six to eight times per header per owned election; `driver.go:4991` is always followed by the identical validation at `:5043`. SQL only, modest cost; delete `:4991`.
8. `..._internal_test.go:697`. `t.Fatalf` dereferences `snapshot` inside the `snapshot == nil ||` branch; the nil case panics instead of reporting.
9. `..._internal_test.go:756`. The rejection test asserts only `err != nil`; any of five earlier constructor errors makes it pass. Assert the identity message.
10. `..._internal_test.go:728`. The identity test never drives a drifted identity through `publishVersionLexerSnapshot`, unlike its three sibling rejection tests.
11. `CHANGELOG.md:105`. "Reject ... before sidecar sharing" overstates the sidecar path: a drifted identity misses the cache and the caller publishes a fresh snapshot bearing it; the request fails only later via the before-snapshot. No counter records identity rejections. Reword, or add a counter.

**Style**

12. `per_version_lexer_state.go:34`: a 27-word comment describing a legacy scheduler path that does not exist.
13. `per_version_lexer_state.go:153`: `clone` repeats the 14-field positional literal; the file's own `out := *s` idiom one function above avoids it.
14. `driver.go:5062`: `diagnosticParserCoreVersionLexerCheckpointIdentity(nil)` as a zero-value factory; a `var` or early return says the same.
15. PR body: the graduation-funnel sentence lists six items inline; the prose standard wants a vertical list.

Recommendation: fix 1 through 4 before merge, land 5 through 11 in the package-two PR or a small follow-up, and treat 12 through 15 as author's choice.

odvcencio:
Package-one review closure at 8d11ce4e:

- Bind adapted scanner identity to the exact target grammar blob.
- Keep an identity-bearing adapter required when the target blob is unavailable.
- Resolve both scheduler identity reads through the shared provider helper.
- Preserve the inner scanner restriction for error-tree incremental reuse.
- Skip accepted headers during owned-lexer seeding.
- Reject ownership activation while a recovery region remains open.
- Run all focused identity and lifecycle witnesses in the persistent CI job.

The isolated Docker certificate passed all eight root witnesses and the Scala C oracle.

Artifact: `harness_out/docker/20260903T050124Z-pr1051-package1-final-forwarders`

No out-of-memory event or timeout occurred. GitHub Actions run 33717272661 now validates the pushed head.

odvcencio:
The remaining review cleanup is complete at 230f120e:

- Reuse the shared checkpoint identity status helper.
- Add a publication test for identity drift.
- Split the nil assertion from identity checks.
- Require the exact incomplete-identity error.
- Make snapshot cloning resilient to future fields.
- Correct the changelog claim.
- Rewrite the pull request body to use the required vertical lists.

The repeated identity lookup is performance-only. I deferred it because this correctness pull request excludes unrelated performance work.

The refreshed Docker certificate remains green: `harness_out/docker/20260903T050516Z-pr1051-package1-review-clean`.

odvcencio:
Final package-one merge receipt for 230f120e:

- GitHub Actions run 33717515048 completed successfully.
- Every compile, parity, fuzz, race, freshness, and performance budget job passed.
- The package-one job executed all eight focused root witnesses.
- `TestCompactPerVersionLexerScalaCOracle` executed and passed.
- The isolated Docker certificate passed without timeout or out-of-memory failure.
- The final independent Luna review found no P0, P1, or P2 blocker.
- The current branch has no merge conflict with `main`.

Pull request #1051 is ready for one independent approval. Branch protection remains its only blocker. No policy bypass was used.

odvcencio:
Re-review at head 230f120e (CI run 33717515048: 46 success, 4 skipped).

The four items I marked blocking are addressed:

1. Identity binding: `AdaptExternalScannerByExternalOrder` now stamps the adapted target with its own `GrammarBlobSHA256`, so two targets adapted from one reference no longer share an identity.
2. Dual mechanism: no direct `.(ExternalScannerCheckpointIdentityProvider)` assertion remains in `parsercore_phase0_driver.go`; both production sites go through the helper.
3. CI execution: a persistent Docker job runs the two tagged snapshot-identity tests and the Scala per-version lexer C-oracle witness.
4. Capability forwarding: the adapter implements `SupportsIncrementalReuseFromErrorTree` and returns false when the inner scanner omits the interface, which is the fail-closed direction.

Test hardening from findings 8 to 10 is partly in (identity message assertion and a publication-path check). Findings 5 to 7 and 11 to 15 remain open and are fine to carry into package two.

From the independent-review side this PR is ready. The only remaining blocker is the branch-protection approval, which a single-collaborator repository cannot satisfy from a second account.

PR #1076, opened 2026-09-04
fix(parser): preserve missing recovery edit dependencies

### LINKED ISSUE #1060: Compact incremental recovery must preserve missing-leaf padding and lookahead
## Summary

Preserve the C missing-leaf padding and lookahead dependency before compact incremental admission.

This is a package 5 incremental-integration gate. It does not block issue #1057.

## C contract

`ts_parser__handle_error` resets the lexer at the stack position before it creates a missing leaf.

It passes these values to `ts_subtree_new_missing_leaf`:

- Padding from the stack position to the lexer token-end position.
- Lookahead bytes from the current lookahead subtree.
- A zero content size.

The padding can be nonzero when included ranges move the lexer to the next range.

The lookahead bytes remain relevant without included ranges. `ts_subtree_edit` uses them when it invalidates a tree after an edit.

## Current compact gap

`Core.MissingLeaf` accepts only one byte offset.

It stores a zero-width range and no padding or lookahead dependency.

Its documentation says the C padding is always zero. That statement is valid only for the current fresh route without included ranges.

The public Go `Node` also has no lookahead-byte field. The existing production recovery path does not preserve this C metadata.

## Package boundary

Issue #1057 stays limited to the fresh UTF-8 route.

That route already declines these modes:

- Included ranges.
- Incremental parsing.
- Retry parsing.

Issue #1057 must correct the route-specific documentation. It must keep a direct zero-padding proof for its admitted route.

This issue owns the metadata needed before package 5 admits compact incremental recovery.

## Required design

Keep `subtreeRecord` at 44 bytes.

Use sparse storage unless measurement proves a smaller representation.

Preserve these values for each missing leaf:

- The stack byte and point position.
- The included-range padding byte and point extent.
- The lookahead byte dependency.

Thread the dependency through materialization and tree edits.

Do not make the visible missing-node content range nonzero. Padding positions the zero-width node.

Reject stale sidecar references during validation. Roll back sidecar writes with the Core transaction.

## Required proof

Add a locked C included-range witness with nonzero missing-leaf padding.

Assert these results:

- Exact tree shape.
- Exact node ranges and points.
- Exact missing and error flags.
- Exact incremental changed ranges after edits in the padding.
- Exact invalidation after edits in the lookahead dependency.
- Unchanged reuse outside the dependency.

Add a no-included-range witness with a multi-byte lookahead.

Edit each byte in that lookahead. Match the locked C incremental result for every edit.

Pin the sparse sidecar size, ownership, rollback, and retention behavior.

Run one grammar at a time in Docker. Keep correctness separate from performance.

## Source evidence

C creates the missing leaf in `src/parser.c` near the `ts_subtree_new_missing_leaf` call.

C stores the metadata in `src/subtree.h` and `src/subtree.c`.

C uses the lookahead dependency in `ts_subtree_edit` when it chooses edited children.

odvcencio:
## What does this PR do?

Preserve C-equivalent padding and lookahead dependencies for zero-width missing leaves.

Thread exact dependency data through compact storage, materialization, clones, tree edits, and arena telemetry. Close #1060.

## Why this approach?

Use sparse sidecars in both compact and public arenas. This keeps the 44-byte compact subtree record unchanged.

Capture the farthest internal and external lexer frontier. Preserve it across failed scans, retries, synthetic tokens, and included ranges.

Fail closed when a dependency receipt is stale or cannot transfer.

## Correctness

- [x] Focused package and unit tests cover the touched behavior.
- [x] Both compact build modes pass focused tests.
- [x] The real C incremental witness passes for all three lookahead bytes.
- [x] The locked C oracle covers padding edits, lookahead edits, ranges, points, flags, and changed ranges.
- [x] The C grammar Docker run remains at the baseline: 20/25 exact and 23/25 error-free.
- [x] The isolated root package and compact-core package sweep passes.
- [x] No Docker run reported an out-of-memory event.

Evidence:

- Focused code gate: harness_out/docker/20260904T044023Z-issue1060-final-gate
- C oracle: harness_out/docker/20260904T044148Z-issue1060-c-oracle
- Arena accounting: harness_out/docker/20260904T044231Z-issue1060-accounting
- Package sweep: harness_out/docker/20260904T045209Z-issue1060-package-gate-clean
- C grammar: harness_out/docker/20260904T050027Z-diag-c_lang

The package sweep excluded TestCompactRouteLifecycleAllowsStaleReceiptOutsideCurrentLineage. Its linked-worktree Git path is not available inside the Docker mount. The focused route tests passed.

## Performance

- [x] Performance is not the goal of this correctness PR.
- [x] The C grammar run stayed bounded at 1,206,848 KiB maximum resident set size.
- [x] Correctness and performance evidence remain separate.

## Maintainability

- [x] Keep missing metadata sparse.
- [x] Keep the compact subtree record at 44 bytes.
- [x] Account for sidecar capacity in arena totals and reports.
- [x] Avoid unrelated performance or release work.
- [x] Add no dependencies.

## Self-review

- [x] Reviewed the full diff and all ownership paths.
- [x] Verified rollback, reset, retention, cloning, and materialization.
- [x] Verified internal, external, retry, invalid UTF-8, and included-range frontiers.
- [x] Ran independent Luna reviews for the implementation, oracle, and gate coverage.

## Due diligence

- [x] Checked the C missing-leaf construction and edit contract.
- [x] Used Canopy for scoped symbol and reference checks.
- [x] Kept this PR within package 5 incremental integration.

### DIFF
--- a/admission_route_equality_leaf_tiling_test.go
+++ b/admission_route_equality_leaf_tiling_test.go
@@ -328,6 +328,8 @@ func TestCompactRouteDeclinesRetiredRecoveryWithoutAliasRule(t *testing.T) {
 
 func TestCompactRouteCompetesMissingInsertionWithErrorAbsorb(t *testing.T) {
 	t.Cleanup(func() { grammars.PurgeEmbeddedLanguageCache() })
+	gts.EnableArenaBreakdown(true)
+	t.Cleanup(func() { gts.EnableArenaBreakdown(false) })
 	lang := grammars.HtmlLanguage()
 	if !lang.CompactStrategy2ErrorRegionCertified || !lang.CompactMissingTokenInsertionCertified {
 		t.Fatal("the HTML artifact lacks complete compact recovery certification")
@@ -351,6 +353,10 @@ func TestCompactRouteCompetesMissingInsertionWithErrorAbsorb(t *testing.T) {
 	}
 	if !routeEqualityTreeContainsMissing(tree.RootNode()) {
 		t.Fatal("the selected recovery tree contains no missing terminal")
+	}
+	breakdown, recorded := tree.ArenaBreakdown()
+	if !recorded || breakdown.MissingNodeDependencyCount == 0 || breakdown.MissingNodeDependencyBytesAllocated == 0 {
+		t.Fatalf("compact recovery did not materialize missing-dependency telemetry: recorded=%t breakdown=%+v", recorded, breakdown)
 	}
 }
 
--- a/arena.go
+++ b/arena.go
@@ -124,6 +124,7 @@ type nodeArena struct {
 	compactCheckpointLeafSlabs      []compactCheckpointLeafSlab
 	compactCheckpointLeafSlabCursor int
 	finalChildSidecars              []finalChildSidecar
+	missingNodeDependencies         []missingNodeDependencyEntry
 	nodeFieldMetadataSlabs          []nodeFieldMetadataSlab
 	nodeFieldMetadataSlabCursor     int
 
@@ -556,6 +557,7 @@ func (a *nodeArena) reset() {
 	a.resetRawShapeChildSlabs()
 	a.resetRawShapeHashCache()
 	a.resetFinalChildSidecars()
+	a.resetMissingNodeDependencies()
 	a.resetCompactCheckpointLeafSlabs()
 	a.resetNodeFieldMetadataSlabs()
 	a.resetChildSlabs()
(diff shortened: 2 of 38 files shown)


odvcencio:
Final review fixes landed in 41b58403. The patch now merges maximum lookahead frontiers only for equal GLR token surfaces, while strict per-version ownership remains distinct. Relex snapshots now capture and restore external lookahead state. Focused root, compact-tag, and C-oracle Docker tests pass with no OOM: harness_out/docker/20260904T052122Z-issue1060-review-fixes.

odvcencio:
CI failure resolution: the external frontier is transient per lex call. Commit 26224222 keeps capture and restore, but excludes the field from persistent snapshot identity. This prevents false scheduler-owner splits while strict token frontier ownership remains intact. The exact 322-test phase0 CI allow-list now passes in Docker: harness_out/docker/20260904T053553Z-issue1060-phase0-ci-fixed. The focused seven-test ownership gate also passes: harness_out/docker/20260904T053513Z-issue1060-owner-fix.

odvcencio:
Affected-grammar check: Swift completes 25/25 without errors and matches 22/25 exact trees on both #1076 and origin/main. The same three baseline type divergences remain. Branch artifact: harness_out/docker/20260904T053920Z-diag-swift. Baseline artifact was compared before its temporary worktree was removed. Branch maximum RSS was 1,984,088 KiB; no OOM occurred.

PR #1081, opened 2026-09-04
Keep focused grammar tests offline

odvcencio:
## Change

Remove the global TestMain download. Focused grammar tests no longer clone Markdown before execution.

Gate 0 and Gate 1 now provision the pinned fixtures explicitly. Each setup has a 90-second deadline and a five-second termination grace period. Both fixture files must exist and contain data.

Keep the existing warning-and-skip policy. Add no workflow, matrix, required gate, or report framework.

## Verification

- Offline Docker reproduced the baseline download attempt during TestJSONNormalize.
- Updated focused tests and compile checks passed without creating a corpus directory.
- Explicit Markdown import parity passed: 1/1 exact.
- Five setup behaviors passed in offline Docker.
- Shell syntax, formatting, and diff checks passed.
- Independent review found and removed a proposed CI download that would not exercise the fixtures.

Markdown-inline generation exceeded the diagnostic container's 4 GiB limit. This change does not modify generation. Gate 0 retains its existing 8 GiB limit.

### DIFF
--- a/cgo_harness/docker/run_grammargen_gates.sh
+++ b/cgo_harness/docker/run_grammargen_gates.sh
@@ -208,6 +208,10 @@ gate0() {
 set -eo pipefail
 cd /workspace
 
+if ! timeout --signal=TERM --kill-after=5s 90s bash grammargen/testdata/grammar_parity_setup.sh; then
+  echo "WARNING: Markdown corpus setup failed; import parity may skip missing fixtures." >&2
+fi
+
 echo "--- Gate 0a: grammargen unit tests ---"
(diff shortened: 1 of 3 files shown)


odvcencio (commented):
Independent Astra-medium subagent review completed at c87f929aa9dfc0e7d565619bcebdac46647d07df.

No blockers found.

- Gate 0 and Gate 1 provision the corpus before the import parity tests.
- Setup failures remain warnings; missing fixtures explicitly skip.
- The timeout terminates the setup process group.
- Corpus validation requires both nonempty grammar files.
- Both changed shell scripts pass syntax checks.

Ordinary focused tests no longer download Markdown. Missing import fixtures can skip outside the explicit gates.

This records the independent agent review required by the project workflow. It is not a separate GitHub account approval.

PR #1082, opened 2026-09-04
Preserve precedence-bearing hidden reductions

### LINKED ISSUE #1067: Fix Scala generated-grammar parity

odvcencio:
## Change

Preserve precedence-bearing mixed hidden reductions and the unary hidden reductions that feed them. Do not flatten their alternatives into sibling rules.

The old transformation removed Scala's constructor reduction and allowed compound types to consume separate parents or a class body. Preserving only the parent reduction was insufficient: flattening its child could still change the conflict state.

The closure follows unary hidden references only. It terminates on cycles and includes generated hidden auxiliaries. Unrelated pure pass-through expansion stays unchanged.

No Scala-specific hint, LR arbitration change, repeat rewrite, parser routing change, or performance optimization remains.

## Regression coverage

- Pin static, dynamic, zero, negative, and associative precedence under field and alias wrappers.
- Assert the normalized reduction graph and exact C-derived leaves for a small A&B grammar.
- Cover wide choices, configured preservation, cycles, and generated hidden bridges.
- Compare twelve Scala annotation, inheritance, template, and generic-parent cases with the locked reference.
- Disable compact routing in generator comparisons to keep attribution separate.

## Verification

Docker checks passed for the focused normalization, conflict, JSON, calculator, and GLR tests. All twelve Scala controls passed.

Strict JavaScript corpus parity passed: 25/25 error-free, S-expression, and deep matches.

The same locked Scala corpus improved against HEAD 2c19a78c:

| Result | HEAD | This change |
|---|---:|---:|
| Error-free samples | 11/25 | 13/25 |
| Exact deep trees | 7/25 | 9/25 |

The Scala grammar remains locked to 97aead18d97708190a51d4f551ea9b05b60641c9. Strict Scala certification still fails. This PR does not close #1067 or graduate the compact route.

Independent review checked closure termination, metadata propagation, skipped roots, generated auxiliaries, and exact sample identities. No new failing sample identity appeared in the final comparison.

Reproduce the focused case with the locked Scala checkout mounted at /tmp/grammar_parity/scala:

```sh
go test ./grammargen -run '^(TestFlattenHidden|TestHiddenMixedPassthrough|TestScalaConstructorOwnershipMatchesLockedC)' -count=1
```

Run that command through cgo_harness/docker/run_parity_in_docker.sh. Use run_single_grammar_parity.sh javascript --require-parity for the independent grammar gate.

### DIFF
--- a/grammargen/flatten_hidden_test.go
+++ b/grammargen/flatten_hidden_test.go
@@ -1,8 +1,190 @@
 package grammargen
 
 import (
+	"reflect"
+	"strings"
 	"testing"
+
+	"github.com/odvcencio/gotreesitter"
 )
+
+func TestFlattenHiddenMixedPassthroughPreservesReductionPrecedence(t *testing.T) {
+	tests := []struct {
+		name string
+		wrap func(*Rule) *Rule
+	}{
+		{"static", func(r *Rule) *Rule { return Prec(7, r) }},
+		{"explicit_zero", func(r *Rule) *Rule { return Prec(0, r) }},
+		{"negative", func(r *Rule) *Rule { return Prec(-1, r) }},
+		{"left", func(r *Rule) *Rule { return PrecLeft(0, r) }},
+		{"right", func(r *Rule) *Rule { return PrecRight(0, r) }},
+		{"dynamic", func(r *Rule) *Rule { return PrecDynamic(3, r) }},
+		{"dynamic_zero", func(r *Rule) *Rule { return PrecDynamic(0, r) }},
+		{"field", func(r *Rule) *Rule { return Field("value", PrecLeft(7, r)) }},
+		{"alias", func(r *Rule) *Rule { return Alias(PrecLeft(7, r), "value", true) }},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			for _, outer := range []bool{false, true} {
+				g := NewGrammar("preserve_reduction_precedence")
+				g.Define("document", Seq(Sym("_value"), Str("!")))
+				compound := Seq(Str("("), Sym("number"), Str(")"))
+				rule := Choice(tt.wrap(Sym("number")), compound)
+				if outer {
+					rule = tt.wrap(Choice(Sym("number"), compound))
+				}
+				g.Define("_value", rule)
+				g.Define("number", Pat(`[0-9]+`))
+				want := cloneRule(g.Rules["_value"])
+				wantDocument := cloneRule(g.Rules["document"])
+
+				got := flattenHiddenChoiceAlts(g, nil)
+				if !reflect.DeepEqual(got.Rules["_value"], want) {
+					t.Errorf("outer=%t: changed the precedence-bearing reduction", outer)
+				}
+				if !reflect.DeepEqual(got.Rules["document"], wantDocument) {
+					t.Errorf("outer=%t: bypassed the precedence-bearing reduction", outer)
+				}
+			}
+		})
+	}
+}
+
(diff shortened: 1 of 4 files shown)


odvcencio (commented):
Independent Astra-medium subagent review completed at dffbf1873f375902b7c4413d23574195c8d4c9ea against 835e1bdc7833e4b40feb78e651d29d6733f7a8e3.

No merge blockers found.

- Precedence protection occurs before width and configuration skips.
- Child closure terminates because each symbol enters its queue once.
- Removing protected children from flattenMap preserves reductions across shared callers.
- Existing flattening controls and new precedence regressions passed independently in Docker.
- All 12 Scala constructor controls passed against locked grammar 97aead18d97708190a51d4f551ea9b05b60641c9.

The review found no unrelated changes. Candidate routing stays disabled for generator attribution. These controls do not certify all Scala grammar output.

This records the independent agent review required by the project workflow. It is not a separate GitHub account approval.

Issue #1063, opened 2026-09-03
Graduate the compact parser through seven gated packages


odvcencio:
## Objective

Graduate the compact parser and retire the legacy parser without losing correctness, reuse, or release safety.

Keep one core implementation pull request active. Run adversarial tests, certification, and maintenance in parallel.

## Current state

The compact engine serves eligible full parses by default. Legacy parsing still supplies unsupported modes and compact fallbacks.

| Package | Merged implementation | Remaining scope |
| --- | --- | --- |
| 1. Per-version lexer state | #1050, #1051 | Certify additional grammar cases. |
| 2. Physical graph-head merging | #1073 | Certify additional recovery topologies. |
| 3. Recovery through ambiguity | #1074 | Certify additional recovery and ambiguity cases. |
| 4. End-of-file recovery | #1075, #1085 | Extend non-EOF growth and unsupported version transitions. |
| 5. Incremental integration | #1076, #1077, #1078, #1083, #1084, #1085, #1088 | Correct token-invariant reuse, then extend recovery reuse and scanner support. |
| 6. Grammar certification | #1064 remains open | Compare shipped grammar output against locked C, one grammar at a time. |
| 7. Performance and release | #1065 remains open | Complete candidate-specific safety and performance validation. |

Do not describe stages 1 through 5 as complete legacy retirement.

## Latest merged package

PR #1088 merged at `7db6e20a`. The compact engine can reuse bounded nonterminals inside an edited top-level item.

The smallest witness preserves the parameter list during `func a(){_=1}` to `func a(){_=x}`.

Dependency coverage includes discarded lexer probes and reduction lookahead. Reuse rejects unknown provenance and changed source dependencies.

The implementation preserves lazy children during edits and invalidates dependencies in borrowed arenas.

Independent review, focused race tests, and exact Go/C comparisons pass. All 51 pull-request CI checks passed or skipped.

The randomized comparison retained zero allocations for both incremental controls. Full-parse allocation increased by 38.46%; timing differences remained inconclusive under host variation.

That allocation cost remains an optimization target. This package does not claim general scanner support or legacy retirement.

PR #1085 also remains merged. Its bounded Go EOF recovery executes on version-owned compact state without normalization.

Recovery edits currently perform a full compact reparse and report zero reused subtrees. Recovery reuse remains unfinished.

The earlier ordinary-leaf cleanup stays unpublished because forced-legacy framing remains unsafe to change.

## Active implementation: #1087

[...]

PR #1085, opened 2026-09-05
Run Go EOF recovery on version-owned compact state

odvcencio:
## Summary

Run bounded Go recovery on version-owned compact lexer state.

- Advance the absorbing version to EOF before its missing-token sibling consumes the unexpected token.
- Recover to an ancestor and retain the competing EOF tree until C-equivalent selection completes.
- Apply native acceptance splicing without result normalization.
- Authenticate production aliases without changing existing recovery profiles.
- Retry incremental recovery with a full compact parse after borrowed-node cleanup.
- Report zero reuse for that fallback. Count discarded attempts in the profiled API.

The exact Go grammar profile enables this route. Other recovery capabilities remain separate.

## Scope

This package supports EOF after the first absorbed token. It does not graduate general recovery.

Prior shared recovery, unsupported lexical reentry, and an unmodeled six-version transition decline before mutation. Recovery SelectedStore construction also declines.

The incremental fallback preserves correctness by reparsing. It does not claim subtree reuse.

## Correctness

Compare against main `4c9f1d40177498a684174c858b53e0fd90d8b8aa` and the locked C runtime.

- Original matrix: 1,199 edits; all 226 compact results match C. Ten previous mismatches become exact.
- Expanded matrix: 4,524 edits; all 480 compact results match C. Fifteen previous mismatches become exact.
- Every other complete tree and per-node error flag matches main.
- Tracked cases cover suffix recovery, package aliases, parameters, loops, strings, repairs, and copy isolation.
- Canonical Go fresh and incremental C parity passes.
- JavaScript retains baseline results across 5,267 S5 cases and 2,954 T3 cases.
- Scala recovery and version-owned lexer witnesses retain exact C parity.
- Focused Docker race tests cover recovery, lexer ownership, cancellation, allocation accounting, and memory limits.
- The emergency build passes.

The broader admission race run reports three inherited failures. Each reproduces identically on the unchanged main tree:

- `TestAdmissionCandidateElmHighlightDirect` reports the same digest difference.
- `TestAdmissionCandidateExactExternalPayloadCorpus/kotlin` reports the same blocker difference.
- `TestAdmissionCandidateExactExternalPayloadCorpus/perl` reports the same blocker difference.

## Performance

Run each benchmark in 20 separate processes with explicit shuffle seeds and `GOMAXPROCS=1`. Use `750ms`, allocation reporting, and CPU 18.

The adjacent baseline and candidate runs use the same bounded Docker configuration.

| Benchmark | Main | Candidate | Result |
| --- | --- | --- | --- |
| Full parse | 14.26 ms | 14.07 ms | No significant change; p=0.242 |
| Single-byte edit | 1.235 µs | 1.239 µs | No significant change; p=0.857 |
| No edit | 3.026 ns | 2.893 ns | -4.41%; p<0.001 |

Allocation counts remain unchanged. Both incremental cases retain zero bytes and zero allocations per operation.

The initial contended candidate run was slower. That result did not reproduce in the isolated adjacent comparison. No performance implementation changes were made.

## Campaign

Advance #1063 toward compact parser graduation. Keep the remaining grammar and release gates open.

### DIFF
--- a/admission_switch_candidate.go
+++ b/admission_switch_candidate.go
@@ -47,6 +47,7 @@ func newAdmissionCandidateRunner(p *Parser) (*parserCoreFreshFullRunner, error) 
 		return nil, errors.New("admission candidate route: parser has no language")
 	}
 	allowRecoverEOF := compactRecoverEOFArtifactConfigured(p.language)
+	allowOwnedEOFRecovery := p.language.CompactOwnedEOFRecoveryCertified
 	options := DiagnosticParserCorePrefixOptions{
 		ReceiptMode:                    DiagnosticParserCoreReceiptSummary,
 		MaxTokens:                      1 << 24,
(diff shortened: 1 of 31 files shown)


odvcencio:
Independent Astra review completed for `2ab5c2953f615bb7c516040be46833cde69b7fb8`.

The reviewers found no unresolved findings after the fixes. Review ownership remained separate from implementation ownership.

The review corrected these defects before publication:

- Preserve rollback on scheduler and materialization failures.
- Pop a prior ERROR before constructing the replacement region.
- Preserve ERROR extras during native acceptance splicing.
- Exclude shared recovery from the new owned EOF admission.
- Preserve existing JavaScript alias admission while adding the native Go path.
- Replace growing linear alias lookup with bounded, accounted storage.
- Count tokens and allocations from discarded incremental attempts.

JavaScript and Scala retain their locked C results. The emergency build passes.

Three broader admission race failures reproduce identically on the unchanged main tree. They concern Elm's digest and Kotlin/Perl blocker assertions.

This review covers bounded Go EOF recovery. It does not certify general recovery, recovery reuse, or recovery SelectedStore construction.

odvcencio:
The follow-up commit isolates two diagnostic frontier fixtures from recovery retries. It changes four test lines only. The original publication, decline, and digest assertions remain unchanged. Docker reproduced both CI failures before the fix, then passed all G18D6b/c tests and the Go recovery controls. Runtime and benchmark results are unchanged.

odvcencio:
An additional variable-width probe compared 280 fresh and incremental outputs across five Go prefixes. All 50 compact results match the locked C trees, including numeric tokens, identifiers, Unicode α, and an open quote. These 50 correct baseline mismatches. The other 230 complete trees and error flags match main unchanged. Unsupported operators, invalid UTF-8, comments with errors, and lexical reentry retain conservative fallback. No runtime or test changes followed this probe.

Issue #1087, opened 2026-09-05
Authenticate earlier lexer dependencies before token-invariant reuse


odvcencio:
## Failure

The same-width token shortcut can return an incorrect tree when an earlier token inspected the edited byte.

Use this grammar:

```text
word = /a|ab+c/
group = "(" word
statement = group letter* ";"
letter = /[bcd]/
```

Edit `(abbbd;` to `(abbbc;` at byte 5.

The edited `letter` keeps its symbol and span. The earlier `word` changes from ending at byte 2 to ending at byte 6.

The shortcut rescans only the edited leaf and returns the old topology.

This reproduces on main's runtime at `21148a5c`. Local evidence: `/tmp/gts-compact-incremental-tests/20260905T035028Z-nested-reuse-dfa-samewidth-base/container.log`.

The nested-reuse work found this inherited defect. Its width-changing counterpart passes on main and needs separate new dependency coverage.

## Required change

- Retain authenticated terminal dependency coverage, or reauthenticate preceding terminals within a bounded scan.
- Reject unknown coverage unless reauthentication succeeds.
- Rescan earlier tokens whose examined ranges intersect the edit. Compare their symbols and spans.
- Preserve the ordinary `1→2` shortcut and its benchmark coverage.

Nonterminal receipts alone do not establish this proof. They omit terminal-only streams and include reduction lookahead beyond lexical dependencies.

## Regression coverage

- Start from compact and forced-legacy trees.
- Compare incremental results with fresh results through repeated edits.
- Preserve token-invariant edits that do not change earlier token boundaries.
- Include a flat terminal stream without an intermediate nonterminal.

This is the next core correctness package after bounded nested reuse. Refs #1063.

odvcencio:
The defect also reproduces with the real Go grammar.

Change `package p\nfunc f(){_=0eX}\n` to `package p\nfunc f(){_=0e1}\n` through one same-width edit.

The incremental shortcut returns the previous error tree with zero reparse work and zero new nodes.

Fresh Go parsing and the locked C parser produce a valid floating-point literal without an error.

The failure reproduces with forced legacy parsing and compact routing enabled. This does not prove compact admission of the malformed initial source.

A bounded Docker run also reproduced related hexadecimal, binary, and octal suffix cases. The committed regression will use the smallest decimal case.

Local evidence: `/tmp/gts-compact-incremental-tests/20260905T060943Z-go-token-invariant-numeric-probe/container.log`.

[...]

PR #1088, opened 2026-09-05
Authenticate compact reuse inside edited declarations

odvcencio:
## Changes

Preserve an unchanged parameter list during `func a(){_=1}` to `func a(){_=x}`.

The compact route now admits bounded nonterminals inside an edited top-level item. It preserves the existing top-level reuse path.

Reuse requires exact parser states, unchanged source dependencies, a matching fresh token, and authenticated node ownership.

The scheduler retains the farthest byte inspected by lexer probes and reduction lookahead. Arena-owned receipts survive safe copies and reject invalid edits.

Receipts use synchronized, coordinate-bound storage. Position-changing edits discard shifted receipts; column-changing copies do not transfer them.

Publish receipts only for nodes that the nested selector can borrow. Authenticate their final ranges and points against the raw derivation.

An interval index restricts edit invalidation to relevant receipts. Repeated edits preserve unaffected receipts, including receipts in shared arenas.

The index uses immutable coordinates and an explicit live bit. A constrained budget selects exact scan-based invalidation without exceeding the limit.

Node and token layouts remain unchanged. Arena and scheduler budgets include the new storage. Reset drops retained receipts.

Scheduler reset clears and retains at most 64 KiB of pointer-free dependency scratch. It drops larger buffers and preserves budget accounting.

## Review-driven regressions

- `a+b ;` to `a+b *c;` must preserve C's multiplication precedence.
- A discarded longer DFA match must retain dependencies beyond the accepted token span.
- Alternate whitespace probes must retain both inspected frontiers, regardless of which token wins.

The first two tests failed against the initial implementation. The JavaScript precedence test uses the locked C parser.

## Scope

The new route rejects recovery, ambiguous producer history, unknown dependency receipts, and dependencies beyond available source bytes.

Scanners must satisfy the existing stateless contract: local lookahead and parser-provided valid symbols only. The Go scanner satisfies that contract.

Contextual token repairs remain excluded until their source dependencies are recorded. Generated zero-width sentinels and unproved synthetic tokens also decline.

Only physical EOF permits a certified external zero-width token. Its beyond-source dependency prevents borrowing a subtree that contains it.

Existing top-level reuse remains unchanged. This change does not retire the legacy parser or claim general scanner support.

## Verification

Focused tests cover exact trees, errors, pointer identity, parent links, copied trees, released source trees, scanner capabilities, and inspected lexer ranges.

Race tests cover shared arenas, repeated edits, empty index ranges, budget rejection, storage reset, and interrupted dependency capture.

Full CI exposed eager child materialization during `Node.Edit` and a nonstandard memory-budget diagnostic. Both now have focused fixes.

The edit walk follows lazy entries into borrowed arenas without creating public children. A new regression checks this ownership boundary directly.

The standard stop-control error preserves the established memory-budget diagnostic. All affected tests pass with the race detector and the diagnostic build tag.

Follow-up review found no actionable issues. Go incremental and certification comparisons still match the locked C parser.

Follow-up artifacts: `/tmp/gts-compact-incremental-tests/20260905T060343Z-nested-ci-tagged-lazy-budget`.

The first benchmark rejected a prototype: single-byte edits became 30.5 times slower, and full-parse allocation increased by 179%.

Candidate-only publication, indexed invalidation, and bounded scratch reuse replace that prototype.

The final comparison alternates baseline and candidate for each of 20 seeds. It reverses their order on even seeds.

Each process uses `GOMAXPROCS=1`, `-count=1`, `-benchtime=750ms`, and `-benchmem` through `scripts/run_randomized_benchmarks.sh`.

| Benchmark | Baseline time | Candidate time | Baseline allocation | Candidate allocation |
| --- | ---: | ---: | ---: | ---: |
| Full parse | 18.73 ms | 18.88 ms | 399.0 KiB | 552.5 KiB |
| Single-byte edit | 1.616 µs | 1.767 µs | 0 B | 0 B |
| No-edit parse | 3.684 ns | 3.757 ns | 0 B | 0 B |

Host load produced 15–18% timing variation. Benchstat found no significant timing difference, but this does not prove equal performance.

Full-parse allocation increased by 38.46%, with 0.18% more allocations. This bounded correctness cost remains an optimization target, not a speed claim.

All 40 benchmark processes completed. Docker reported no out-of-memory failure or timeout. Both incremental controls retained zero allocations per operation.

Final local artifacts:

- `/tmp/gts-compact-incremental-tests/20260905T052845Z-nested-retained-scratch-final`: focused race and Go/C checks.
- `/tmp/gts-compact-incremental-tests/20260905T050151Z-nested-index-geometry-javascript`: JavaScript C comparison.
- `/tmp/gts-compact-incremental-tests/20260905T053224Z-nested-retained-paired-trio`: paired benchmark campaign.

The earlier 5,723-edit comparison found no changed output or route classification. Selective publication subsequently tightened admission; final focused checks cover those changes.

The existing Go `same_line_length_change` canonical case also fails on the unchanged baseline. Its scanner-quiescence admission limitation remains separate.

Issue #1087 records an inherited same-width shortcut defect. This package does not claim to fix that distinct terminal-dependency problem.

Refs #1063.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,6 +166,8 @@ jobs:
           TestAdmissionSwitchParseIncrementalDoesNotCountFullRoute
           TestCompactIncrementalExecution.*
           TestCompactBorrowed.*
+          TestCompactNestedReuse.*
+          TestCompactReuseDependency.*
           TestAdmissionSwitchParseProductionWhenOff
           TestAdmissionSwitchParseRoutesCandidateWhenOn
           TestAdmissionSwitchRecoveryDoesNotLeakCompactTree
@@ -1283,7 +1285,7 @@ jobs:
             --cpus 2 \
             --gomemlimit 3GiB \
             --wall-timeout 10m \
-            -- "cd /workspace/cgo_harness && go test . -tags 'treesitter_c_parity gts_parsercorephase0' -run '^(TestGoCompactIncrementalExecution.*Parity|TestStage6GoCompactCertification)$' -count=1 -parallel 1 -timeout 5m -v"
+            -- "cd /workspace/cgo_harness && go test . -tags 'treesitter_c_parity gts_parsercorephase0' -run '^(TestGoCompactIncrementalExecution.*Parity|TestGoCompactNestedReuse.*LockedC|TestStage6GoCompactCertification)$' -count=1 -parallel 1 -timeout 5m -v"
 
       - name: Compact recovery incremental JavaScript witnesses
         run: |
@@ -1299,7 +1301,7 @@ jobs:
             --test-parallel 1 \
             --c-ref-build-jobs 1 \
             --timeout 20m \
-            -- "cd /workspace/cgo_harness && go test . -tags 'treesitter_c_parity gts_parsercorephase0' -run '^(TestCompactRecoveryJavaScriptIncrementalCOracle|TestCompactRecoveryJavaScriptInsertionRegression)$' -count=1 -parallel 1 -timeout 20m -v"
+            -- "cd /workspace/cgo_harness && go test . -tags 'treesitter_c_parity gts_parsercorephase0' -run '^(TestCompactRecoveryJavaScriptIncrementalCOracle|TestCompactRecoveryJavaScriptInsertionRegression|TestJavaScriptCompactNestedReuseLookaheadLockedC)$' -count=1 -parallel 1 -timeout 20m -v"
 
       - name: Compact recovery incremental YAML witness
         run: |
--- a/arena.go
+++ b/arena.go
@@ -125,6 +125,11 @@ type nodeArena struct {
 	compactCheckpointLeafSlabCursor int
 	finalChildSidecars              []finalChildSidecar
 	missingNodeDependencies         []missingNodeDependencyEntry
+	compactReuseDependencyMu        sync.RWMutex
+	compactReuseDependencies        map[*Node]compactReuseDependency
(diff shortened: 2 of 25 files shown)


odvcencio:
Independent review completed with an Astra subagent at medium reasoning effort.

The review covered:

- Scheduler dependency capture and discarded lexer probes.
- Candidate publication, raw ranges, source points, and projection changes.
- Indexed invalidation, shared arenas, copies, and borrowed ownership.
- Memory accounting, bounded scratch retention, reset, and interrupted execution.
- Exact C regressions and continuous integration coverage.

The reviewer found no unresolved correctness issue in the final implementation.

The root agent independently reviewed the precedence regression written by the reviewer. The reviewer excluded that test from its own signoff.

Focused race and Go/C tests passed after the final retention change. Performance results appear in the pull request description.

This review does not claim general scanner support, legacy-parser retirement, or a fix for the separate issue #1087.

odvcencio:
The independent Astra review also covers follow-up commit `0ab483207d09e0442be6372de5a82213ef67fac5`.

It found no actionable issue in the non-materializing ownership walk or the standard memory-budget diagnostic.

Focused race tests passed, including the borrowed-node regression and every failure from the previous CI run.

Go incremental comparisons and grammar certification match the locked C parser. The emergency legacy build also passes the edit tests.

The full CI run completed with all 51 checks passed or skipped: https://github.com/odvcencio/gotreesitter/actions/runs/33948929548.

The separate incremental wrong-tree defect remains tracked in #1087. This change does not claim to fix it.

PR #709, opened 2026-08-12
fix(apex, parser): fix apex class_literal election parity with c oracle

odvcencio:
## Summary

Aligns Apex GLR derivation selection with the locked C oracle for `Type.class`. The grammar declares both `class_literal` and `field_access` as live parallel paths. This branch adds a per-stack DFA relex to prevent lexical starvation and refines the accept-time tie-breaker to match C behavior.

## Changes

- Add a per-stack DFA relex before killing a no-action stack. Starved forks now receive their required token type instead of failing due to shared lexer state.
- Skip raw-shape fallback in accept-time comparison for certified languages. The branch-order signal now prefers the non-forked derivation to match the C oracle.
- Update documentation for normalizeApexClassLiteralAccess. The function now serves only the GSS-forest experimental fast path.
- Add regression tests for production, compact, and incremental routes. Include negative controls and explicit forest-route guards.
- Update parity harness and materialization census to reflect the corrected tree shape.

## Testing

- Run `bash cgo_harness/docker/run_single_grammar_parity.sh apex` to verify C oracle parity.
- Run `go test ./grammars -run TestApexClassLiteralElection -v` to execute the native regression suite.
- Run `go test ./parser_result_test -run TestMaterializationSubpassCensus -v` to validate census updates.

## Related Issues

- Refs #601

## Review Focus

Focus on the per-stack relex logic in parser.go and the certification gate in parser_result.go. Verify that apex parity tests pass locally before review.

PR #1099, opened 2026-09-07
fix(parser): route fresh parses through the production engine by default

odvcencio:
## Summary

The compact route became the default for fresh full parses in v0.52.0. Issue #454 then measured 2 to 3 times slower clean full parses, lost incremental reuse, and a non-terminating Go delete on large editor fixtures. This change restores the production route as the default, keeps the compact route opt-in until graduation, and fixes a production-route GLR regression the compact default had hidden.

## Changes

- Flip the process default for eligible fresh full parses from the compact route to the production route. An unset or unrecognized GTS_ADMISSION_CANDIDATE now resolves to the production route; opt in with GTS_ADMISSION_CANDIDATE=1, SetAdmissionCandidateRouteDefault(true), or the per-Parser override.
- Add test-binary init hooks in the root, cgo_harness, grammars, and roottest packages so the certification and parity suites keep exercising the compact route when the environment variable is unset.
- Halt a GLR stack at a no-action point when a sibling stack already shifts or accepts the lookahead, before the previous-shift recovery runs. This fixes the C++ consecutive storage-class specifier regression that the compact default masked since v0.49.0.
- Add a C++ production-vs-compact route parity test pinning the five storage-class specifier witnesses, and a package test documenting the compact-route opt-in.

## Testing

- go test . -run 'TestAdmissionSwitchEnvVarContract|TestPackageTestsOptIntoCompactRoute' -count=1
- go test . -run TestCppConsecutiveStorageClassSpecifiersProductionRouteMatchesCompact -count=1
- go test ./... -run '^$' -count=1 and go vet ./...
- go test . -count=1 with GTS_ADMISSION_CANDIDATE unset and with GTS_ADMISSION_CANDIDATE=0 (only the pre-existing TestResultCompatibilityRetiredCommitProvenance failure remains on main)
- go test ./grammars ./parser_result_test ./roottest/... -count=1 with the variable unset and with GTS_ADMISSION_CANDIDATE=0
- GOMAXPROCS=1 go test -tags gts_parsercorephase0 . -run '^(TestAdmission|TestCompact|TestDiagnosticParserCoreSelectedStore)' -count=1
- go run ./cmd/issue454bench go 137 delete

## Related Issues

- Refs #454
- Refs #709

## Review Focus

Review the GLR halt condition glrSiblingAcceptsLookahead for interaction with existing multi-stack recovery, and confirm the compact-route opt-in semantics stay consistent for any opt-out value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,42 @@ for tags and release notes while still in `0.x`.
 
 ## [Unreleased]
 
+### Fresh parse route default
+
+- Route eligible fresh full parses through the production engine by default.
+  The compact candidate route is now opt-in. Select it with
+  `GTS_ADMISSION_CANDIDATE=1`, `SetAdmissionCandidateRouteDefault(true)`, or
+  the per-Parser override. An unset or unrecognized value keeps the production route.
+- Issue [#454](https://github.com/odvcencio/gotreesitter/issues/454) compared
+  v0.52.0 with v0.48.1 on 137 KiB editor fixtures. Clean full parses ran 2 to
+  3 times slower on 41 of 48 grammars. Seven grammars lost most incremental
+  reuse. A single-byte Go delete did not terminate.
+- Pull request [#631](https://github.com/odvcencio/gotreesitter/pull/631)
+  removed the 64 KiB size decline, so large files reached the compact route
+  for the first time. The compact route was already the process default.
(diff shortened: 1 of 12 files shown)


odvcencio:
Superseded by #1101. The owner chose to keep the compact route as the default and repair it rather than make it opt-in. The C++ sibling-halt fix, the harness, and the route-default report carry over to #1101.

PR #1107, opened 2026-09-09
fix(bash): preserve whitespace after empty assignments

### LINKED ISSUE #1106: [parser]: Bash: spurious missing word in negated commands with an empty environment assignment
### Report type

Incorrect parse or tree

### gotreesitter version or commit

v0.52.0

### Grammar or language

Bash

### Minimal reproduction

Minimal

```shell
! A=x B= command
```

Real world:

```shell
if ! GOBIN="$bindir" GOFLAGS= $command; then
```

### Expected behavior

Expected: a complete tree with HasErrorOrMissing() == false. The reference C parser produces:

```
(program
  (negated_command
    (command
      (variable_assignment
        name: (variable_name)
        value: (word))
      (variable_assignment
        name: (variable_name))
      name: (command_name (word)))))
```

In particular, B= is an assignment with an empty value, and command remains the command name.

### Actual behavior and fresh-parse comparison

ParseStrict error: <nil>
HasErrorOrMissing: true

The reduced input produces a missing word node at zero-based byte offset 16.

### Environment

| Component                                          | Version / revision                         |
| -------------------------------------------------- | ------------------------------------------ |
| [Unswell](https://github.com/stokaro/unswell)                                            | `599ea7f8fe54df7041eb45e55ba98176cefe721a` |
| gotreesitter release                               | `v0.52.0`                                  |
| Additional tested upstream commit                  | `a50f15320a0ed4f2209d0122a5e90e18c7e7f27c` |
| Locked `tree-sitter-bash` revision                 | `a06c2e4415e9bc0346c6b86d401879ffb44058f7` |
| Go                                                 | `1.27.1`                                   |
| Platform                                           | Darwin ARM64                               |
| Bash                                               | GNU Bash `3.2.57`                          |
| C compiler                                         | Apple Clang `21.0.0`                       |
| Python binding used to load the C reference parser | `tree-sitter 0.25.2`                       |

### Optional diagnostics

Two source-level observations may help:

BashExternalScanner already handles EMPTY_VALUE by returning an external token without advancing when the lookahead is whitespace, EOF, ;, or &. Empty assignments are not generally unsupported. Tracing that token around B= and comparing the selected parser/scanner states with C would be a useful next diagnostic step, not a confirmed root cause.

The bashMustParseNoError helper in bash_parse_regression_test.go checks only HasError(). tree_health_test.go explicitly verifies that HasError() does not detect a descendant missing node, while HasErrorOrMissing() does. A regression test for this case should use the latter and check the tree structure against C.

### Checklist

- [x] I reduced this to a minimal reproduction or explained why I could not.
- [x] I removed secrets, private source, and other sensitive data from this report.

denisvmedia:
Fixes #1106.

`! A=x B= command` must leave `B` empty and preserve `command` as the command name. The opening-parenthesis probe consumed the separating whitespace before the scanner reached `EMPTY_VALUE`. Some inputs produced a missing word. Others produced an incorrect assignment without an error flag.

Preserve whitespace when `EMPTY_VALUE` is valid. The existing scanner then emits its zero-width token. The change adds one condition and leaves grammar data, token identities, and parser recovery unchanged.

Add shared inputs and blackbox tests:

- Check assignment values, command names, byte ranges, and `HasErrorOrMissing()`.
- Cover negation, multiple assignments, tabs, CRLF, Unicode, substitutions, redirects, pipelines, case bodies, arrays, and quoted values.
- Keep incomplete syntax invalid.
- Compare valid inputs with the locked C grammar across strict, production, and raw Go parsing.
- Include the new C comparison in the existing CI job.

Validation used a dedicated Linux Docker container with two CPUs and an 8 GiB memory limit. No repository-wide host sweep ran.

- Before the fix: 12 of 18 new cases failed.
- With the fix: all 18 cases and available Bash regressions passed.
- The locked C comparison passed for 16 valid inputs across three routes. It checks symbols, fields, byte/point spans, flags, child order, and deep digests.
- Existing Bash fresh/incremental parity and C-reference regressions passed.

- The final fix passed Bash race tests on Go 1.27.0.
- Bash tests passed with `CGO_ENABLED=0` on Go 1.22.0, the module's declared minimum compiler.
- Actionlint passed for the changed workflow.

`TestBashForestRepetitionFoldDispatch` skipped because its optional real-corpus fixture was absent.

Reproduce the scoped checks inside a Docker container:

```sh
go test -race ./grammars -run '^Test(Issue1106Bash|Bash)' -count=1
CGO_ENABLED=0 GOTOOLCHAIN=go1.22.0 go test ./grammars -run '^Test(Issue1106Bash|Bash)' -count=1
cd cgo_harness
go test . -tags treesitter_c_parity -run '^(TestBash.*COracleParity|TestIssue1106BashEmptyAssignmentsCParity)$' -count=1
GTS_PARITY_MODE=smoke go test . -tags treesitter_c_parity -run '^TestParity(FreshParse|IncrementalParse|HasNoErrors)$/^bash$' -count=1
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1265,7 +1265,7 @@ jobs:
             --test-parallel 1 \
             --c-ref-build-jobs 1 \
             --timeout 20m \
-            --run '^(TestCompactT3OracleAdjudication|TestCompactT3JavaScriptRecoveryProfileCharacterization|TestCompactT3JavaScriptRecoveryRouteSeedOracleParity|TestCompactT3JavaScriptRecoveryMutationDifferential|TestCompactJavaScriptS5RecoveryMutationDifferential|TestCompactRouteLockedCExceptions)$'
+            --run '^(TestCompactT3OracleAdjudication|TestCompactT3JavaScriptRecoveryProfileCharacterization|TestCompactT3JavaScriptRecoveryRouteSeedOracleParity|TestCompactT3JavaScriptRecoveryMutationDifferential|TestCompactJavaScriptS5RecoveryMutationDifferential|TestCompactRouteLockedCExceptions|TestIssue1106BashEmptyAssignmentsCParity)$'
 
       - name: Elm admission contract against locked C
         run: |
--- a/cgo_harness/bash_issue1106_parity_test.go
+++ b/cgo_harness/bash_issue1106_parity_test.go
@@ -1,0 +1,94 @@
+//go:build cgo && treesitter_c_parity
+
+package cgoharness_test
+
+import (
+	"encoding/json"
+	"os"
+	"testing"
+
+	gotreesitter "github.com/odvcencio/gotreesitter"
+	harness "github.com/odvcencio/gotreesitter/cgo_harness"
+	"github.com/odvcencio/gotreesitter/grammars"
+	"github.com/odvcencio/gotreesitter/internal/benchfixtures"
+	sitter "github.com/tree-sitter/go-tree-sitter"
+)
+
+func TestIssue1106BashEmptyAssignmentsCParity(t *testing.T) {
+	data, err := os.ReadFile("../grammars/testdata/bash_empty_assignments.json")
+	if err != nil {
+		t.Fatal(err)
+	}
+	var cases []struct {
+		Name    string
+		Source  string
+		Invalid bool
+	}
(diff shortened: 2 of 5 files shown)


odvcencio (approved):
Reviewed the focused Bash scanner change and its regression coverage. Scoped Go, race, and locked C-parity tests pass.

PR #1109, opened 2026-09-13
fix(markdown): preserve repeated block closes in nested lists

denisvmedia:
Markdown lists with five or more nested levels produce error trees when several blocks close at the same byte. The scanner emits one zero-width `_block_close` per open block. The general four-token guard replaces the fifth close with EOF, or skips a rune at an internal dedent.

Add Markdown `_block_close` to the existing set of repeatable external tokens. Keep the existing 4,096-token bound and parser resource limits. Other languages retain their current behavior.

The blackbox regression checks depths 4, 5, 8, and 32 with two-space and four-space indentation. It covers EOF, a following paragraph, and an outer sibling. It verifies error state, source coverage, item count, and nesting depth.

Validation: `CGO_ENABLED=0 GOMAXPROCS=2 GOFLAGS=-p=1 go test ./grammars -run Markdown -count=1 -timeout=5m` passed in an isolated Linux Docker container. The new cases reproduce the failure on v0.52.0 without the patch. The same suite passes with the patch on v0.52.0 and on this PR's upstream-based commit `94dbf98`. Test compilation used a two-CPU quota and 2 GiB of memory.

Found while checking [Unswell issue #231](https://github.com/stokaro/unswell/issues/231), which includes the pinned FastAPI document that triggered the failure.


denisvmedia:
The three red jobs on this pull request never ran. Their check-run annotations all read:

> The job was not started because it repeatedly failed to be acquired (5 attempts).

That is runner acquisition on the hosted pool, not a test result. `race_root` and `build` are aggregators and failed because those three did.

The affected jobs are `race_packages (support, /grep$|/taproot($|/))`, `compact-route-lifecycle-history` and `race_root_isolated (TestIncrementalInvariantGateC)`.

Two checks that the change is not implicated:

- This branch is exactly one commit ahead of `main` at 21e4312b, with no drift. Run 34572942982 on that same base commit ran all three jobs and all three passed.
- All three pass locally on the branch: `go test . -race -count=1 -run '^TestIncrementalInvariantGateC$'`, `go test ./grep ./taproot/... -race -count=1`, and `GTS_REQUIRE_HISTORICAL_RECEIPT_PROOF=1 go test . -run '^TestCompactRouteCampaignLifecycleRegistry$' -count=1`.

A re-run of the failed jobs should clear it. I have no write access here to trigger one; say the word and I will push to the head branch instead.

denisvmedia:
Re-ran by pushing the same tree under a new commit SHA. Run 34939554339 is green: 45 jobs, no failures.

The earlier red was runner acquisition, not the change. The tree is byte for byte what it was; only the commit hash moved.

Ready for review.

odvcencio (approved):
Review completed. The change permits repeated _block_close tokens only for Markdown and retains the existing 4,096-token bound. Focused Markdown tests pass in Docker. Fresh, incremental, and error-state C comparisons pass. An additional local review test compares all 24 nested-list cases against the pinned C parser; every complete tree matches. No blocking defects found.

PR #1112, opened 2026-09-16
fix(typescript): preserve contextual in properties across newlines

### LINKED ISSUE #1108: TypeScript object-type property key in produces an ERROR node (strict parse) — v0.51.0 and v0.52.0
## Summary

In an **object type** (interface body / type alias), a property named `in` —
specifically the sequence `default: <type>` followed by `in: <type>` — makes
`Parser.ParseStrict` return a tree with `RootNode().HasError() == true`.
The reserved word `in` is routed into a `for...in`-style branch instead of
being accepted as a `property_identifier`, and the parser then swallows the
surrounding members into an `ERROR` node.

Verified on **v0.51.0** and **v0.52.0** with the built-in TypeScript grammar
(`grammars.DetectLanguage("source.ts")`). Real-world trigger: any TS design
token / theme file using an `easing: { default, in, out, bounce }` shape
(tailwind-style easing tokens) fails to parse, which in downstream tools with
fail-closed strict parsing (e.g. indexing a whole repository) rejects the
entire input.

## Minimal reproduction

```go
package main

import (
	"fmt"

)

func try(name, source string) {
	lang := grammars.DetectLanguage("source.ts").Language()
	parser := gotreesitter.NewParser(lang)
	parser.SetTimeoutMicros(2_000_000)
	tree, err := parser.ParseStrict([]byte(source))
	if err != nil {
		fmt.Printf("%-14s ParseStrict err: %v\n", name, err)
		return
	}
	fmt.Printf("%-14s hasError=%v stoppedEarly=%v\n",
		name, tree.RootNode().HasError(), tree.ParseStoppedEarly())
	tree.Release()
}

func main() {
	try("default+in", "interface A {\n  easing: {\n    default: string\n    in: string\n  }\n}\n") // FAILS
	try("in+default", "interface A {\n  easing: {\n    in: string\n    default: string\n  }\n}\n") // ok
	try("in+in",      "interface A {\n  easing: {\n    in: string\n    in: string\n  }\n}\n")      // FAILS
	try("default+out","interface A {\n  easing: {\n    default: string\n    out: string\n  }\n}\n") // ok
}
```

Output (identical on v0.51.0 and v0.52.0):

```
default+in     hasError=true  stoppedEarly=false
in+default     hasError=false stoppedEarly=false
in+in          hasError=true  stoppedEarly=false
default+out    hasError=false stoppedEarly=false
```

## Parse tree evidence (v0.52.0)

Input:

```ts
interface A {
  easing: {
    default: string
    in: string
  }
}
```

Tree (byte offsets):

```
program [0:67]
  interface_declaration [0:66]
    interface [0:9]
    type_identifier [10:11]
    interface_body [12:66]
      { [12:13]
      property_signature [16:64]
        property_identifier [16:22]            // "easing"
        type_annotation [22:64]
          :
          object_type [24:64]
            { [24:25]
            property_signature [30:45]
              property_identifier [30:37]      // "default" — accepted
              type_annotation [37:45]
                : [37:38]
                predefined_type [39:45]        // "string"
            ERROR [50:60]                      // ← swallows "in : string"
              in [50:52]                       // keyword token, NOT property_identifier
              : [52:53]
              string [54:60]
            } [63:64]
      } [65:66]
```

Notes:

- `default` alone parses fine (`default-key` ok); `in` alone after a **blank
  or comment-only** prefix also parses in some positions (`in+default` ok).
- The failure is **order-sensitive**: `default` then `in` fails, `in` then
  `default` succeeds. That points to a parser-state routing problem in the
  object-type property-name productions (the LR state after accepting a
  property signature appears to mis-route the lookahead `in`), not a token
  definition problem.
- Same ERROR shape appears with `type A = { ... }` aliases and inside
  real-world files; in one real token file the ERROR node was
  `ERROR [2557:2574]` containing `default: string` + `in`, with the following
  `out: string` left as an orphaned `type_annotation`:

```
                    property_signature [2550:2582]
                      property_identifier [2550:2557]   // "default"
                      ERROR [2557:2574]
                        : [2557:2558]
                        predefined_type [2559:2565]     // "string"
                        in [2572:2574]                  // keyword token
                      type_annotation [2574:2582]       // orphaned ": string" for "out"
```

## Expected behavior

`in` is a valid TypeScript object-type property name (contextual keyword;
`tsc`, esbuild, Babel, and upstream tree-sitter-typescript all accept it):

```ts
interface Easing { default: string; in: string; out: string; bounce: string }
```

`ParseStrict` should return `hasError=false` for the inputs above.

## Environment

- github.com/odvcencio/gotreesitter v0.51.0 and v0.52.0 (both reproduce)
- Grammar: built-in `grammars` TypeScript (`source.ts` detection path)
- Go 1.24, windows/amd64

## Suggested investigation

1. Check the object-type `property_name` alternatives for a missing
   `in`/reserved-word branch (`primary_expression` vs `property_identifier`
   routing).
2. The order sensitivity (`default` → `in` fails, `in` → `default` passes)
   suggests the state reached after `predefined_type` in
   `type_annotation` mis-predicts the next property signature; compare the
   goto table entries for that state against upstream tree-sitter-typescript
   `property_signature` handling.
3. Regression guard: the three failing inputs above are one-line additions
   to any existing strict-parse corpus.

Downstream impact: fail-closed strict consumers cannot index any TS file
containing `in:` (or `in`-adjacent) property keys in object types; a
workaround on the consumer side is per-file degradation, but the grammar
itself is where the fix belongs.

odvcencio:
## Change

Permit automatic semicolons before contextual `in` properties in TypeScript and TSX type members.
Preserve binary `in` expressions across newlines.
Apply the same scanner guard to the pinned C grammar patch.

## Validation

- Reproduced both reported failures on main. Three control cases passed.
- Passed five strict-parse regressions for TypeScript in Docker.
- Passed five strict-parse regressions for TSX in a separate Docker run.
- Passed 14 C comparison cases per language, including binary `in` across a newline.

The C comparisons use the pinned, patched grammar. They do not establish equivalence with an unpatched upstream scanner.
No performance claim or compact graduation claim is made.
This branch contains six source, test, and grammar-patch files. It contains no evidence archives or source snapshots.

The branch commit body incorrectly says the change stops semicolon insertion. The change permits insertion in the affected property context.
Use a corrected squash message when merging.

Fixes #1108


odvcencio (commented):
Self-review completed. No blocking defects found.

- Both scanner implementations use the same guard.
- The guard preserves the binary-expression context through the existing logical-or validity bit.
- Strict-parse regressions cover property order and repeated contextual names.
- Separate Docker runs compare both languages against the pinned, patched C parser.
- The outgoing branch contains no archive files or development-branch ancestry.

The commit-body wording needs correction in the squash message. Wait for all checks before the authorized admin merge.

Issue #1052, opened 2026-09-02
Track the compact graduation maintenance lane


odvcencio:
## Purpose

Run a maintenance lane beside the compact core, adversarial, certification, and performance lanes.

Keep parser decisions out of this lane. Do not mix this work into a core parser pull request.

## Current baseline

The deep audit used commit `24f23d6` in a conflicted checkout. Current main already includes these results:

- `.graftignore` excludes `.codex/`, malformed worktree paths, `harness_out/`, and private worktrees.
- The campaign inventory and lifecycle registry define owners, gates, receipts, and deletion conditions.
- The five reported leaf wrappers and the stale `parseIterations` comment are absent.
- `go_timing_oracle_10s.c` is a three-line parameter wrapper around the shared implementation.
- The compact package keeps its current name until behavioral graduation.

## Remaining work

Perform these items in order:

1. Verify one private incremental metric schema against all public mappings.
2. Generate or mechanically check flattening and aggregation code.
3. Move one retry or limit policy into an exact-blob runtime profile.
4. Share the Lua-pattern compiler between production and the parity harness.
5. Standardize generated markers and generator ownership.
6. Retire one result-compatibility mechanism after producer parity passes.
7. Extract pure parse bookkeeping and finalization after compact correctness stabilizes.
8. Rename the compact package only after graduation.

## Lane rules

- Make no parser decision change in this lane.
- Keep each change small and mechanically verifiable.
- Run focused tests before and after each change.
- Keep correctness gates separate from performance gates.
- Do not open a maintenance pull request while a ready pull request waits for review.
- Stop a change when it overlaps the active core package.

## Completion

Close this issue when the live items move into scoped issues or land with focused proof. Preserve stable release telemetry and the emergency fallback path.

odvcencio:
## Maintenance stack review

Closed pull requests #1058, #1059, and #1062 without merging them.

The three pull requests formed one stack. The final head deleted 102,764 lines across 35 files.

The review found these blockers:

- The final tree retained 43 references to deleted paths.
- CHANGELOG.md contained 38 of those references.
- The fleet-manifest generator still emitted the deleted output paths.
- The lifecycle registry still named two deleted documents.
- One of eleven sampled Hyphae replacements was accepted.
- Nine sampled replacements were unreviewed.
- The C0e/C0f replacement had partial status.
- Current Hyphae specifications still cited the deleted performance document by line.
- Replacement code comments used vague record names instead of exact Hyphae object identifiers.

[...]

PR #1116, opened 2026-09-16
Add standalone packages for every grammar

### LINKED ISSUE #1110: Per-grammar packages so a go install consumer can embed one grammar without build tags
[selfdoc](https://github.com/smm-h/selfdoc) is a documentation generator distributed as a Go CLI; users install it with `go install github.com/smm-h/selfdoc/cmd/selfdoc@v0`. It calls `grammars.PythonLanguage()` and needs that one grammar, never any other.

`grammar_subset` works well and cuts the binary substantially, but `go install` accepts no `-tags`, so only the release archives our own CI builds get the small binary — everyone who installs the module the ordinary way gets all 206 blobs.

Measured on v0.52.0, linux/amd64, Go 1.26.6 (bytes):

| binary | default | `-tags 'grammar_subset grammar_subset_python'` |
| --- | --- | --- |
| selfdoc | 57,774,287 | 38,225,404 |
| minimal program calling `grammars.PythonLanguage()` | 21,047,055 | 4,011,551 |

The existing per-grammar package does not close the gap by itself: a minimal program importing only `grammars/lean` and calling `lean.Language()` builds to 21,133,180 bytes. It embeds its own `lean.bin`, but it imports `grammars`, and the wildcard `//go:embed grammar_blobs/*.bin` in `blob_source_embedded.go` is reachable through `readGrammarBlob`, so all 206 blobs stay linked in.

**Request:** importable per-grammar packages a consumer can select without build tags — say `grammars/python` on the `grammars/lean` model — with the blob source split so that a program importing one grammar package does not pull the all-grammars wildcard embed into its import graph. A `grammars` package that imports every per-grammar package would keep today's behavior for existing consumers unchanged.

Any equivalent shape solves it just as well: a documented, tag-free way to register a single grammar's table blob (the consumer embeds the `.bin` itself and hands it to something like `grammars.LoadLanguage(name, blob)`, with the detection entry and external scanner registered), or whatever fits your design better.

This is shape (A) from #88, which was resolved with the `grammar_subset` tags. Those fit a consumer who controls their own `go build`; they cannot reach a consumer whose distribution channel is `go install`.

Happy to put together a PR if you tell me which shape you want.

odvcencio:
## Change

Closes #1110. Refs #1052.

Provide `grammars/<name>` for all 206 blob-backed grammars, without consumer build tags.
Each package selects one independently linked blob and its scanner support.
Share decoding, repairs, exact-artifact profiles, and caches through `grammars/runtime`.
Keep existing aggregate APIs through generated aliases and forwarding functions.
Remove Lean's aggregate dependency while preserving optional file detection.

The earlier Python work remains in this PR:

- Share Python scanner and checkpoint code with Bitbake, Mojo, and Starlark.
- Preserve raw-shape ordering for forest-local alternatives.
- Match the pinned C parser for the Python f-string splat result.
- Remove the obsolete interpolation compatibility pass. Keep the remaining Python dispatcher active.

No grammar blobs changed in the generic extraction.
The branch contains source changes only. Evidence archives and source snapshots remain outside Git.

## Verification

- All 206 standalone packages compile and exclude the aggregate dependency.
- Each standalone binary retains exactly one grammar blob, checked through linked symbols.
- All 206 standalone definitions and parse probes match aggregate loading.
- All 206 aggregate definitions and parse probes match the pre-extraction branch, `ae7afc9f`.
- Python standalone, focused race, fresh-parse parity, incremental parity, and checkpoint regressions pass in Docker.
- Shared loader, cache eviction, blob precedence, and decoder regression tests pass with the race detector.
- Lean registration, scanner attachment, queries, incremental parsing, and blob regeneration tests pass.
- Production builds pass in default, subset, core, external-blob, and fallback modes.
- Generator checks cover all 206 wrappers and 966 generated Go files.

Each language comparison runs in a separate process inside a bounded container.
The comparison checks every exported serializable language field, scanner type, and a small parse probe.
It does not replace application-corpus or pinned C parity coverage.

Subset test compilation has existing failures. The baseline reproduces those failures.
Subset production builds pass. This PR does not claim to repair the existing test configuration.

The earlier Python corpus sweep covered three generated fixtures and 30 constructed cases, with zero C divergences.
It did not establish external application-corpus coverage.
Earlier Bitbake, Mojo, and Starlark subset builds and C comparisons also passed.

## Performance safety

The generic extraction compares `78dc2465` against `ae7afc9f` with the primary Go benchmark trio and warm grammar lookup.
The comparison uses `scripts/run_randomized_benchmarks.sh` and 20 alternating, paired shuffle seeds.
Each process uses `GOMAXPROCS=1`, `-count=1`, `750ms`, and allocation metrics.
Docker pins CPU 2 on chi-1. Raw output remains outside the repository.
All 20 paired runs completed without a crash, memory failure, or timeout.

| Benchmark | Baseline | Candidate | Time change | Candidate allocations |
| --- | ---: | ---: | --- | ---: |
| Full Go parse | 9.596 ms | 9.618 ms | Not significant | 36 |
| Single-byte edit | 126.4 us | 126.2 us | Not significant | 3 |
| No-edit parse | 2.579 ns | 2.558 ns | -0.80% | 0 |
| Warm grammar lookup | 29.05 ns | 29.80 ns | +2.58% | 0 |

Allocation counts did not change. Full-parse bytes per operation did not change significantly.
Single-byte edits retained 119 bytes per operation. The other two benchmarks retained zero bytes per operation.
The warm lookup regression adds approximately 0.75 ns. Accept this measured packaging tradeoff without claiming a parser speedup.
The baseline contains the same added warm-lookup benchmark as the candidate.
Container Git metadata was unavailable. The source revisions above were checked on the host before and after the run.

Reproduce the comparison with the same warm-lookup benchmark in both worktrees:

```sh
bash scripts/run_randomized_benchmarks.sh \
  --output /evidence/candidate.txt \
  --baseline-root /baseline \
  --baseline-output /evidence/baseline.txt \
  --bench-regex '^(BenchmarkGoParse(FullDFA|IncrementalSingleByteEditDFA|IncrementalNoEditDFA)|BenchmarkGrammarLanguageWarm)$' \
  --require-benchmarks BenchmarkGoParseFullDFA,BenchmarkGoParseIncrementalSingleByteEditDFA,BenchmarkGoParseIncrementalNoEditDFA,BenchmarkGrammarLanguageWarm
```

Run inside Docker with CPU 2 pinned, one CPU, 4 GiB memory, and `GOMEMLIMIT=3GiB`.
Use Go 1.25.14 on linux/amd64. Keep the baseline at `ae7afc9f` and the candidate at `78dc2465`.

### Binary size and memory safety

Build minimal `NewParser(Language()).ParseStrict(source)` programs with Go 1.25.14 and `-trimpath -ldflags="-s -w"`.
The Python aggregate and standalone programs parse the same `x = 1` source.

| Import | Binary bytes |
| --- | ---: |
| Aggregate, selecting Python | 28,307,640 |
| `grammars/python` | 10,002,616 |
| `grammars/go` | 10,694,840 |
| `grammars/json` | 9,908,408 |
| `grammars/lean` | 9,965,752 |

The Python standalone program is 64.7 percent smaller than the aggregate program.
All five programs parse successfully. The separate symbol test confirms exactly one retained blob in each standalone package.

A standalone Go probe parsed 25,000 generated functions five times with one parser.
Each function used `func fN() int { v := N; return v }`, after `package main`.
The source contained 1,052,793 bytes. Every parse covered the source without errors or an early stop.
Release each tree and collect garbage after each parse.
Retained heap stabilized at 286,541,344 bytes. Maximum resident memory was 459,372 KiB, measured with `/usr/bin/time -v`.
The Docker run used CPU 2, `GOMAXPROCS=1`, a 4 GiB memory limit, and `GOMEMLIMIT=3GiB`.
Focused scratch, Go compatibility, and compact memory-budget tests also passed.

The earlier Python-only slice recorded neutral parse and edit times with unchanged allocations.
Its no-edit benchmark increased 1.71 percent. That measured tradeoff remains recorded here.
Removing the interpolation compatibility pass reduced its measured traversal time by 77.47 percent.
Those earlier measurements do not measure the generic extraction.

## Review

The review checked package cycles, blob reachability, scanner registration, profile identity, cache sharing, and generator ownership.
The shared runtime keeps scanner certification interfaces in the same Go package as their implementations.
Aggregate catalog selection takes precedence over individual providers, including external-file read errors.
Generated wrappers carry subset constraints. Existing scanner tests move with private implementation details.

The review also corrected package-sensitive Go authentication, preserved the original SQL scanner identity, and repaired certification references.
Go 1.22 generator tests pass. The Python retirement fixture uses the integration commit that contains the actual deletion.

The final 206-package check and safety probes pass at `78dc2465`.
All 48 remote checks passed in run 35165091083. Three conditional checks skipped.
The exact reviewed head merged as `3877deef941bec56584307faad7c32958be70ba3`.


odvcencio:
Integration hold: this PR is now a draft. Issue #1110 needs a reusable packaging design, not a separate handwritten loader for each language. The Python scanner extraction and verified parser changes remain useful, but loader setup and runtime-profile ownership need a shared design. Python and Lean will serve as dependency and behavior checks for that design. Do not merge this PR or close #1110 yet.

odvcencio (commented):
Reviewed commit `78dc2465dfde11c1cc9326635ae3c6069134eac6` against `5ddbca350be307b2b96307563c0d91dcc03a18ac`.

No blocking code findings remain after the certification fixes.

- All 206 standalone packages exclude the aggregate catalog and retain one blob each.
- Their serialized definitions and parse probes match aggregate loading.
- The aggregate decoder, cache, scanner registration, and source precedence remain shared.
- Go authentication recognizes the relocated scanner and rejects uncertified wrappers.
- SQL retains its original certified source hash and checkpoint identity.
- Generated API aliases preserve existing callers. Remote API compatibility checks pass.
- The Python retirement records the actual source deletion and its integration ancestry.
- The 20-seed paired comparison preserves parser allocation counts and shows no significant full-parse or edit-time regression.
- Warm lookup adds approximately 0.75 ns, with zero allocations. The PR records this tradeoff.
- The large Go probe completes five parses, with stable retained heap and no memory-limit failure.
- The branch adds no binary artifacts, evidence archives, or source snapshots.

The complete catalog check uses small probes. It does not establish application-corpus parity for every grammar.
Existing subset test compilation failures also remain outside this change; the baseline reproduces them.

Remote correctness and race jobs are still running. Merge only after they finish successfully.

PR #1117, opened 2026-09-16
Restore the compact parser opt-out build

odvcencio:
## Summary

Restore builds with `gts_no_parsercorephase0`. Main currently fails before tests can run with this tag.

- Exclude the compact materializer from the emergency build.
- Match the recovery stub signature to its callers.
- Exclude the compact-only table analyzer when its implementation is absent.
- Build this configuration and run its existing fallback test in continuous integration.

The default build retains the same implementation. No grammar, parser algorithm, or evidence archive changes are included.

## Verification

Docker checks pass:

- `go build -tags gts_no_parsercorephase0 ./...`
- The existing emergency fallback test and concurrent parser-pool test.
- The public memory-budget test with the emergency tag.
- Default-build Go fresh, incremental, and error-free comparisons against C.

The optional C diagnostic harness still has compact-only tests without opt-out guards.
This PR restores the public module build, not that diagnostic tag combination.

## Queue

PR 1105 remains draft with unresolved parser failures. PR 1116 remains draft pending shared grammar packaging.
Neither draft is superseded by this focused build repair.
This change supports the emergency fallback requirement in #1052. It does not close that issue or establish release readiness.


odvcencio (commented):
Self-review complete. The four-file change restores the opt-out build without changing the default parser implementation. The recovery stub now matches both callers. The existing fallback test remains unchanged and is added to continuous integration. Docker module builds, fallback routing, concurrent pool, and memory checks pass. Default Go comparisons against C also pass. The optional compact diagnostic harness remains outside the repaired tag combination. Await passing remote checks before merge.

odvcencio (commented):
The failed root race shard passed its tests. Artifact finalization then failed with HTTP 403 from an intermediary. I requested a rerun of that job. The check remains required; no merge is authorized by this infrastructure diagnosis alone.

Issue #1100, opened 2026-09-07
c: large designated-initializer arrays parse ~100-400x slower than expected (v0.51.0)


townsendmerino:
## Summary

On `v0.51.0` the `c` grammar shows severe slowdowns (not crashes — every case
below returns `ParseStopAccepted`) on real driver-style C files that use
large designated-initializer arrays of the form `[IDENT] = &expr.field,`. A
representative file (43 KB, syntactically valid, complete) took **~4.7–4.9s**
to parse across repeated runs, where files of similar size in the same
codebase typically parse in tens of milliseconds.

This looks like the same failure *class* as #972 (`c_sharp`'s collection
initializer) — a large homogeneous-shaped initializer list costing far more
than its size suggests — on a different grammar, so it may share a root
cause in core list/ambiguity handling rather than being `c`-grammar-specific.
Filing separately since the trigger construct and grammar differ.

## Repro (real file — I could not reduce this to a clean synthetic minimal
case; see "What I tried and why it's inconclusive" below)

```go
package main

import (
	"fmt"
	"os"
	"time"

	"github.com/odvcencio/gotreesitter"
	"github.com/odvcencio/gotreesitter/grammars"
)

func main() {
	// torvalds/linux @ v6.6, commit ffc253263a1375a65fa6c9f62a893e9767fbebfa
	// path: drivers/clk/qcom/gcc-mdm9607.c
	data, err := os.ReadFile(os.Args[1])
	if err != nil {
		panic(err)
	}
	entry := grammars.DetectLanguageByName("c")
	pool := gotreesitter.NewParserPool(entry.Language(),
		gotreesitter.WithParserPoolTimeoutMicros(0)) // no timeout
	for i := 0; i < 3; i++ {
		t0 := time.Now()
		tree, err := pool.Parse(data)
		fmt.Printf("run %d: %d bytes  %7.3fs  err=%v  stop=%v\n",
			i, len(data), time.Since(t0).Seconds(), err, tree.ParseStopReason())
	}
}
```

```
$ git clone --depth 1 --branch v6.6 --filter=blob:none --sparse https://github.com/torvalds/linux /tmp/linux
$ cd /tmp/linux && git sparse-checkout set drivers/clk/qcom
$ go run repro.go drivers/clk/qcom/gcc-mdm9607.c
run 0: 42995 bytes    4.862s  err=<nil>  stop=accepted
run 1: 42995 bytes    4.732s  err=<nil>  stop=accepted
[...]

PR #1118, opened 2026-09-16
Expose deterministic parser work limits

odvcencio:
## Summary

Expose production parser thresholds through `ParseWorkLimits`, `Parser.SetParseWorkLimits`, and `WithParserPoolParseWorkLimits`.
This source-only harvest addresses the deterministic-limit request in #1100. It does not close the issue.

- Configure iteration, node, and primary stack-depth thresholds.
- Inherit thresholds in recovery and snippet parsers.
- Restore configured thresholds when a pool reuses a parser.
- Bypass compact and forest speculation while explicit thresholds are configured.
- Preserve partial trees and strict-method stop errors.
- Restore normal routing when all fields are cleared.

The documentation states the limits precisely. Each production loop receives the thresholds; they do not count total operation work or allocated bytes.
Node and depth checks occur between steps, so a step can exceed a threshold.
Timeouts, cancellation, and memory limits remain active. Selecting production can change performance and tree selection.

## Verification

Focused Docker checks pass:

- Threshold normalization, exact resolution, and all three stop reasons.
- Recovery inheritance, snippet reset, pool reuse, and route restoration.
- Public strict full and incremental parsing.
- Focused race tests, including concurrent pool use.
- Go changed-lookahead comparisons against the locked C parser.
- Separate Go and C fresh, incremental, and error-free parity tests.
- Existing memory-budget tests, including the large table fixture.

The large composite probe completes with maximum resident memory of 126,816 KiB.
No verification container exhausted memory or reached its wall-time limit.

## Performance

Twenty paired, alternating seeds compare main `5e484edf` with this change, using Go 1.25.14 on linux/amd64.
The campaign uses `scripts/run_randomized_benchmarks.sh`, `GOMAXPROCS=1`, one pinned CPU, and `-benchtime=750ms`.
Each process uses one seed and `-count=1`.

| --- | ---: | ---: | --- |
| Go full parse | 9.587 ms | 9.556 ms | No significant change |
| Go single-byte edit | 124.7 us | 125.7 us | 0.79% slower |
| Go without an edit | 2.613 ns | 2.569 ns | 1.68% faster |
| Complete C file from #1100 | 22.89 ms | 22.97 ms | No significant change |

Go allocations remain 36, 3, and 0 respectively. C allocations show no significant change.
The edit slowdown is a recorded tradeoff for the public safeguard, not an optimization claim.
Three inline integer thresholds increase each parser from 2280 to 2304 bytes on amd64.
The exact layout test includes these 24 bytes. Other layout assertions remain unchanged.
The small no-edit difference does not establish a causal speed improvement.
Raw output and local probes remain outside the repository.

## Issue 1100 findings

The workload is [gcc-mdm9607.c at the reported Linux revision](https://github.com/torvalds/linux/blob/ffc253263a1375a65fa6c9f62a893e9767fbebfa/drivers/clk/qcom/gcc-mdm9607.c).
Its 42,995 bytes have SHA-256 `0040f12a1c9323a575ac8073d8897480a8834e35306f60c607196ada4ebdf7b7`.
The full-file benchmark reuses a `ParserPool` with no timeout and releases each accepted, full-span tree.

With an iteration threshold of 2,000, three runs stop at byte 7,741 with identical partial trees and stop reasons.
Without explicit limits, both baseline and candidate return the same full tree and parser work counts.

The full tree does not match locked C. Both parsers report errors, but the root has 149 Go children versus 147 C children.
The unchanged Go deep digest is `1fbfd4e7aeea590b444e7a9e16c5889f6eaba4627aafc86441f72290723490a6`.
The C digest is `efa77b2eb81969a06d13031a2f10c9d122c2ded757d4a8178e9db8907d5516cb`.
This pre-existing difference and the broader reported workload remain unresolved.

No evidence archives, source snapshots, or archive-containing development ancestry are included.


odvcencio (commented):
Self-review complete. The API preserves source-derived defaults and reports configured thresholds through ParseRuntime. Recovery and snippet parsers inherit them; pooled parsers restore them on checkout. Explicit thresholds prevent compact and forest speculation. Clearing thresholds restores normal routing. Tests cover all stop reasons, strict full and incremental calls, pool reuse, race behavior, and existing memory budgets. The documentation distinguishes per-loop thresholds from total work and explains checkpoint overshoot. The 0.79 percent Go edit slowdown is recorded. The C tree difference in #1100 also exists on main and remains unresolved. Await passing remote checks before merge.

odvcencio:
CI found one stale layout assertion. The inline work thresholds add exactly 24 bytes on amd64, from 2280 to 2304. Commit c1df8266 updates that exact assertion and documents the cost. The runtime, tree, and field-offset assertions remain unchanged. The focused Docker race test passes. Production code and the measured candidate remain unchanged; CI is running on the updated head.

odvcencio (commented):
Final self-review covers c1df8266. The exact parser size assertion now includes the 24-byte inline configuration. All other layout assertions remain intact. Production behavior and the benchmarked candidate are unchanged by this follow-up. All 48 checks succeeded, with three conditional checks skipped. This includes exhaustive parity and isolated race gates. No evidence archives or diagnostic files are in the outgoing range.

PR #1119, opened 2026-09-16
Share Lua pattern compilation across query consumers

odvcencio:
## Summary

Use one Lua-pattern compiler for production queries, both C query comparisons, and the comparison report.
This harvest extends the local branch's compiler extraction to remove the highlight and report copies too.

- Remove the duplicate compiler and both private forwarding functions.
- Move the identical unit cases into the shared package.
- Evaluate `#lua-match?` in the outline comparison.
- Test the Lua uppercase class through Thrift outline and highlight queries.
- Check that the highlight query keeps `Upper` and rejects `lower`.
- Add a focused query-compilation benchmark.
- Correct the report's literal-dot behavior and test wildcard matching against production queries.

The change removes 163 lines overall, including its new tests and benchmark.
Continuous integration runs the shared compiler race tests, both C comparisons, and the report regression explicitly.
The shared compiler body matches both original bodies after function-name and blank-line normalization.
The production translation rules and parser algorithms remain unchanged.
The report used a different compiler. It now uses the same supported pattern rules as production.

This completes the shared-compiler item in #1052. The other maintenance items remain open.

## Verification

Focused Docker checks pass before and after extraction.
The shared unit tests pass with the race detector. Production predicate tests pass.
Both Thrift comparisons pass under default and exhaustive scope with nonempty results.
The report's wildcard case fails before extraction and passes afterward. Its full unit suite passes with the race detector.

The shared callback is not an independent Lua implementation. Explicit expected-match unit tests retain the translation contract.
The report-only follow-up does not change the production code measured below.

## Performance

Twenty paired, alternating seeds use Go 1.25.14, linux/amd64, one pinned CPU, `GOMAXPROCS=1`, and `-benchtime=750ms`.
Each process runs one seed with `-count=1` through `scripts/run_randomized_benchmarks.sh`.

| --- | ---: | ---: | --- |
| Lua query compilation | 2.797 us | 2.792 us | No significant change |
| Go full parse | 9.727 ms | 9.772 ms | No significant change |
| Go single-byte edit | 125.4 us | 126.8 us | 1.16% slower |
| Go without an edit | 2.628 ns | 2.607 ns | 0.82% faster |

Lua compilation remains at 61 allocations and 6.305 KiB per operation.
Go allocation counts remain 36, 3, and 0. Byte allocations show no significant change.
The edit regression is recorded, despite unchanged parser code. This is a maintenance improvement, not a speed improvement.

An earlier run had an unused C benchmark in the control worktree. The table uses the clean rerun, not that run.
Raw output and local source probes remain outside the repository. No evidence archives are included.


odvcencio (commented):
Self-review complete at 113e7dd1. Four consumers now use one compiler. The production compiler body matches the original production and highlight implementations. The report copy had different wildcard behavior; the new regression fails before replacement and passes afterward. All report unit tests pass with the race detector. The shared compiler and both Thrift comparisons also pass. CI explicitly includes the new package and all three comparison tests. The change removes 163 lines overall. The measured 1.16 percent Go edit regression remains documented; production performance code did not change in the report follow-up. No archives, full inputs, or diagnostic programs are included. Await passing remote checks before merge.

odvcencio (commented):
Final self-review confirms the unchanged head 113e7dd1 and 47 successful checks, with four conditional checks skipped. The shared compiler race suite, report wildcard regression, C comparison tests, and isolated parser race gates all passed. The outgoing range contains only source, tests, a benchmark, a changelog entry, and CI configuration. The net reduction remains 163 lines. No evidence archives are included.

PR #1120, opened 2026-09-16
perf(recovery): Reuse the standard memo array

### LINKED ISSUE #1100: c: large designated-initializer arrays parse ~100-400x slower than expected (v0.51.0)

odvcencio:
## Summary

Reuse the standard recovery memo array when a later parse operation grows its small initial view.
Clear the active entries and hidden tail before reuse. Keep allocation behavior unchanged for other capacities.

- Keep the existing collision threshold and logical initial size.
- Preserve temporary-tier ownership and bounded retained memory.
- Test recursive growth, epoch wrap, slot addressing, and oversized-capacity replacement.
- Require zero allocations for warm standard regrowth.

The final diff changes two files. The production change adds seven lines and removes one.
It does not change recovery decisions, parser layout, grammar tables, or public APIs.

## Review-driven scope reduction

The first revision also changed pause progress and recovery version selection.
That revision fixed the C tree from #1100 but failed existing Go correctness tests in continuous integration (CI).
It also changed KDL, AWK, included-range, and dispatcher receipts.
The pause-only experiment retained several failures. The semantic changes remain on a local investigation branch.

This revision removes those changes from the final diff. No existing correctness assertion or expected digest is weakened.
The memo-only change does not fix the C tree or close #1100.

## Verification

Focused Docker race tests pass for cache growth, epoch wrap, recursive collisions, temporary ownership, and retained memory bounds.
The unchanged Go issue 490, scanner-edit sweep, and reuse-budget regression tests pass.
The sweep covers 2,160 edit sites. It exercises reuse at 2,139 sites.
Separate C and Go smoke comparisons pass. Go changed-lookahead parity and included-range receipts pass.
The unchanged AWK and KDL tree assertions pass.

The new warm-regrowth test fails on main: it replaces the array and allocates one object per operation.
The candidate retains the array and allocates zero objects for that operation.
The public memory tests pass, including the 19 MB table case.
The large Go probe reaches 127,116 KiB maximum resident memory, without an out-of-memory failure or timeout.
The complete C input retains its full-span accepted tree across three cold and warm parses on each side.
Its tree expression hash stays `b3bb3060137bcc76e5c8d2f5dc8e4a93edc3d0f5b0b5a2e7478946ecce12bcb1`.
The repository hygiene tests pass. CI run `35158007595` finishes with 48 successful checks and three conditional skips.

## Performance

The comparison uses main `758d3bc3` as the baseline and its merge with candidate `e547d2c6` as the treatment.
It uses Go 1.25.14 on linux/amd64, 20 paired alternating shuffle seeds, and CPU 2 affinity.
Both sides use `scripts/run_randomized_benchmarks.sh`, `GOMAXPROCS=1`, `-count=1`, `-benchtime=750ms`, and allocation reporting.

| --- | ---: | ---: | --- |
| Complete C file | 23.08 ms | 22.91 ms | No significant change |
| Go full parse | 9.626 ms | 9.692 ms | 0.68% slower |
| Go single-byte edit | 125.2 us | 126.8 us | 1.34% slower |
| Go without an edit | 2.613 ns | 2.583 ns | 1.13% faster |

C allocated bytes fall from 8.441 MiB to 8.076 MiB per operation, a 4.32% decrease.
C allocation counts decrease 0.01%. Go allocation counts remain 36, 3, and 0; Go allocated bytes show no significant change.
The memory reduction is the measured benefit. The small Go timing regressions are an explicit tradeoff.
These results replace the first revision's measurements. They do not show a general parser speed improvement.

The complete Linux input is 42,995 bytes from commit `ffc253263a1375a65fa6c9f62a893e9767fbebfa`.
Its path is `drivers/clk/qcom/gcc-mdm9607.c`.
Its SHA-256 is `0040f12a1c9323a575ac8073d8897480a8834e35306f60c607196ada4ebdf7b7`.

All local commits use Buckley. Full inputs, diagnostic probes, and raw benchmark output remain outside the outgoing files.
No evidence archives, source snapshots, or original development ancestry are included.

## Reproduce the C measurement

Fetch the pinned input from the [Linux source](https://raw.githubusercontent.com/torvalds/linux/ffc253263a1375a65fa6c9f62a893e9767fbebfa/drivers/clk/qcom/gcc-mdm9607.c).
Check its byte count and SHA-256 against the values above.
Keep it outside Git and mount its directory at `/integration` in Docker.
Place this temporary benchmark in both worktrees. Do not commit it.

```go
package gotreesitter_test

import (
    "os"
    "testing"

    gts "github.com/odvcencio/gotreesitter"
)

func BenchmarkIssue1100CompleteCFile(b *testing.B) {
    source, err := os.ReadFile("/integration/gcc-mdm9607.c")
    if err != nil { b.Fatal(err) }
    pool := gts.NewParserPool(grammars.CLanguage(), gts.WithParserPoolTimeoutMicros(0))
    b.SetBytes(int64(len(source)))
    b.ReportAllocs()
    b.ResetTimer()
    for i := 0; i < b.N; i++ {
        tree, err := pool.Parse(source)
        if err != nil || tree == nil { b.Fatalf("parse: %v", err) }
        if tree.ParseStopReason() != gts.ParseStopAccepted || tree.RootNode().EndByte() != uint32(len(source)) {
            b.Fatalf("incomplete parse: %s", tree.ParseRuntime().Summary())
        }
        tree.Release()
    }
}
```

Use baseline `758d3bc3`. Merge candidate `e547d2c6` with that baseline in a separate worktree.
Run the command below from `/workspace` in the candidate's Docker container.
Mount the baseline at `/integration/memo-only-baseline`.
Set one CPU, affinity to CPU 2, a 4 GiB memory limit, and `GOMEMLIMIT=3GiB`.

```sh
bash scripts/run_randomized_benchmarks.sh \
  --output /integration/memo-only-after.txt \
  --baseline-root /integration/memo-only-baseline \
  --baseline-output /integration/memo-only-before.txt \
  --runs 20 --benchtime 750ms \
  --bench-regex '^(BenchmarkGoParse(FullDFA|IncrementalSingleByteEditDFA|IncrementalNoEditDFA)|BenchmarkIssue1100CompleteCFile)$' \
  --require-benchmarks BenchmarkGoParseFullDFA,BenchmarkGoParseIncrementalSingleByteEditDFA,BenchmarkGoParseIncrementalNoEditDFA,BenchmarkIssue1100CompleteCFile
```

The script sets `GOMAXPROCS=1` and alternates the baseline and candidate for each seed.
Compare the complete outputs with `benchstat`. Keep those outputs outside Git.


odvcencio (commented):
Self-review of e547d2c6bf3889e4c419025697daa616d4c03464. This is not an independent approval.

The final diff contains only standard memo reuse and its focused tests.
Parser decisions, layouts, grammar tables, and public interfaces remain unchanged.

- Reuse requires exact standard capacity. Oversized arrays still receive a bounded replacement.
- Clear every entry after extending the slice. This removes stale entries from the hidden tail.
- Preserve the collision threshold, epoch, returned-slot address, and temporary-tier ownership.
- Require zero allocations during warm standard regrowth.

The local Docker recovery race suite passes with the unchanged Go regression and scanner-reuse tests.
Separate C, Go, AWK, and KDL checks pass. Go included-range receipts and changed-lookahead parity remain unchanged.

The earlier semantic change is excluded after correctness failures. No existing assertion or expected digest was relaxed.
The outgoing files contain no raw evidence, full source snapshots, or original development ancestry.

The merge remains blocked until the new CI run and isolated randomized performance comparison finish.

odvcencio (commented):
Final self-review of e547d2c6bf3889e4c419025697daa616d4c03464. This is not an independent approval.

CI run 35158007595 completed with 48 successful checks and three conditional skips.
The exhaustive C comparison and isolated race gates passed.
The final two-file diff preserves recovery decisions and all existing correctness assertions.

The 20-pair comparison reduces C allocated bytes by 4.32%, with unchanged C timing statistically.
Go full parsing regresses 0.68%, and single-byte edits regress 1.34%. These tradeoffs are recorded in the PR body.
The public memory tests pass. The large-file probe reaches 127116 KiB maximum resident memory.
The new allocation regression fails on main and passes on the candidate.
The complete C input preserves its selected tree before and after memo reuse.

No blocking finding remains for the memo-only diff. The excluded semantic change remains blocked by Go correctness failures.
The user authorized an admin merge after review and passing checks. Branch protection remains unchanged.

PR #1105, opened 2026-09-08
fix(go): align the embedded grammar with the locked C baseline

odvcencio:
## Scope

Regenerate the Go grammar from the locked C revision. Preserve the application binary interface, version 15, and reserved-word tables.
Omit the legacy external scanner when the grammar has no external tokens.
Correct keyword admission and recognition metadata.

This draft includes dependent parser ordering, recovery, and incremental reuse changes.
It preserves source hashes, expected tree digests, and historical test records.
It preserves all 206 standalone packages and the shared runtime introduced by #1116.
The standalone Go package uses the migrated artifact hash.

Continue this existing draft. Do not open a replacement PR.
The original migration worktree retains its uncommitted changes.

## Main integration

Buckley merge commit `a48b28320c9b646b66cee6874895949c20ad9bc0` integrates main `152e4e89600cc6ec4a7e5f22f3849caea230db7f`.
It resolves the changelog conflict and preserves both entries.
The exhaustive stack-hash flag test and focused Go conflict comparisons pass after integration.

The earlier integration at `9492a0e6` resolved four recovery calls after removal of the previous scheduler helper.

## Runtime corrections already in this draft

- Renew the incremental reduction allowance only when the stack reaches a new minimum depth.
- Preserve the limit for unchanged depth, increasing depth, and repeated growth followed by shrinkage.
- Preserve surviving stack order when a merge bucket discards a weaker alternative.
- Apply this order to normal, deferred, and large-cap merges.
- Repair affected indices and recalculate hashes and the weakest entry for the changed bucket.
- Record a diagnostic removal instead of replacing an earlier version.

The reduction correction fixes two C edits that previously omitted 577 source bytes.
The survivor-order correction at `78ebe21e` fixes the production Go binary-expression mismatch against C.

## Issue #454 test correction

Two failing C tests required a full retry even after the deletion began reusing the unchanged suffix.
The updated tests require complete fresh-tree equality, genuine reuse, and bounded byte counters.
The existing allocation ceiling remains 800,000 nodes.
Direct tests separately check budget calculation, the reuse threshold, and bounded full retries.

The 137 KiB deletion reuses 135,832 of 140,287 bytes without a full retry.
Replacement and insertion also preserve exact fresh-tree equality.
These are correctness observations, not comparative performance measurements.

The one-KiB malformed witness now matches the complete locked-C digest:
`35c4b6da0092ff35252de94385dbe959f282c317b6904a9c1e2a380d5481fbf9`.
[...]

Issue #1065, opened 2026-09-03
Pass the compact parser performance and release gates


odvcencio:
## Summary

Run package seven after grammar certification completes.

Keep correctness results fixed while measuring memory, race safety, fallback behavior, and performance.

## Correctness precondition

Do not begin release measurements until issue #1064 has reproducible grammar receipts.

Re-run the focused correctness gate before and after each performance change.

## Required safety gates

- No crash or out-of-memory failure.
- No parser memory-budget violation.
- No runaway retention after repeated parses.
- No race failure in the dedicated container or continuous integration.
- No unbounded frontier, derivation, summary, or recovery growth.
- No unexplained compact fallback.

## Performance loop

Use `scripts/run_randomized_benchmarks.sh` with the repository settings.

Measure:

- Full parse.
- Incremental single-byte edits.
- Incremental no-edit parses.
- Nanoseconds per operation.
- Bytes per operation.
- Allocations per operation.
- Maximum resident set size on large files.

Use `GOMAXPROCS=1`, 20 shuffle seeds, `-benchtime=750ms`, and `-benchmem`.

Keep one focused change only when `benchstat` improves the target metrics without a correctness regression.

## Attribution

Split incremental time into:

- `Tree.Edit(edit)`.
- Reuse cursor and selection work.
- Reparse and rebuild work.

Use profiles to choose between invalidation, recovery, materialization, and allocation work.

## Release receipt

Publish:

- The exact candidate and baseline commits.
- All benchmark commands and artifacts.
- The randomized `benchstat` result.
- Maximum resident set size evidence.
- Race, fuzz, and retention results.
- Compact fallback rates by grammar.
- Every directional regression and its tradeoff.

## Exit gate

Complete this package when the compact route can become the default without hidden exceptions.

odvcencio:
Package-one performance follow-up:

- `external_scanner_adapter.go` now implements `ErrorTreeIncrementalReuseExternalScanner`.
- The adapter forwards an explicit inner capability.
- The adapter returns `false` when the inner scanner omits the optional interface.
- A direct scanner defaults to `true` when that interface is absent.

This fail-closed difference can force a fresh parse for an adapted scanner on an error-bearing old tree. It preserves correctness but can reduce incremental reuse.

Measure this case during package seven. Do not relax the fallback without a correctness certificate for adapted error-tree reuse.

odvcencio:
[...]

PR #1121, opened 2026-09-17
fix(parser): separate selected reuse from attempted work

odvcencio:
## Summary

Fix incremental profile accounting for the downstream report in #454.
Discarded attempts contributed reuse bytes to the selected result.
The 2 KB Go deletion reported 3,985 reused bytes for 2,086 source bytes.

- Report reuse coverage and result status from the selected attempt.
- Aggregate work and phase times across attempted parses.
- Retain maximum resource counts across fallback attempts.
- Add public, retry-selection, schema, and compact fallback regression tests.

This change does not alter parser selection, grammar data, or memory budgets.
It does not resolve the complete downstream report.

## Verification

- Reproduce the public regression on main before the fix.
- Pass focused Docker tests for profiles, retries, and fallback accounting.
- Pass the public regression with the compact parser disabled.
- Pass the C memory-budget fallback regression in Docker.
- Pass targeted race tests in Docker.
- Match the pinned C parser exactly for compact decline and changed-lookahead cases.

The final candidate also passes the focused race rerun.
The 137 KiB reproduction completes 100 incremental parses in a bounded Docker run.
Its final deep tree digest matches a fresh parse. Maximum resident set size is 225,860 KiB.
Selected reuse is 136,787 bytes for 140,367 edited source bytes.
This Linux diagnostic does not establish downstream Windows performance or complete retention certification.

## Performance comparison

Compare main `3877deef941bec56584307faad7c32958be70ba3` with candidate `d75ef834fe505032cfe43b6f1bd9fb6cfaf55072`.
Both trees used the same benchmark helper from this PR.

Run `scripts/run_randomized_benchmarks.sh` with paired, alternating order and 20 explicit shuffle seeds.
Use Docker, CPU 2, `GOMAXPROCS=1`, one process per seed, `-count=1`, `-benchtime=750ms`, and `-benchmem`.

| Benchmark | Baseline time | Candidate time | Allocations, both |
| --- | ---: | ---: | ---: |
| Full Go parse | 9.819 ms | 9.706 ms | 36 |
| Single-byte Go edit | 128.3 us | 127.1 us | 3 |
| No-edit Go parse | 2.579 ns | 2.578 ns | 0 |
| Profiled issue 454 deletion | 462.2 us | 459.5 us | 73 |

Benchstat reports no significant timing difference for any benchmark.
Median bytes per operation remain statistically unchanged, except a rounded zero-percent distribution change in the single-byte control.
The full-parse medians are 225.5 KiB and 225.9 KiB. The deletion medians are both 353.0 KiB.
The single-byte medians are both 119 bytes. The no-edit benchmark allocates zero bytes.

Review found an escaping timing record in the first revision.
The second Buckley commit returns that record by value and restores the baseline allocation count.
The first campaign was stopped and is not comparison evidence.

Raw evidence remains outside Git under `/tmp/gts-main-integration-20260916.joX1O8/profile-value-evidence`.

- Baseline SHA-256: `e8dd722b578971916f073d8f0b1bd62e3a23a30f335e36ba1d7e1c76eff98b61`.
- Candidate SHA-256: `0c6cdaca378ee993b06c632c96ec9b16e487f6ad423d3ac56fddf1ae7605baec`.
- Shared benchmark helper SHA-256: `79d90ee93d76d832ce5f482204d11adcbd6951cc830d3eea99005735def8ede5`.

Keep final continuous integration green before merging.

## Scope

Harvest source changes from development commits 967940ff and 1d79eaf.
Do not include their ancestry or evidence files.
Keep #1105 separate because its grammar migration still fails correctness gates.
Keep #454 open for the remaining performance and platform reports.

Refs #454, #1065


odvcencio (commented):
Review of d75ef834fe505032cfe43b6f1bd9fb6cfaf55072 against main 3877deef.

No blocking code findings remain in this review.
This is a maintainer self-review, not an independent approval.

- The selected retry supplies reuse coverage and result status. Discarded attempts supply work counters only.
- Fresh fallback accounting preserves prior work and selects the returned boundary.
- Peak resource counters use maxima. Retry control fields remain operation-scoped.
- The schema test checks every timing field and public profile field.
- Adoption and rejection tests verify different selected reuse counts with identical total work.
- The public reproduction fails on main and passes on the candidate.
- Compact decline and changed-lookahead trees match the pinned C oracle exactly.
- Parser selection, grammar artifacts, and memory budgets remain unchanged.

Review found one extra heap allocation in the first revision.
Returning timing by value removes it. The complete paired 20-seed comparison confirms unchanged allocation counts and no significant timing difference.

Keep #454 open. Nested fresh-parse retry telemetry and the remaining platform reports require separate work.
Do not interpret this change as complete compact graduation or a throughput improvement.
The branch contains two Buckley commits, source, tests, and documentation only.
No archive ancestry, evidence archives, profiles, or source snapshots are included.

Final continuous integration and the final focused race rerun must pass before merge.

Issue #1064, opened 2026-09-03
Certify compact grammar parity against the locked C baseline


odvcencio:
## Summary

Run package six after the first five core packages merge.

Certify one grammar at a time against the locked C runtime. Reduce every new mismatch before any broad matrix run.

## Required comparisons

Compare these outputs exactly:

- Tree symbols and child order.
- Byte and point ranges.
- Named, extra, missing, and error flags.
- Field assignments.
- Error node structure.
- Fresh parse results.
- Incremental parse results after each edit.
- Reuse and invalidation telemetry.

## Per-grammar procedure

1. Record the runtime, grammar, blob, and source identities.
2. Run the smallest locked corpus in Docker.
3. Record the first exact mismatch.
4. Reduce that mismatch to the smallest source.
5. Add a regression test before implementing a correction.
6. Re-run the grammar until it is exact.
7. Publish the grammar receipt before moving to the next grammar.

Do not batch grammars while diagnosing a mismatch or memory failure.

## Certification receipt

Record for each grammar:

- Grammar identity and artifact digest.
- Source manifest digest.
- Fresh exact, mismatch, fallback, skip, and error counts.
- Incremental exact, mismatch, fallback, skip, and error counts.
- The maximum live frontier and derivation counts.
- The compact route counters.
- The first divergence for each accepted exception.
- The Docker artifact location and digest.

## Exit gate

Complete this package only when every supported grammar has a reproducible receipt.

Do not carry an unexplained mismatch into package seven. Document any accepted exception with its exact reopening condition.

odvcencio:
Scala preliminary receipt: issue #1067.

The isolated 25-case run found 6 exact deep trees, 14 generated errors on clean C trees, and 5 clean structural mismatches. It used 3,053,268 KiB maximum resident set and had no timeout or out-of-memory failure.

This is a generated-parser baseline problem, not yet a compact-route verdict. Issue #1067 now owns the reductions and the strict Scala rerun. Package-six certification must not count these failures as compact recovery exceptions.

odvcencio:
PR #1124 harvests a focused generator fix: resolve public aliases in supertype maps and remove duplicate public subtype entries.

The synthetic regressions fail on main and pass with the fix. The generated Go `_simple_type` map also matches locked C with the fix.
Without the fix, that map omits `type_identifier` and adds `identifier`. The bounded Docker runs completed without an out-of-memory failure.

[...]

PR #1124, opened 2026-09-17
fix(grammargen): resolve public aliases in supertype maps

odvcencio:
Generated supertype maps record internal symbols when a production exposes a public alias. This rejects valid subtype queries and accepts invalid ones.

Resolve aliases through the existing name-and-namedness map. Remove duplicate subtype entries after resolution. Preserve the first occurrence of each public symbol.

Validation used bounded Docker runs on chi-1:

- The new synthetic regressions fail on main and pass with the fix.
- The generated Go `_simple_type` map fails against locked C on main and matches with the fix.
- Existing alias, supertype, serialization, and generated C runtime checks pass.
- Go nonterminal alias-map parity passes with the pinned grammar source mounted.
- Continuous integration now executes both synthetic alias tests and the seeded Go C-oracle regression.
- The exact added commands pass in bounded Docker runs. The workflow passes actionlint.

Reproduce the C regression after seeding Go with `cgo_harness/seed_parity_repos.sh --langs go`:

```sh
cd cgo_harness
GTS_PARITY_REPO_ROOT=/tmp/grammar_parity go test . -tags treesitter_c_parity -run '^TestGeneratedGoAliasedSupertypeLockedCParity$' -count=1
```

Run that command inside `cgo_harness/docker/run_parity_in_docker.sh`, with the seeded source mounted at `/tmp/grammar_parity`.

The original changes came from the preserved `gts-go-alias-candidate-20260912` worktree. No embedded grammar blobs changed.
The existing embedded Go query divergence therefore remains until its artifact update passes the required gates.
An exploratory comparison found a separate `_statement` difference: generated tables include `_declaration`, which C omits.

Refs #1064. This change does not certify complete Go parity or close that issue.


odvcencio (commented):
Reviewed commit `d75b62064d64a357e9fd302c05d39d57ab2476af`. No blocking source defect found.

- Alias resolution uses the exact name and namedness pair from the alias table.
- Duplicate public symbols retain their first occurrence order.
- Regression cases cover named aliases, anonymous name collisions, and subtype query captures.
- The workflow seeds the pinned Go grammar before comparing its generated supertype map with C.

Focused regression tests pass in Docker. Hosted checks are running for the updated commit.

odvcencio (commented):
Reviewed integration commit `39fa33ffecd29424c550f7eaa81aaef91afefbba`. No blocking source defect found.

PR #1125, opened 2026-09-18
fix(grammargen): preserve HCL splat associativity

odvcencio:
Generated HCL parsers place later attributes and indices outside splats in nested resource blocks. This produces a different tree from locked C.

Apply right associativity when structural checks prove a repeat continuation with equal precedence. Preserve existing behavior outside that boundary.
Add a blocking conflict regression and an explicit locked C regression to continuous integration.

Validation ran in bounded Docker containers:

- Main fails the conflict regression and the nested HCL fixture against both references.
- The candidate passes the existing focused conflict tests and six associativity boundary cases.
- Five direct fixtures match locked C, including nested splats and multiple indices.
- Direct C corpus parity passes for 100 of 100 eligible samples.
- Strict real-corpus parity passes for 100 of 100 samples, including deep tree structure.

The change harvests the focused source fix from `58db14c2` and adds current-main verification.
No expected trees, corpus floors, or embedded grammar blobs changed. No runtime performance improvement is claimed.

Refs #1064.


odvcencio (commented):
Reviewed commit `d6d8f3dcc7c559e56c1cb4dc7218f5b9281a5d62`. No blocking source defect found.

- The change selects a shift only after proving a lowered repeat continuation.
- Right associativity applies only when the shift and reduction have equal precedence.
- The regression preserves conflicts for other associativity and precedence combinations.
- The workflow compares five generated splat fixtures with the pinned C grammar.

Focused regression tests pass in Docker. Hosted checks are running for the updated commit.

odvcencio (commented):
Reviewed integration commit `13db47f391ac4704e72e506375607035de894269`. No blocking source defect found.

PR #1126, opened 2026-09-18
fix(grammargen): preserve immediate token precedence

odvcencio:
Generated CSS parsers truncate longer values after an immediate token accepts a prefix. For example, `padding-right: 8px\9` produces error nodes where locked C accepts a plain value.

Preserve authored precedence for named immediate tokens. Keep paths to longer preferred tokens after an immediate accept.
Add blocking lexer regressions and explicit locked C coverage to continuous integration.

Validation ran in bounded Docker containers, one grammar at a time:

- Main fails the lexer, normalization, and CSS regressions.
- The candidate passes eight lexer and precedence boundary cases, plus the existing focused tests.
- Four direct CSS fixtures match locked C.
- Direct C parity and strict real-corpus parity both pass for all 43 eligible CSS samples.

The broader C grammar gate retains existing failures: 71 matching trees from 85 eligible samples, with 23 reported divergences.
The generated C grammar blob changes, so its bytes are not identical.
A diagnostic overlay confirms identical source and `dump.v1` tree hashes for all 85 C samples before and after this patch.
This preserves observed C corpus behavior; it does not certify complete C parity.

The change harvests the focused source fix from `b7f0a14e` and adds current-main verification.
No expected trees, corpus floors, or embedded grammar blobs changed. No runtime performance improvement is claimed.

Refs #1064.


odvcencio (commented):
Reviewed commit `6c5ce01155600ca30439a112f1e20c34dee7b1fe`. No blocking source defect found.

- Named immediate tokens retain authored lexical precedence across all three extraction groups.
- Transition pruning preserves paths to preferred tokens and tokens with better priority.
- The regression retains pruning for an unpreferred token at equal priority.
- The workflow compares ordinary and escaped unit fixtures with the pinned C grammar.

Focused regression tests pass in Docker. Hosted checks must complete for the updated commit.

odvcencio (commented):
Reviewed integration commit `ec3835877bd2afcec498f26c9d5a41dcae8e2518`. No blocking source defect found.

The final combined tree also passes the generated Go supertype and HCL splat comparisons with the pinned C grammars.

PR #1128, opened 2026-09-18
perf(parsercore): reuse recovery symbol tables within each parse

odvcencio:
Repeated recovery pricing rebuilt the language symbol table during one compact parse. Reuse one projection across all nine scheduler call sites. Discard it at reset and include its capacity in memory accounting. The change has no language-specific conditions.

The real Go query compiler fixture allocates **81.58% fewer bytes** and **37.22% fewer objects** per parse. Twenty paired seeds report 69,428 to 12,788 bytes and 317 to 199 allocations. Parse timing shows no significant change. The Go primary benchmark trio and clean JavaScript control show no significant regression. A separate 20-pair Python comparison also shows no significant change.

Validation passed in bounded Docker containers:

- Recovery, visibility, reset, and memory-accounting regressions, including race detection.
- Fresh and incremental C parity for Go, Python, and TypeScript, one language per container.
- JavaScript recovery selection and incremental recovery against the pinned C parser.
- PHP issue 454, including complete tree digests and fallback behavior.
- Both large Go memory-budget regressions on a 19.9 MB source.

The retained table counts toward memory limits, so a parse near its limit can stop earlier by that table's size. The large-source resource probes record 579,504 and 580,412 KiB maximum resident memory. One observation per revision does not establish a resident-memory improvement or regression.

[Read the reproducible report, raw benchmark samples, source hashes, and memory logs](https://github.com/odvcencio/gotreesitter/blob/codex/shared-parser-perf-20260918/docs/perf/recovery-symbol-policy/README.md).


odvcencio (commented):
Reviewed commit `dc055fe07652293aac5d0d9967dd5fe7a14a7f58`. No blocking source defect found.

- All nine scheduler call sites use the same immutable symbol projection.
- Scheduler reset discards the projection. The regression test verifies changed metadata between parses.
- Footprint accounting includes the retained table capacity. Poll frequency remains unchanged.
- Projection width, default visibility, and namedness preserve the previous recovery-cost inputs.
- Twenty paired samples support the allocation claims. The report correctly avoids a timing-improvement claim.
- Local recovery, race, C-parity, and large-source memory checks passed.

The selected-store builder still duplicates the symbol projection. Consolidate that logic in a separate maintainability change.

This is an author-side code review. Hosted checks are still running, and GitHub requires approval from another account before a normal merge.

PR #1132, opened 2026-09-19
fix(tree): Fix use-after-release in unchanged incremental parse

odvcencio:
## Summary

An unchanged incremental parse returned the old tree pointer itself, so a caller that released the old tree also invalidated the result. This is the common C idiom: release the old tree right after ParseIncremental. The fix gives each returned tree its own handle, so callers can release the old tree and the new tree independently. It also removes the Tree sync.Pool, which let a stale pointer release a later parse result.

## Changes

- Add an extraHandles counter to Tree. retainUnchangedIncrementalResult adds one handle when an unchanged parse returns its old tree; Release drops one handle before it frees the tree.
- Route all six unchanged-source fast paths (parseIncrementalInternalWithMergePerKeyOverride, ParseIncremental, ParseIncrementalProfiled, parseIncrementalWithTokenSource, ParseIncrementalWithTokenSourceProfiled) through the new retain helper.
- Remove the Tree sync.Pool. A caller can keep a *Tree pointer after Release, and a pooled value would let that stale pointer free a later parse result. Trees are now allocated fresh in newTreeWithUniqueArenas.
- Saturate the handle count at maxTreeExtraHandles. Release never frees a saturated tree; the garbage collector reclaims it when unreachable.
- Make Release on an already-released tree a no-op instead of a double-free path.
- Document the ownership contract in README.md and on ParseIncremental: release each returned tree once, and do not read the old tree from another goroutine during an incremental parse because shared-node parent links are updated in place.
- Clarify in Node.Edit docs that callers must not apply both Tree.Edit and Node.Edit for the same edit; C code that calls ts_node_edit after ts_tree_edit needs no Go equivalent.
- Add black-box tests for the C release-idiom, the identity-check idiom, and double-Release safety, plus internal tests for handle saturation and retain-after-release.

## Testing

- go test . -run 'TestTreeHandleLifetime|TestTreeReleaseTwiceKeepsLaterParseValid|TestParseIncrementalUnchanged' -count=1
- go test . -run 'TestTreeExtraHandlesSaturate|TestTreeRetainAfterReleaseDoesNothing' -count=1
- bash scripts/run_randomized_benchmarks.sh --output /tmp/gotreesitter-bench.txt to confirm the unchanged-parse fast path stays at single-digit ns/op with zero allocations

## Related Issues

- Refs #308

## Review Focus

Check that every unchanged-tree return path uses the retain helper, and that no remaining code path returns a pooled Tree.


odvcencio:
Performance note. This is a quick local run, not randomized evidence: GOMAXPROCS=1, -benchtime=500ms, -count=3, on the same host as the main baseline at 152e4e89.

| Benchmark | main | this branch |
|---|---|---|
| BenchmarkGoParseFullDFA | ~9.9 ms, 37 allocs | ~9.9 ms, 38 allocs |
| BenchmarkGoParseIncrementalSingleByteEditDFA | ~130 µs, 123 B, 3 allocs | ~133 µs, 3,628 B, 5 allocs |
| BenchmarkGoParseIncrementalNoEditDFA | 2.66 ns, 0 allocs | 2.85 ns, 0 allocs |

Tradeoff: each returned tree now costs one new `Tree` value. That value is 3,296 bytes, and 3,096 of them are the embedded `ParseRuntime`. The tree pool made a stale `Release` free a later parse result. Safety wins here.

Follow-up: move `ParseRuntime` out of `Tree` into a separately pooled block (31 production references). Callers never hold pointers to that block, so pooling it is safe, and it removes most of the byte cost.

PR #1137, opened 2026-09-19
add(parser): Scale default parse memory budget with input size and add per-parser API

odvcencio:
## Summary

The fixed default per-parse memory budget stopped large valid inputs, so callers had to raise it manually for every file. The default now grows with the input, which removes that manual step while a new per-parser setting still caps memory in constrained services.

## Changes

- Default per-parse memory budget is now the larger of 512 MiB and 512 bytes per source byte. The factor keeps about 2.5x headroom over the 7-203 bytes/source measured on valid corpus inputs (September 2026). Pathological growth stays superlinear, so a linear budget still stops it.
- Added `Parser.SetMemoryBudgetBytes` / `Parser.MemoryBudgetBytes`: positive sets a fixed budget, zero restores the default, negative turns the per-parse budget off. The value lives in the parser cold state, so setting zero on a parser with no cold state allocates nothing.
- Added `WithParserPoolMemoryBudgetBytes` pool option; `applyDefaults` re-applies the budget on each checkout.
- The process-heap hard ceiling now scales with the effective budget: max(2 GiB, 2x budget), clamped against overflow via `maxScaledParseMemoryBudget` (1<<60). This prevents the backstop from firing before the scaled budget.
- `GOT_PARSE_MEMORY_BUDGET_MB` and `GOT_PARSE_MEMORY_HARD_CEILING_MB` keep their old fixed semantics: an explicitly set environment value disables scaling for that knob (tracked by new `...FixedByEnv` helpers).
- Updated the phase-0 fresh-full runner's stop-control path to use the parse-aware ceiling helper so diagnostics match production routing.
- Documented the new default, the setter, the pool option, and the `ParseStopMemoryBudget` partial-tree behavior in README.md.
- Added internal and black-box tests covering scaling, env-fixed values, setter override/restore/off semantics, pool application, and end-to-end stop reasons.

## Testing

- go test . -run 'TestParseMemoryBudget|TestSetMemoryBudgetBytes|TestDefaultMemoryBudgetScalesWithInput|TestParserPoolMemoryBudgetBytesOption' -count=1
- go test ./... -run 'MemoryBudget' (scoped; avoid repo-wide host sweeps per AGENTS.md)

## Review Focus

Check the interaction between the scaled budget, the derived hard ceiling, and the env-var escape hatches; confirm no path lets the 2 GiB default ceiling fire before a legitimately scaled budget.


odvcencio:
Measurement behind the scaled default. For each input, a binary search finds the smallest per-parse budget that still returns `accepted`. The measurement ran on main plus this branch, using `SetMemoryBudgetBytes`.

| Input | Size | Smallest budget | Budget bytes per input byte |
|---|---|---|---|
| Bash (concatenated scripts) | 1.0 MB | 203 MiB | 203 |
| Python (concatenated modules) | 1.0 MB | 174 MiB | 181 |
| Go `proc.go` | 0.12 MB | 18 MiB | 159 |
| C `cluster.c` ×8 | 1.7 MB | 256 MiB | 155 |
| JSON, valid or with errors | 3.7 MB | 489 MiB | 140 |
| HTML, valid generated | 1.1 MB | 35 MiB | 32 |
| Rust, PHP, Java, C | 0.2–1.0 MB | — | 27–32 |
| YAML, CSS, TypeScript, JavaScript | 0.25–2.0 MB | — | 13–22 |
| C++, Ruby | 1.0 MB | — | 7–12 |
| HTML, one file repeated 30× (error recovery) | 1.0 MB | 663 MiB | 666 |

New default: the larger of 512 MiB and 512 bytes per input byte. That is about 2.5× headroom over the worst valid input. Heavy error recovery on multi-megabyte files can still stop, and bounding that case is what the budget is for. The hard ceiling is the larger of 2 GiB and twice the budget.

Original failure: valid 7.1 MB JSON used to stop with `memory_budget` and report `HasError() == true`. With default settings it now returns `accepted`, peak RSS 1.18 GB.

PR #1138, opened 2026-09-19
fix(parser): Promote reserved words to keyword token to match C

odvcencio:
## Summary

The Go parser treated a reserved word as a plain identifier when the current parse state had no action for that keyword. C tree-sitter instead promotes it to the keyword token and reports the error there. This PR fixes that divergence in promoteKeyword, and closes the data gap behind it: six checked-in grammar blobs predate cmd/ts2go's ABI 15 reserved-word extraction, so their decoded languages carried no ts_reserved_words table at all. Reserved-word sidecars now supply that table through a fail-closed attach guard.

## Changes

- parser_dfa_token_source.go: promoteKeyword now sets tok.Symbol to the keyword symbol and marks the keyword flag before returning false for a reserved word, so the parser fails on the keyword exactly as C's ts_parser__lex does.
- cmd/ts2go: adds a -reservedwords-only batch mode (RunReservedWordsOnlyManifest) that regenerates only *_reserved_words_gen.go sidecars. It never touches blobs, register stubs, or the embedded loader aggregate, so it is safe against blobs pinned to an older ABI. It supports the -only language filter and skips grammars with no ts_reserved_words table.
- grammars/runtime: adds ReservedWordTable, RegisterReservedWords, LookupReservedWords, and attachRegisteredReservedWords. The attach fails closed: it requires that the language has no reserved-word data yet, that the decoded symbol count matches, that every referenced symbol name matches its recorded generation-time name, and that each lex mode's ReservedWordSetID indexes a valid row.
- grammars/runtime/embedded_loader.go: calls attachReservedWordsForBlob inside decodeLanguageBlobData before the repair helpers that append synthetic symbol names, which keeps the exact symbol-count check meaningful.
- Adds generated sidecars for javascript, python, php, ocaml, pkl, and templ.
- scripts/check_generated_ownership.sh: claims grammars/runtime/*_reserved_words_gen.go for cmd/ts2go.

## Testing

- go test ./grammars/runtime -run '^TestAttachRegisteredReservedWords' -count=1
- go test ./grammars -run '^TestReservedWord|^TestJavaScriptReservedWordBlocksVarIf' -count=1
- go test . -run '^TestReservedWordForcesPromotion$|^TestNonSQLKeywordPromotionRemainsCaseSensitive$' -count=1
- bash cgo_harness/docker/run_single_grammar_parity.sh javascript

## Related Issues

- 

## Review Focus

Review the behavior change in promoteKeyword first: this branch reverses an existing test expectation (TestReservedWordBlocksPromotion renamed to TestReservedWordForcesPromotion). Second, confirm the attach guard runs before the synthetic-symbol repair helpers, since the guard validates an exact symbol count.


odvcencio:
Comparison with locked C for tree-sitter-javascript 58404d8c, using the tree-sitter CLI 0.26.6.

- `var if = 1;`
  - C: `(program (ERROR [0,0]-[0,8]) (expression_statement (number)))`
  - Before this PR: a clean `variable_declaration` with no error.
  - After this PR: `(program (ERROR [0,0]-[0,8] (ERROR [0,7]-[0,8])) (expression_statement (number)))`.
  - The top-level shape now matches C. One nested ERROR node remains, which is a recovery-shape difference outside this change.
- `const o = { if: 1, class: 2 }; o.if = o.class; class A { if() { return this.while; } }`
  - C: 0 ERROR nodes.
  - After this PR: 0 ERROR nodes. Keywords stay valid in property and method positions.

Blob hashes do not change. The reserved-word tables attach at load, from sidecars generated from the locked parser.c. The attach requires an exact match of symbol count, symbol names, and set IDs, so runtime profiles pinned to a blob hash still attach.

PR #1140, opened 2026-09-19
optimize(parser): Skip included-range check on no-edit fast path

odvcencio:
## Summary

Tree reuse for an unchanged source calls includedRangesMatchTree on every parse, even though most parses have no included ranges. The branch now checks the range counts first and skips the call when both the old tree and the request carry no included ranges. This removes needless work from the no-edit fast path measured by BenchmarkGoParseIncrementalNoEditDFA.

## Changes

- Add a guard in canReuseUnchangedTree that only calls includedRangesMatchTree when either oldTree.includedRanges or the requested included list is non-empty.
- Keep behavior identical: trees with matching empty range sets still return true from the reuse check, so no-edit reuse still succeeds.

## Testing

- bash scripts/run_randomized_benchmarks.sh --output /tmp/gotreesitter-bench.txt and compare BenchmarkGoParseIncrementalNoEditDFA ns/op and allocs/op against the main baseline.
- Run the focused parser unit tests inside Docker isolation to confirm no-edit tree reuse still works.


odvcencio:
Paired randomized evidence: 20 shuffle seeds, alternating order, `-benchtime=750ms`, `GOMAXPROCS=1`, in the harness container with 4 GiB of memory and one pinned CPU. Baseline: the v0.53.0 candidate code (`48503fef`). Head: this branch.

| Benchmark | Baseline | Head | Change |
|---|---:|---:|---:|
| GoParseIncrementalNoEditDFA | 4.410 ns | 3.603 ns | -18.30% (p=0.001, n=20) |
| GoParseFullDFA | 10.26 ms | 10.59 ms | no significant change (p=0.529) |
| GoParseIncrementalSingleByteEditDFA | 136.2 µs | 133.3 µs | no significant change (p=0.718) |

This recovers part of the 1.4 ns no-edit regression recorded in the v0.53.0 release notes. The rest comes from the tree handle increment.

PR #1146, opened 2026-09-19
optimize(parser): Snapshot hot-path env knobs and keep warm compact runner in pool

odvcencio:
## Summary

Two incremental-parse hot-path optimizations. The parser now reads its environment configuration knobs once per process instead of calling os.Getenv on every parse; os.Getenv takes a process-wide lock, so the per-parse reads added latency to each keystroke. The parser pool also keeps a checked-out parser's warm compact runner across reuse, so pooled parsers do not reallocate their arenas on the next parse.

## Changes

- Add a sync.Once-guarded parseEnvKnobs snapshot in parser_config.go that captures GOT_TRANSIENT_REDUCE_CHECKPOINT_MB, the three GLR v2 boolean knobs, the transient-reduce enable/language names with their documented fallbacks, GOT_PARSE_PROGRESS, and GOT_C_RECOVERY.
- Route parseTransientReduceCheckpointBytes, parseCompactFullLeavesEnv, parsePendingParentsEnv, parseFinalChildRefsEnv, parseTransientReduceEnabled, parseTransientReduceLanguageEnabledForSource, errorCostCompetitionLanguage, cRecoveryGateReason, and hasActiveParseObservability through the snapshot instead of os.Getenv.
- Extend ResetParseEnvConfigCacheForTests to clear the new knob snapshot so tests can refresh it after t.Setenv.
- Stop clearing admissionCandidateRunner in ParserPool.applyDefaults: the runner holds no request state between parses, SetParseWorkLimits still drops it when limits change, and keeping it avoids a cold arena allocation after each checkout.
- Update roughly 40 existing env-sensitive tests to reset the cache after each t.Setenv and register a cleanup reset.
- Add parser_env_knobs_test.go (snapshot read-once, reset refresh, transient-reduce fallback names) and parser_pool_runner_test.go (warm runner survives applyDefaults; changed work limits drop it).

## Testing

- go test . -run 'TestParseEnvKnobs|TestParserPoolKeepsWarmCompactRunner|TestTransientReduce|TestCRecoveryGate|TestParseEnvConfig' -count=1
- bash scripts/run_randomized_benchmarks.sh --output /tmp/gotreesitter-bench.txt and compare ns/op, B/op, allocs/op for BenchmarkGoParseIncrementalSingleByteEditDFA against baseline

## Review Focus

Check that no remaining per-parse os.Getenv call needs the same treatment, and that the behavioural change to env-var edits mid-process is acceptable (values are now fixed at first use outside tests).


odvcencio:
Paired randomized evidence: `scripts/run_randomized_benchmarks.sh`, 20 shuffle seeds, alternating order, `-benchtime=750ms`, `GOMAXPROCS=1`, tag `gts_parsercorephase0`, in the harness container with 4 GiB memory and one pinned CPU. Baseline: `be2fd62e` (the branch point). Head: this branch at `e325f3e0`. Other work ran on the host at the same time, so absolute times are higher than the release epoch; the paired design controls for that.

| Benchmark | Baseline | Head | Change |
|---|---:|---:|---:|
| ParserPoolSerial | 15.72 ms, 10.03 MiB/op | 13.36 ms, 3.08 MiB/op | −15.02% time (p=0.000), −69.34% B/op |
| GoParseFullDFA | 12.50 ms | 12.52 ms | no significant change (p=0.947) |
| GoParseIncrementalSingleByteEditDFA | 155.8 µs | 152.8 µs | no significant change (p=0.529) |
| GoParseIncrementalNoEditDFA | 4.06 ns | 4.21 ns | no significant change (p=0.723) |

The pool gain comes from keeping the warm compact runner across checkouts. The environment snapshot removes about twelve `os.Getenv` calls per parse; that cost sits below the noise floor of these benchmarks, as expected, and the change is kept for the reduced lock traffic under concurrent parsing.

Raw evidence: `/home/draco/work/gts-env-knob-evidence-20260919` (base.txt, head.txt, container metadata).

PR #1155, opened 2026-09-19
fix(parser): Make bash stack culling arena-independent

odvcencio:
## Summary

Bash used the generic GLR stack comparator only on incremental arenas, so an incremental parse could cull branches in a different order than a fresh parse and select a different tree. This makes stackCullLanguageForArena return the language's comparator regardless of arena class, and adds a regression gate that proves incremental and fresh bash parses agree.

## Changes

- Remove the arena-class special case in stackCullLanguageForArena (parser.go). A parse now always uses its language's cull comparator, so incremental and fresh parses cull GLR stacks in the same order.
- Document the invariant in the function comment: the arena class must not change which GLR branches survive.
- Add TestIncrementalInvariantGateBash to incremental_invariant_gate_test.go to guard the shared cull order between incremental and fresh bash parses.
- Add testdata/incremental_gate/bash_clean_old.sh (a real-world 165-line bash script) and register it in incrGateCorpus with stride 2 as the bash gate input.

## Testing

- go test . -run '^TestIncrementalInvariantGateBash$' -count=1


odvcencio:
Docker C-parity evidence for bash on this branch (harness container, 4 GiB, `GTS_PARITY_MODE=smoke GTS_PARITY_LANGS=bash`):

- --- PASS: TestParityHasNoErrors (0.14s)

Bash subtests passed: 2. Package result: `ok`, exit 0, no OOM.

Host gates: the new `TestIncrementalInvariantGateBash` reports 0 divergences over 5,450 clean edit sites both before and after the change; the nine bash regression tests pass.

